PR #27, opened 2026-04-22
Fix isSessionReadyForPrompt regex that classified every idle session as busy

zollak:
## Summary
`isSessionReadyForPrompt` guards against text already parked in the input buffer with:

```ts
const slice = lines.slice(start, footerIdx === -1 ? undefined : footerIdx).join('\n')
if (/❯\s+\S/.test(slice)) return false
```

Because the lines are joined with `\n` and `\s` matches `\n`, the regex spans from an idle input line (`"❯ "`) across the newline into the decorative horizontal-rule `─` character on the line below — so `/❯\s+\S/` matches **every** idle pane.

Net effect: `startMessageRouter` and the schedule runner both call `isSessionReadyForPrompt` before sending; with every session reporting busy, every outbound message sat in `pending` forever, even on completely idle agents.

## Fix
Iterate the candidate lines and test `/❯[^\n]*\S/` *within each line*. That keeps the intent — "the ❯ input prompt has non-whitespace content after it" — while refusing to span newlines.

## Observed
On an install with four idle sub-agents plus an idle main agent, `tmux capture-pane` for every session showed the `bypass permissions` footer and an empty `❯` input line, yet every router tick logged `Router skip: session busy`. After the fix, messages drain normally.

## Test plan
- [ ] Leave all agents idle. Post an inter-agent message via `POST /api/messages`. Verify the next router tick delivers it (not `session busy`).
- [ ] Type some text into an agent's tmux input (without pressing Enter). Verify `isSessionReadyForPrompt` still returns false for that session.
- [ ] Verify `[Pasted text #N]` detection still works.

### DIFF
--- a/src/web.ts
+++ b/src/web.ts
@@ -1483,8 +1483,13 @@ function isSessionReadyForPrompt(session: string): boolean {
     const lines = pane.split('\n')
     const footerIdx = lines.findIndex(l => /bypass permissions on \(shift\+tab to cycle\)/.test(l))
     const start = Math.max(0, footerIdx - 20)
-    const slice = lines.slice(start, footerIdx === -1 ? undefined : footerIdx).join('\n')
-    if (/❯\s+\S/.test(slice)) return false
+    const sliceLines = lines.slice(start, footerIdx === -1 ? undefined : footerIdx)
+    // Match ❯ followed by non-whitespace ON THE SAME LINE. The previous
+    // version joined the lines with \n and used /❯\s+\S/, where \s matches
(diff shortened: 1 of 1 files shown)


zollak:
Superseded: upstream already has a better fix on main. commit 6aaee7d-equivalent arrived via the batch of PRs merged shortly after this one (scans only inside the ── separators, adds an 'esc to interrupt' busy check). My version would have been redundant and slightly worse (the scrollback-scan version still matched historical ❯ lines).

PR #32, opened 2026-04-23
connectors: show every active MCP and detect installed state correctly

koorbela:
## Why this matters

The dashboard's **Connectorok** page currently gives a misleading view of the MCP layer:

1. **Installed tab lists 1 of 8 active MCP servers.** The endpoint reads `~/.claude/settings.json` `enabledPlugins`, `project/.mcp.json`, and `~/.claude.json` `mcpServers` directly. It skips the claude.ai OAuth connectors because those have no file-based representation, they only surface through `claude mcp list`.

2. **Gallery state-detection is broken.** The parser reads `claude mcp list` but does not strip the `claude.ai ` or `plugin:<pkg>:` prefixes from the reported names. Every catalog item appears "not installed" even when the matching claude.ai connector is already active. A well-meaning click on "Gmail" → "Telepítés" then runs `claude mcp add` and creates a duplicate entry in `~/.claude.json`, after which `claude mcp list` emits a "local wins over claude.ai" warning.

3. **Built-in section rows look clickable but are plain `<span>`s** with the enable instructions hidden in a `title` tooltip. The user's mental model ("click to enable") is broken.

4. **Cold-start / refresh failure shows no feedback.** The previous frontend rendered "Nincsenek MCP connectorok" whenever the list happened to be empty, contradicting the info-box that says the data loads at dashboard start.

Reproducible on any host whose user has a claude.ai subscription with the Gmail / Google Calendar / Google Drive connectors enabled: the dashboard marks all three "not installed" and the Installed tab only shows the telegram plugin.

## Change

Fixes #1 and #2 without invasive restructuring. Three distinct areas:

**Cache layer (new)**

- `mcpListCache` holds the parsed output of `claude mcp list`. Refreshed once at startup with a 30s delay (so the main channels session settles first), and on explicit `POST /api/connectors/refresh`. No periodic auto-refresh: every call spawns stdio and plugin MCPs for a health check, which can race the live Telegram bot's single-poller.
- `applyRefreshOutcome` keeps the previous entries when the CLI exits non-zero but stdout still parsed something (the steady state when a connector's OAuth expired). Only a blank stdout plus an exec error wipes the cache and surfaces `cache.error`.
- The working directory for `claude mcp list` is a `mkdtempSync` dir with `0700` perms, created on demand and cleaned up via `process.on('exit', ...)`. Avoids the `/tmp/.mcp.json` poisoning vector on multi-user hosts.
- `scrubPaths` strips absolute user paths (Users/home/root/tmp/var/opt/private/var) from any error message before it lands in an API response or a log line. The regex has a trailing lookahead so dirs whose names extend a keyword (`/homes/...`, `/Users-backup/...`, `/opts-archive/...`) do not trigger partial scrubs, and zero-or-one-segment tails collapse to `<path>` so the username itself is not leaked. 25 tests cover the scrubber alone.

**`src/mcp-list-parser.ts` (new pure module)**

- `parseMcpListLine`, `parseMcpList`, `slugify`, `applyRefreshOutcome`, `scrubPaths` are pure functions with no Node side effects. `src/web.ts` injects the node-specific runtime (exec, homedir, fs) around them.
- 52 vitest cases: line-shape guards, claude.ai / plugin / local sources, slug normalisation for dotted and parenthesised names, priority ordering, retained-stale outcomes, full sweep of path-scrubbing edge cases.

**`/api/connectors` and `/api/mcp-catalog` rebuilt**

- `/api/connectors` now merges file-based sources AND the cached `claude mcp list` output, tagging each entry with a `source` field (`'plugin'`, `'claude.ai'`, `'local-user'`, `'local-project'`, or `'local'` for CLI-only entries whose scope we cannot determine). Dedupe keys are lowercased so `plugin:Foo` and `plugin:foo` collapse into one row.
- Plugin detail endpoint reads the last colon-separated segment, so `plugin:telegram:telegram` resolves to the correct settings key, and disabled plugins (`enabledPlugins[k] === false`) return 404.
- `/api/connectors/status` returns `cacheLastRefreshed` + `cacheError` + `refreshing` so the UI can distinguish cold-start from a genuinely empty list.
- `/api/connectors/refresh` returns HTTP 502 on upstream error so external monitoring (uptime probes, CI scripts) see the failure.
- `/api/mcp-catalog` installed-state now fuzz-matches catalog `id` and `name` against the parser's slug-normalised ids, so Gmail / Google Calendar / Google Drive / Notion / Slack resolve as installed when the claude.ai connector is active.

**Frontend**

- "Frissítés" button on the Connectors page that triggers `POST /api/connectors/refresh` and surfaces `{ok, count, error}` in a toast.
- Info-box at the top of the page explaining that sub-agents share the main session's HOME and see every bekapcsolt MCP automatically.
- Source badge per entry in the Installed tab (`claude.ai`, `plugin`, `local`, etc.).
- Read-only cards for `source === 'claude.ai'` entries: no click handler, no hover highlight, `Kezelhető: claude.ai` hint. Prevents a click that would 404 on the file-backed detail endpoint.
- Cold-start state machine: `connectorCacheWarming` defaults to `true` and only flips to `false` after an authoritative `cacheLastRefreshed > 0` reading. A failed refresh on a cold cache shows the actual error instead of instructing the user to click the button that just failed.
- Stale-data banner above the grid when the cache has live entries from a previous successful refresh, the latest refresh failed, AND claude.ai entries are present (file-based entries are always fresh, so the banner is scoped to the subset that can go stale).
- Gallery "Telepítés" button is hidden when the catalog item is already active (`installedSource === 'claude.ai'`), so a click cannot produce the duplicate-entry symptom the PR was born from.

## Scope / compatibility

Left intact:
- The "Hozzárendelés" per-agent file-copy flow (`POST /api/connectors/:name/assign` and `POST /api/skills/:name/assign`). In today's single-HOME setup every sub-agent inherits `~/.claude/*` already, so the copy is a no-op; the endpoints survive for a potential multi-HOME deployment and are simply not wired to new UI in this PR.
- `mcp-catalog.json` shape.
- Existing `/api/connectors/:name` detail endpoint semantics for file-based sources.
- Refresh is manual-only; no background poll that could surprise the Telegram bot.

API additions (all forward-compatible -- consumers that ignore unknown fields are unaffected):
- `/api/connectors` items gain a `source: 'plugin' | 'claude.ai' | 'local-user' | 'local-project' | 'local'` field.
- `/api/mcp-catalog` items gain `installedSource: string | undefined`, and `installed` is derived from normalised-slug matches.
- `GET /api/connectors/status` (new) returns `{ cacheLastRefreshed, cacheError, refreshing }`.
- `POST /api/connectors/refresh` (new) returns `{ ok, count, lastRefreshed, error? }` with HTTP 502 on upstream error, 200 on success.

## Test plan

- [x] `npx vitest run` -- 130/130 passing (52 in `mcp-list-parser.test.ts`)
- [x] Live smoke: dashboard on a host with 7 claude.ai connectors + telegram plugin. Before the patch: Installed tab shows 1 entry, Gallery marks Gmail/Google Calendar/Google Drive as "not installed". After: Installed shows all 8 with source badges, Gallery marks the three as "Telepítve ✓ (claude.ai)" and hides the install button.
- [x] Refresh while claude.ai Canva connector is active: `POST /api/connectors/refresh` returns `{ok:true, count:8}` in ~5s, telegram bot remains alive.
- [x] Pidfile-style stale-cache test: planted a stale `cacheError`, reloaded the Connectors page, stale banner rendered above the entries.
- [x] Cold-start test: hit the page 1s after boot (before the 30s warmup); UI showed the "MCP lista még nem töltődött be..." message instead of "Nincsenek MCP connectorok".
- [x] Sensitive-data audit on the diff: zero matches for operator/user/agent/path identifiers.
- [x] Em-dash audit on the diff: zero U+2014.

## Known limitations

- Windows path scrubbing is not implemented (`C:\Users\foo\...` patterns). Marveen's deploy target is darwin + Linux today; worth a separate portability pass.
- `claude mcp list` still spawns stdio and plugin MCPs for a health check on every invocation, which is why the refresh is manual-only. A `claude mcp list --no-spawn` flag (if it ever ships) would let us auto-refresh on a timer without racing the Telegram bot.
- The wiring around `/api/connectors` merge, plugin dedupe, and last-segment parsing has no dedicated integration test; the pure-logic helper underneath has 52 cases. A future test harness that stubs `mcpListCache` and `settings.enabledPlugins` could close that gap cheaply.

## Context

This PR is the first of a 3-PR series addressing the Connectors page behaviour. The next two are:

- PR2: Skills page shows user skills + plugin skills + built-in skills (listing parity with the Claude CLI), with the "Hozzárendelés" gomb removed from the UI because every sub-agent already inherits the user-global set.
- PR3: Built-in section (Computer Use, Claude in Chrome) made into a real clickable row that opens a modal with the enable instructions.

This PR stands on its own and can merge before the others.

### DIFF
--- a/src/__tests__/mcp-list-parser.test.ts
+++ b/src/__tests__/mcp-list-parser.test.ts
@@ -1,0 +1,449 @@
+import { describe, it, expect } from 'vitest'
+import {
+  parseMcpListLine,
+  parseMcpList,
+  applyRefreshOutcome,
+  scrubPaths,
+  type McpListEntry,
+} from '../mcp-list-parser.js'
+
+describe('parseMcpListLine -- guard rails', () => {
+  it('returns null for an empty line', () => {
+    expect(parseMcpListLine('')).toBeNull()
+  })
+
+  it('returns null for whitespace-only', () => {
+    expect(parseMcpListLine('   \t  ')).toBeNull()
+  })
+
+  it('returns null for the "Checking MCP server health..." banner', () => {
+    expect(parseMcpListLine('Checking MCP server health...')).toBeNull()
+    expect(parseMcpListLine('Checking MCP server health…')).toBeNull()
+  })
+
+  it('returns null when the colon separator is missing', () => {
+    expect(parseMcpListLine('not a valid line at all')).toBeNull()
+  })
+
+  it('returns null when the status separator is missing', () => {
+    expect(parseMcpListLine('name: endpoint only')).toBeNull()
+  })
+})
+
+describe('parseMcpListLine -- claude.ai connectors', () => {
+  it('parses a Gmail connector with slug normalisation', () => {
+    const result = parseMcpListLine('claude.ai Gmail: https://gmailmcp.googleapis.com/mcp/v1 - Connected')
+    expect(result).not.toBeNull()
+    expect(result?.source).toBe('claude.ai')
+    expect(result?.name).toBe('claude.ai Gmail')
+    expect(result?.normalizedId).toBe('gmail')
+    expect(result?.endpoint).toBe('https://gmailmcp.googleapis.com/mcp/v1')
+    expect(result?.status).toBe('connected')
+  })
+
+  it('slugifies a two-word connector name', () => {
+    const result = parseMcpListLine(
+      'claude.ai Google Calendar: https://calendarmcp.googleapis.com/mcp/v1 - Connected',
+    )
+    expect(result?.normalizedId).toBe('google-calendar')
+  })
+
+  it('handles multiple spaces between words in the name', () => {
+    const result = parseMcpListLine(
+      'claude.ai Google   Drive: https://example.com/mcp - Connected',
+    )
+    expect(result?.normalizedId).toBe('google-drive')
+  })
+
+  it('parses a needs-auth status', () => {
+    const result = parseMcpListLine(
+      'claude.ai PostHog: https://mcp.posthog.com/mcp - ! Needs authentication',
+    )
+    expect(result?.status).toBe('needs_auth')
+  })
+
+  it('parses a failed status', () => {
+    const result = parseMcpListLine(
+      'claude.ai Broken: https://example.com - Failed to connect',
+    )
+    expect(result?.status).toBe('failed')
+  })
+
+  it('preserves URLs containing colons in the endpoint', () => {
+    const result = parseMcpListLine(
+      'claude.ai n8n: https://example.app.n8n.cloud/mcp-server/http - ! Needs authentication',
+    )
+    expect(result?.endpoint).toBe('https://example.app.n8n.cloud/mcp-server/http')
+  })
+})
+
+describe('parseMcpListLine -- plugin entries', () => {
+  it('parses a plugin:package:name entry to the trailing slug', () => {
+    const result = parseMcpListLine(
+      'plugin:telegram:telegram: bun run --cwd /path ... - Connected',
+    )
+    expect(result?.source).toBe('plugin')
+    expect(result?.normalizedId).toBe('telegram')
+  })
+
+  it('handles a plugin entry with distinct package and server slugs', () => {
+    const result = parseMcpListLine(
+      'plugin:acme-bundle:my-server: node /path/server.js - Connected',
+    )
+    expect(result?.source).toBe('plugin')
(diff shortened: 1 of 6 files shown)


Szotasz:
Merged manually via rebase due to conflict with #31. See ae8a693 on main.

PR #37, opened 2026-04-23
scheduler: tighten pane-state detection so mid-thinking is never seen as ready

koorbela:
## Why this matters

Live smoke testing of the PR #35 (zombie-proof startup lock) and #36 (persistent busy-retry queue) pair revealed a third failure mode in the same incident class that the operator triaged as Bug 3. The scheduler's `isSessionReadyForPrompt` relied on a single regex, `/esc to interrupt/`, to decide whether a target Claude Code tmux session could accept a new scheduled prompt. Between a user turn's submission and the spinner's first render there is a ~1-frame window where the footer shows plain `⏵⏵ bypass permissions on (shift+tab to cycle)` WITHOUT the `· esc to interrupt` suffix. A scheduler tick landing in that window mis-detected "ready", called `sendPromptToSession`, and the prompt sat in the input buffer until the post-send retry gave up after 5 Enter attempts. Reproduced cleanly in the PR #35+#36 live smoke log:

```
[11:10:40.583] INFO: Scheduled task fired task: smoke-test-11-10 agent: <main>
[11:10:57.723] WARN: Scheduled prompt still stuck after 5 Enter retries -- giving up
```

PR #36's persistent retry queue only activates on the `busy` path. When the frame-gap classified the pane as ready, the task flowed through the `fired` path and was lost -- exactly the silent-schedule failure the operator called out, just triggered by a different race than the zombie-dashboard one PR #35 fixed.

This PR closes that gap so PR #35 + #36 + this one, together, fully resolve the Bug 1 + Bug 2 + Bug 3 class.

## Change

**New module `src/pane-state.ts` (pure logic, zero imports).** Single entry point `detectPaneState(pane: string)` classifies a raw `tmux capture-pane -p` string into `'idle' | 'busy' | 'typing' | 'unknown'`. Plus `isReadyForPrompt(pane)` = `detectPaneState(pane) === 'idle'`.

Busy indicators are deliberately narrow and **lifetime-scoped**, so they disappear the moment a turn ends and do NOT starve the scheduler from stale scrollback content:

1. `/\besc to interrupt\b/` -- footer-scoped. Claude Code overwrites the footer the instant the turn completes.
2. `/\(\s*\d+s\s*·\s*↓\s*\d/` -- the tokens-down-arrow runtime counter. Load-bearing active-turn signal; only rendered while the agent is generating.
3. `/\b(?:Combobulating|Beaming|Thinking|...)…\s*\(\s*\d+s\s*·\s*↓/` -- 23 known spinner labels paired with the runtime tail on the same line. The tail requirement kills the Markdown-heading false positive (`# Thinking…` / `Step 1: Crafting…`).

Tool-use summary lines (`Searched for N patterns`, `Listed N directories`, `Read N files`) are explicitly NOT busy indicators: Claude Code leaves them rendered in scrollback after the turn ends, so matching them would classify an otherwise idle agent as busy forever and starve the scheduler -- the exact Bug-3-class silent loss this PR is supposed to eliminate.

**`src/web.ts` refactor.** The old `IDLE_FOOTER_RX` constant and inline regex logic are gone. `isSessionReadyForPrompt(session)` is now:

```ts
1. capturePane(session) -> detectPaneState -> if not 'idle', return false
2. sleep 250ms (PANE_READY_CONFIRM_DELAY_S)
3. capturePane(session) -> detectPaneState -> must also be 'idle'
```

Double-sample confirmation closes the one-frame race: if the first capture lands in the footer-gap window, the 250ms settle has us safely past the spinner's first render on the second sample. Busy paths pass through layer 1 immediately with no delay, so only the ready path pays the ~250ms cost.

**Tests: 42 vitest cases in `src/__tests__/pane-state.test.ts`.**

- Positive: `IDLE_BYPASS`, `IDLE_STRICT` (different footer variants), `IDLE_AFTER_TOOL_USE` (post-turn scrollback, must still be idle), `IDLE_WITH_SCROLLBACK_CARET` (historical ❯ in scrollback must not register as parked input).
- Busy: `BUSY_FULL_FOOTER`, `BUSY_FOOTER_FRAME_GAP` (regression for the smoke-test-11-10 bug), `BUSY_TOKENS_ONLY` (unknown spinner label but tokens tail present), `BUSY_TOOL_USE_ACTIVE` (tool summary + live spinner).
- Typing: `TYPING_PARKED` (text in input box), plus `mergeTypingAsBusy` option.
- Pending paste: `PENDING_PASTE`.
- Non-Claude: `NON_CLAUDE` returns 'unknown'.
- 23-label `it.each` covering Combobulating/Beaming/Thinking/Pondering/Reticulating/Configuring/Noodling/Ruminating/Percolating/Cogitating/Deliberating/Contemplating/Musing/Brewing/Synthesizing/Distilling/Refining/Simmering/Crafting/Formulating/Consulting/Unfurling/Unspooling/Unraveling.
- Negative regression: `does NOT classify a bare spinner-label word as busy (Markdown heading in reply text)` -- covers `# Thinking…`, `Step 1: Crafting… the plan`, `Beaming… a message through the router`.
- Edge cases: empty pane, whitespace-only, single separator, no separators, unrelated numerics (`latency 5s, count 42`).

## Scope / compatibility

- No public API changes. `isSessionReadyForPrompt` keeps the same signature and return contract (false-on-ambiguity). The 3 call sites (`web.ts:1248` agent-message router, `web.ts:1435` marveen plugin reconnect, `web.ts:1914` scheduler) all want the stricter ready semantics.
- Ready-path overhead: ~250ms sleep + a second `tmux capture-pane` round-trip (typically tens of ms). Scheduler runs on a 60s interval, so the added latency is immaterial.
- Busy-path overhead: zero -- the first sample short-circuits before the sleep.

## Test plan

- [x] `npx vitest run` -- 120/120 passing (78 prior + 42 new in `pane-state.test.ts`)
- [x] `npm run build` -- clean
- [x] **Live smoke (combined branch with this PR + #35 + #36)**: created cron task `smoke-pr3-busy` with `*/1 * * * *` targeting an agent held in a mid-thinking state. Three consecutive scheduler ticks produced the exact end-to-end behavior this series promises:
  - 11:45:44 tick: `Schedule target session busy or has pending input, will retry` -- PR3 correctly classified the mid-thinking pane as busy. Without this PR, the single-regex detector would have reported ready and the prompt would have been lost.
  - Pending row inserted: `{"id":231,"taskName":"smoke-pr3-busy","agentName":"dev2","attemptCount":1,"lastReason":"busy",...}` -- PR #36's `insertPendingTaskRetryIfNew` handled the first miss.
  - 11:46:44 tick: still busy. Pending row updated to `attemptCount=2, lastAttempt` refreshed -- PR #36's `updatePendingTaskRetry` validated.
  - Agent freed (ESC sent). 11:47:45 tick: `Scheduled task fired`. Pending queue returned to `[]` -- success-path delete validated.
- [ ] Run against a real Anthropic API outage for multi-hour stuck scenarios (can't fabricate in smoke; covered by PR #36's unit tests for the >1h alert threshold).

## Independence and merge order

Standalone -- this PR only touches `src/pane-state.ts` (new), `src/web.ts` (surgical edit of `isSessionReadyForPrompt`), and a test file. Zero conflict with #35 or #36. Any merge order works; the combined branch used for the live smoke above merged all three cleanly.

## Known limitations

- The 23-label spinner list is a snapshot of shipping Claude Code UI. A future Claude Code release that adds a new spinner label will miss the label regex until we update the list, but the tokens-down-arrow fallback (`(Ns · ↓N ...`) catches every extended-thinking turn regardless of label.
- The token-count regex will miss a Claude Code release that changes the runtime counter format. If that happens, the `esc to interrupt` footer marker still catches the busy state; the detector degrades to the old behavior rather than silently ignoring busy.
- Tool-use summary scrollback persistence means a just-idle agent whose last output was a tool call is classified idle (correct), but any same-pane output that coincidentally matches a BUSY_INDICATOR regex (e.g. a literal `(Ns · ↓N` string pasted into an inter-agent message) would classify busy until the scrollback rolls it out. Extremely unlikely in practice.

### DIFF
--- a/src/__tests__/pane-state.test.ts
+++ b/src/__tests__/pane-state.test.ts
@@ -1,0 +1,295 @@
+import { describe, it, expect } from 'vitest'
+import { detectPaneState, isReadyForPrompt } from '../pane-state.js'
+
+// Realistic pane fixtures modelled on actual `tmux capture-pane -p`
+// output from shipping Claude Code builds. Whitespace and box-drawing
+// characters (U+2500 ─, U+2771 ❯, U+23F5 ⏵) preserved exactly so the
+// regex matches exercise the same byte sequences they would in prod.
+
+const SEP = '─'.repeat(80)
+
+const IDLE_BYPASS = [
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle)',
+].join('\n')
+
+const IDLE_STRICT = [
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ? for shortcuts',
+].join('\n')
+
+const BUSY_FULL_FOOTER = [
+  '✢ Combobulating… (52s · ↓ 2.6k tokens · thinking some more)',
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt',
+].join('\n')
+
+// The smoke-test bug scenario: spinner rendered, but the footer is still
+// in its one-frame idle state before `· esc to interrupt` is appended.
+const BUSY_FOOTER_FRAME_GAP = [
+  '✢ Combobulating… (52s · ↓ 2.6k tokens · thinking some more)',
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle)',
+].join('\n')
+
+// Spinner label missing (older/newer Claude Code build). Only the
+// token-count pattern is present. Must still classify as busy.
+const BUSY_TOKENS_ONLY = [
+  '✶ (4s · ↓ 120 tokens)',
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle)',
+].join('\n')
+
+// Tool-use summary lines persist in the scrollback AFTER a turn ends --
+// Claude Code does not overwrite them. Including them as busy signals
+// would classify an otherwise idle agent as busy forever, starving
+// the scheduler. This fixture models the post-turn idle state: the tool
+// summary is on screen but no spinner, no tokens, no esc-to-interrupt.
+const IDLE_AFTER_TOOL_USE = [
+  '  Searched for 3 patterns, listed 4 directories (ctrl+o to expand)',
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle)',
+].join('\n')
+
+// Real busy-with-tool-use: spinner line present alongside the tool summary.
+const BUSY_TOOL_USE_ACTIVE = [
+  '  Searched for 3 patterns, listed 4 directories (ctrl+o to expand)',
+  '✢ Combobulating… (12s · ↓ 480 tokens)',
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt',
+].join('\n')
+
+const TYPING_PARKED = [
+  '',
+  SEP,
+  '❯ Valami amit a felhasznalo elkezdett geppelni, meg nem kuldte el',
+  SEP,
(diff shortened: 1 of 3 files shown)


Szotasz:
Merged manually via rebase due to conflict with the V3 series. See 5947bd6 on main.

PR #38, opened 2026-04-23
hardening: lock-strict npm install, tight DB file modes, request-body caps

koorbela:
## Why this matters

A four-way security audit of a running Marveen deployment (running dashboard + SQLite DB + npm tree + git history) surfaced three defects that belong together because they all concern what a single malicious input or a supply-chain compromise can reach on the operator's machine:

1. **The update path trusted the npm registry.** `update.sh` ran `npm install --silent` after `git pull`, so a hijacked transitive dep shipping a new semver-compatible version would quietly sneak in. The npm ecosystem takes such attacks several times a year (`event-stream`, `ua-parser-js`, `colors`). Every Marveen operator was one compromised dep away from arbitrary code running under their UID during the next update.

2. **The SQLite DB was world-readable.** better-sqlite3 creates the file with the process umask (typically 0o644). Any same-UID process -- malicious npm postinstall, rogue shell script, an auditable-yet-trusted dev tool -- could open `store/claudeclaw.db` for read and exfiltrate every memory, kanban card, conversation history, and agent config. The token file (`store/.dashboard-token`) and the telegram channel env are already created at 0o600 by the app, but the DB was not.

3. **Request bodies had no size cap.** A single authenticated POST of a multi-GB body would OOM the dashboard. The operator's threat model assumed localhost-only access, but a foothold anywhere on the same machine (browser extension, compromised dev tool, stolen dashboard token) was enough to crash the process.

### `update.sh`

- `npm install --silent` -> `npm ci --silent`. Byte-exact install against the committed lockfile; a registry-side compromise cannot land during an update because the hashes must match.
- New `npm audit --audit-level=high --omit=dev --silent` preflight. **Warn-and-continue**, not hard-abort: a transient network blip or an upstream CVE disclosure should not leave a half-upgraded install (new source + new node_modules + stale dist + old services). The operator sees a `FIGYELEM` line with a re-run pointer and decides.
- Both `npm ci` and `npm audit` wrapped in `if ! ... then echo ... exit 1 fi` blocks so the operator gets a human message instead of a silent non-zero exit.

### `src/db.ts`

- New `tightenDbPermissions(dbPath)` helper: chmod 0o600 on the main DB and its sidecars (`-wal`, `-shm`, `-journal`) whenever they exist. Runs on every `initDatabase()` so an older install whose files sit at 0o644 is corrected on next dashboard restart.
- `initDatabase()` now pre-creates the main DB file with `openSync(path, 'wx', 0o600)` when it doesn't exist yet, so better-sqlite3 inherits the tight mode rather than creating at umask. Closes the TOCTOU window where a same-UID attacker could `open()` the file for read between `new Database()` and the chmod.
- `initDatabase()` is now idempotent for re-entry: any previous handle is `db?.close()`'d before the new one opens, so repeated calls (tests, hot-reload, recovery paths) do not leak the previous better-sqlite3 fd.

### `src/web.ts`

- `readBody()` signature: new `{ maxBytes }` option (default 20 MB -- well above any legitimate upload, low enough to stop OOM-by-body). New `RequestBodyTooLargeError` class; the reader calls `req.destroy()` and rejects the promise once the limit is exceeded.
- `/api/schedules` POST and PUT now pass `maxBytes: 256 * 1024` to `readBody` and return 413 on overflow. Paired with a new `MAX_SCHEDULED_TASK_PROMPT_LEN = 50_000` character cap that rejects oversize prompts within a well-formed body. A 50k-char prompt is already far beyond any legitimate scheduled-task payload (real ones are usually <1 KB).
- `/api/updates/apply` spawn now redirects child stdout/stderr into `store/update.log` (0o600). The old `stdio: 'ignore'` made the new `npm audit` warning invisible in the dashboard path -- operators clicking the dashboard Update button never saw the security posture signal. The redirect lets the audit check actually function as an audit check.

### `src/__tests__/db.test.ts`

Four new vitest cases in a dedicated `database file permissions` describe block. Crucially, the block's `beforeAll` first **loosens** every sidecar to 0o644 and then re-calls `initDatabase()` -- so the tests verify enforcement (not just observation). If `tightenDbPermissions` were removed from initDatabase, these tests would fail.

## Scope / compatibility

- **No public API changes.** All modifications are additive: new options on internal helpers, new validation responses, tightened file modes.
- **Happy path unchanged** for non-schedule endpoints (their readBody calls pick up the 20 MB default, which exceeds every legitimate upload in the repo).
- **Lockfile requirement:** `npm ci` requires `package.json` and `package-lock.json` to be in sync. If an author forgets to commit a lock update, the update aborts with a human message pointing at `npm ci` for the detailed error. Net improvement over `npm install` silently succeeding on semver drift.

## Test plan

- [x] `npx vitest run` -- 82/82 passing (78 prior + 4 new DB-mode cases)
- [x] `npm run build` -- clean
- [x] **Live smoke: DB mode tightening.** Chmod store/claudeclaw.db{,-wal,-shm} to 0o644 to simulate a pre-PR4 install. Ran a short Node script invoking the built `initDatabase()`. All three files were back at 0o600 after init.
- [x] **Live smoke: prompt length 413.** On a combined branch containing PR #35 + #36 + #37 + this PR, POSTed a schedule with a 60,000-char prompt: `HTTP 413 {"error":"Prompt too large (60000 chars, max 50000)"}`. Matches the guard.
- [x] **Live smoke: body size cap.** POSTed a 300KB JSON body (above the 256KB schedule cap): connection reset by `req.destroy()` after the cap was reached, curl exited with connection-closed. The guard drops the stream before parsing, so no JSON.parse runs on the oversize payload.
- [ ] `update.sh` with intentionally poisoned lockfile: not exercised in this smoke (would require breaking a working lock).
- [ ] Warn-and-continue on an actual `npm audit` high-severity finding: current tree is clean (zero CVEs), so the warning path was not triggered live. Verified by code inspection + unit coverage.

## Related PRs

- #35 (zombie-proof startup lock)
- #36 (persistent busy-retry queue)
- #37 (tightened pane-state detection)

This PR is independent of all three but shares the combined smoke branch.

## Known limitations

- `npm audit` contacts the configured registry; on air-gapped or strict-network installs, the preflight will print FIGYELEM on every update. Acceptable trade-off; operators can drop the check locally if needed.
- The pre-create of the main DB file does not narrow the WAL/SHM sidecars preemptively; those are still created by the WAL pragma at process umask, then chmod'd. A same-UID attacker with microsecond timing could open those for read before the chmod, but the WAL contains only recent uncheckpointed writes -- the main file is the full-dataset target and that race is closed.
- `update.log` is not rotated; grows unbounded over many updates. Non-critical follow-up.

### DIFF
--- a/src/__tests__/db.test.ts
+++ b/src/__tests__/db.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, beforeAll } from 'vitest'
+import { statSync, existsSync } from 'node:fs'
+import { join } from 'node:path'
 import {
   initDatabase,
   getSession,
@@ -10,6 +12,7 @@ import {
   getMemoriesForChat,
   buildFtsMatchExpression,
 } from '../db.js'
+import { STORE_DIR } from '../config.js'
 
 beforeAll(() => {
   // Teszt adatbázis inicializálás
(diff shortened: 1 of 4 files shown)


Szotasz:
Merged manually via rebase due to conflict with the V3 series. See 68aab88 on main.

PR #48, opened 2026-04-26
recovery: add SIGHUP stage + Linux systemctl support

gezabenko-ai:
## Summary

- Add stage 0 (SIGHUP) to the Marveen Telegram plugin recovery escalator. Sends SIGHUP to the bun polling process via `bot.pid` before touching the Claude Code session -- cheapest possible fix for the common "MCP alive, polling dead" failure mode.
- Add Linux hard restart via `systemctl --user restart` (alongside existing macOS `launchctl`).
- Close `/mcp` dialog with Escape after soft reconnect so it doesn't block the session.
- NaN guard on pid file parse.

## Test plan

- [ ] Verify SIGHUP recovery works when Telegram polling dies but MCP stays alive
- [ ] Verify soft reconnect closes the /mcp dialog properly
- [ ] Verify hard restart uses systemctl on Linux and launchctl on macOS
- [ ] Verify NaN pid values don't crash the recovery flow

### DIFF
--- a/src/web.ts
+++ b/src/web.ts
@@ -339,17 +339,24 @@ const agentLastRestart: Map<string, number> = new Map()
 const AGENT_RESTART_GRACE_MS = 90_000
 const PLUGIN_ALERT_DEDUP_MS = 30 * 60 * 1000
 
-// Marveen recovery is a 4-stage escalator because killing the session
+// Marveen recovery is a 5-stage escalator because killing the session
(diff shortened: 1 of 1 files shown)


gezabenko-ai:
Closing: the refactor moved this code to src/web/telegram-monitor.ts. Will apply the relevant fixes directly there instead.

PR #94, opened 2026-05-11
agent-process: wire stuck-input + truncated-preamble detectors into sendPromptToSession

koorbela:
## Why this matters

PR #93 added two pure-logic detectors for stuck `tmux send-keys`
states (placeholder / verbatim parked input, and stale trust-marker
preamble). This PR wires them into `sendPromptToSession` so router
and scheduler prompts actually survive the two failure modes in
production.

Without the wire-up: the detectors live but nothing calls them, and
the observed 60% stuck-rate on byte-identical broadcast payloads
keeps eating inter-agent deliveries.

With the wire-up: a stale preamble gets `Ctrl-U` cleared before the
new send, and a parked-after-Enter buffer gets up to two follow-up
Enters with `shouldRetrySubmit()` gating each one. The retry budget
is bounded so a pathologically stuck pane gives up cleanly rather
than spinning.

## Change

Stacked on **#93**. Targets `main`, but the diff includes the
helpers from #93 -- merging this PR will not introduce them twice
if #93 lands first (GitHub auto-rebases stacked branches).

`src/pane-state.ts`:
- New pure helper `decideSubmitFollowup(pane, payloadHint, attempt, maxAttempts)`
  returning `'retry-enter' | 'done' | 'give-up'`. Keeps the loop in
  `agent-process.ts` testable without mocking tmux: snapshots and
  attempt counters in, action out.

`src/web/agent-process.ts`:
- `sendPromptToSession()` now does a pre-flight `capture-pane` plus
  `shouldClearTruncatedPreamble()` check. On a positive, it sends
  `Ctrl-U` to clear the input buffer, waits 100ms, then chunks the
  new prompt as before.
- After the trailing Enter, a post-send poll loop hands each new
  snapshot to `decideSubmitFollowup()`. Up to `SUBMIT_RETRY_MAX_ATTEMPTS`
  (= 2) extra Enters are sent with a 300ms gap. The loop terminates
  on `'done'`, on `'give-up'` (logged as a warn), or on any send-keys
  exception (also logged).
- New internal `clearInputBuffer(session)` helper isolates the
  `Ctrl-U` send-keys call so the pre-flight stays readable.
- The public signature `(session: string, text: string): void` is
  unchanged.

`src/__tests__/pane-state.test.ts`:
- 5 new test cases for `decideSubmitFollowup`: null pane (give-up),
  not-stuck pane (done), stuck pane under cap (retry-enter), stuck
  pane at cap (give-up), `maxAttempts=0` (give-up on first stuck).

## Scope / compatibility

No call-site changes. `src/web/message-router.ts` and
`src/web/schedule-runner.ts` keep working unmodified -- their
existing calls to `sendPromptToSession` now transparently gain the
pre-flight guard and the retry-Enter.

A small overlap exists: `schedule-runner.ts` already has its own
post-send re-submit loop (longer 2s initial wait + 5 retries). The
new internal loop completes within ~1 second; the scheduler loop
starts at +2 seconds, so they sequence rather than race. Collapsing
the two would be a follow-up refactor.

## Test plan

- `npm run typecheck` -- zero errors
- `npx vitest run --dir src` -- 367 / 367 passing
  - Existing 358 tests unchanged
  - 9 new tests across `decideSubmitFollowup` (5) and the
    `shouldRetrySubmit` / `shouldClearTruncatedPreamble` paths
    extended in PR #93 (already accounted for there)

Manual smoke: not included in this PR. The intended verification
is to broadcast a 600+ byte payload to several sub-agents and
confirm the stuck-rate drops from the observed 60% baseline.

## Known limitations

If `capture-pane` itself times out during the pre-flight, the
truncated-preamble check is skipped and the prompt is sent
unconditionally. This matches the prior behaviour (no guard at
all) and only fails open on the additional protection -- it does
not introduce a regression. A future hardening could fall back to
an unconditional `Ctrl-U` on capture failure, but that has its own
side-effect risk (clearing operator-typed text) and is left out of
scope here.

### DIFF
--- a/src/__tests__/pane-state.test.ts
+++ b/src/__tests__/pane-state.test.ts
@@ -1,5 +1,10 @@
 import { describe, it, expect } from 'vitest'
-import { detectPaneState, isReadyForPrompt } from '../pane-state.js'
+import {
+  detectPaneState,
+  isReadyForPrompt,
+  shouldRetrySubmit,
+  shouldClearTruncatedPreamble,
+} from '../pane-state.js'
 
 // Realistic pane fixtures modelled on actual `tmux capture-pane -p`
 // output from shipping Claude Code builds. Whitespace and box-drawing
@@ -111,7 +116,7 @@ const IDLE_WITH_SCROLLBACK_CARET = [
 
 // A pane that is not Claude Code at all (regular shell).
 const NON_CLAUDE = [
-  'zino@marveen ~ $ ls',
+  'user@host ~ $ ls',
   'README.md  src/  test/',
 ].join('\n')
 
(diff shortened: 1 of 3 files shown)


Szotasz:
Closing: rebased onto main after #93 merge, re-opening as new PR from Szotasz/marveen branch.

PR #170, opened 2026-05-26
fix(channels): make Discord plugin actually work end-to-end (tagged dev-channels flag + channel-group bootstrap)

DeliLevente99:
Follow-up to #163.

PR #163 added the Discord provider plumbing, but spawning a channels
session with `--channels plugin:discord@claude-plugins-official` makes
the plugin's MCP server fail at start (footer shows "1 MCP server
failed"), because `discord@claude-plugins-official` is not on Claude
Code's org-approved plugins allowlist. Even past that, the plugin's
outbound `reply` tool gates server-channel sends on
`access.groups[channelId]`, so a fresh install with `DISCORD_CHANNEL_ID`
set still gets "channel `<id>` is not allowlisted -- add via
`/discord:access`" until the operator runs the slash command manually.

This PR closes both gaps with two small commits:

## Commits

**`fix(channels): Discord plugin needs tagged dev-channels flag + warning auto-accept`**

The bypass flag exists, but with two non-obvious requirements:

1. The flag REQUIRES a tagged entry. Bare `--dangerously-load-development-channels` exits 1 with `entries must be tagged: plugin:<name>@<marketplace>` (verified on `claude 2.1.150`). The correct form is `--dangerously-load-development-channels plugin:<id>` and runs ALONGSIDE `--channels plugin:<id>`, not instead.
2. The flag triggers an interactive "Loading development channels" warning dialog ("I am using this for local development / Exit") before the plugin loads. Without auto-accept the session sits at the dialog forever.

Applied on both POSIX channels-session spawn paths:
- `scripts/channels.sh`: `DEVCHANNELS_FLAG="--dangerously-load-development-channels plugin:${PLUGIN_ID}"` when `CHANNEL_PROVIDER=discord`, injected into the tmux `new-session` command. Auto-accept case added to the 12-iteration dialog loop alongside trust + bypass + welcome.
- `src/web/channel-monitor.ts::resumeMarveenSession`: same tagged flag inlined into `claudeCmd` when `provider.type === 'discord'`.

Telegram + Slack unaffected -- their plugin IDs are on the default allowlist and don't trigger the warning, so `DEVCHANNELS_FLAG` stays empty for them.

**`feat(channels-discord): auto-add DISCORD_CHANNEL_ID to access.groups on boot`**

When `DISCORD_CHANNEL_ID` is set in the project `.env` (the canonical signal that the operator picked a target channel) the dashboard can add the channel-group entry itself on boot. `ensureDiscordChannelGroup` reads `~/.claude/channels/discord/access.json`, inserts the entry if missing (`requireMention: false, allowFrom: []`), and writes back. No-op when the entry is already present, when `CHANNEL_PROVIDER !== 'discord'`, when `DISCORD_CHANNEL_ID` is unset, or when `access.json` is corrupt (safer to leave alone than overwrite).

Wired into `src/index.ts` boot right before `startInviteMonitor`, so the entry is present before the first inbound Discord message can fire the plugin's gate.

## Out of scope / known gaps

- Operator-issued first-pair for DMs still goes through `/discord:access pair <code>` (unchanged behaviour).
- Dashboard web UI has no input for `DISCORD_CHANNEL_ID` yet -- it has to be edited into the project `.env` manually before the bootstrap helper can act.

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -35,6 +35,15 @@ case "$CHANNEL_PROVIDER" in
   *)        PLUGIN_ID="telegram@claude-plugins-official" ;;
 esac
 
+# Discord plugin is not on Claude Code's org-approved channels list, so
+# `--channels plugin:discord@...` makes the plugin's MCP server fail at
+# start ("1 MCP server failed" in the pane footer). The dev-channels flag
+# bypasses the allowlist, but REQUIRES a tagged entry -- bare flag exits 1
+# with "entries must be tagged". Telegram + Slack are on the default
+# allowlist and don't need this.
+DEVCHANNELS_FLAG=""
+[ "$CHANNEL_PROVIDER" = "discord" ] && DEVCHANNELS_FLAG="--dangerously-load-development-channels plugin:${PLUGIN_ID}"
+
 # Extra safety net for existing installs whose tmux server already has a
 # polluted global env -- scrub channel tokens so new child sessions don't
 # inherit them. The main agent's plugin will still load its token from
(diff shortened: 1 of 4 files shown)


DeliLevente99:
ez a változás NINCS test planban.  Csak a discord plugin forráskódjának (server.ts + SKILL.md) átolvasásával és a spawn-argumentum hibájának reprodukálásával lett validálva (claude 2.1.150). End-to-end Discord round-trip ezen a branchen nem futott — kérlek ellenőrizd merge előtt.

PR #219, opened 2026-05-31
feat(kanban): filter by assignee + owner 'what's on me' view, robust assignee chip

kleinheinczg:
Generic, deployment-agnostic kanban assignee view + a robustness fix for the assignee chip.

## Filter by assignee + owner "what's on me" view
- A **per-assignee dropdown** in the kanban toolbar focuses the board on one person's cards (case-insensitive match).
- A one-click toggle shows just the cards assigned to the **board owner** — the assignee whose `type` is `owner` — so the lead can see at a glance what is waiting on them.
- The owner is resolved **by type, never by a hard-coded name**. If a deployment has no owner-type assignee, the quick toggle hides itself and only the general per-assignee dropdown is shown.
- The filter UI is injected in JS so the kanban toolbar markup stays self-contained.

## Robust assignee chip
The card chip (`createCardEl`) and the detail view (`showCardDetail`) matched the assignee with an exact, case-sensitive `===` against the known list. A casing mismatch (e.g. a card stored as `gorcsevivan` vs the list entry `GorcsevIvan`) silently dropped the chip / showed "-- nincs --".

Now both match case-insensitively and, when the assignee is not in the list at all, render a fallback chip with the raw stored name and a neutral `.assignee-dot.unknown` dot (added to `style.css`).

## Scope
Frontend only (`web/app.js`, `web/style.css`). No backend / build changes.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -103,6 +103,11 @@ let kanbanCards = []
 let kanbanAssignees = []
 let kanbanProjects = []
 let kanbanProjectFilter = ''
+// Assignee filter for the kanban board. '' = show all. Set via the
+// assignee dropdown / "Csak Gábor" toggle injected by setupAssigneeFilter().
+// Matched case-insensitively against card.assignee so a casing mismatch
+// (e.g. card "gorcsevivan" vs list "GorcsevIvan") still filters correctly.
+let kanbanAssigneeFilter = ''
 
 const cardModalOverlay = document.getElementById('cardModalOverlay')
 const cardDetailOverlay = document.getElementById('cardDetailOverlay')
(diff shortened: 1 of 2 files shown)


Szotasz:
Koszi a kanban assignee-filter + 'what's on me' nezetet! A PR jelenleg konfliktusos a friss main-nel (DIRTY). Tudnad rebase-elni a main-re? Utana atnezem es merge-elem. Koszi!

kleinheinczg:
Rebased onto the latest `main` (resolves the conflict with #216).

The overlap was in the kanban assignee UI (`createCardEl` + `showCardDetail`), which #216 changed to render `displayName || name`. I kept **both**:
- #216's `displayName || name` rendering everywhere (chip, detail meta), value/key stays the agent id;
- this PR's robust matching (case-insensitive, raw-name fallback chip with the `.assignee-dot.unknown` dot) and the assignee filter + owner "what's on me" view.

The new "Felelős:" filter dropdown also shows `displayName || name` for consistency with #216. `node --check` clean. Ready for another look.

PR #211, opened 2026-05-29
feat(dashboard): live agent activity panel

theisszsolt:
Read-only live view of fleet activity — what each agent is doing right now (working/idle/waiting) + last lines of its Claude pane. Zsolt-requested MVP.

## What
- New endpoint `GET /api/agents/activity`: for the main agent + each sub-agent, reads the tmux pane (capturePane) and classifies it with detectPaneState (busy/typing->working, idle->idle, unknown/error), returns last ~8 non-empty tail lines.
- New dashboard page 'Aktivitás' (nav link + page): polls the endpoint every 3s, renders per-agent cards with a colored state badge + live pane tail. Poll stops on navigation away.
- CSS for the activity cards/badges/tail.

## Notes
- Read-only monitor (no control), additive (new route + new UI page), no schema changes.
- tsc clean. Live-verified: endpoint returns correct shape, shows jarvis/devy/disy states + tails.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #221, opened 2026-05-31
feat(auto-restart): per-agent scheduled session restart + live context size

kleinheinczg:
Per-agent scheduled session restart, to keep long-lived sessions lean. A long session accumulates context — every turn re-reads the whole transcript, so a big context is slower, costlier, and hits the per-session limit sooner. Restarting periodically (by default after the nightly dream consolidation — "the brain sleeps, tidies up, wakes fresh") keeps sessions fast.

## Modes
- **fresh** — drop the conversation (start without `--continue`); the speed-up.
- **continue** — keep the conversation (`--continue`) but re-spawn the process, so a tier / limit-budget refresh takes effect without losing context.

The main channels session is launchd-managed and always starts fresh (per `channels.sh`), so it restarts via `launchctl kickstart` (fresh-only); sub-agents use `restartAgentProcess({fresh})`.

## Safety
- **Idle-guard** — never restart a busy pane (mid-turn), including the **main channels session**, so a live conversation is never cut off; it defers to the next tick. Uses `detectPaneState`.
- **Seed-on-first-sight** — on the first sweep it only records "last restart = now" without acting, so a daily slot that already passed before the dashboard started does not trigger a spurious restart on boot. In-memory, so a restart at worst skips one slot — never double-fires.

## Pieces
- `src/auto-restart.ts` — pure config type + `normalizeAutoRestartConfig` + `restartDue` + `parseHHMM` (dependency-free, unit-tested: 13 cases).
- `src/web/auto-restart-store.ts` — `store/auto-restart.json` (one map keyed by agent, main orchestrator included).
- `src/web/active-model.ts` — `readContextTokensFromProjectDir`: live context size (input + cache_read + cache_creation of the last turn).
- `src/web/agent-process.ts` — `start`/`restartAgentProcess` gain `{fresh}` (omit `--continue`).
- `src/web/auto-restart-runner.ts` — 60s sweep with the guards above.
- `routes/agents.ts` + `routes/marveen.ts` — agent payload gains `autoRestart` + `contextTokens`; `PUT /api/agents/:id/auto-restart`.
- Dashboard UI (`web/index.html`, `web/app.js`) — Settings controls (toggle, mode, daily-time / every-N-hours) + a live context-size readout + **inline** explanatory text (not a tooltip): why smaller context is better, that it never interrupts a busy session, and the sleep analogy for the default DREAM timing. Generic, no per-deployment hard-coding.

## Deploy note
Once enabled for the main session, this supersedes a launchd nightly-restart-of-the-channels-session quick-version some deployments run (e.g. a 03:00 cron job) — disable that to avoid a double restart.

## Tests
`npm test`: 13 new pure-logic cases green. `tsc --noEmit` + `npm run build` clean.

### DIFF
--- a/src/__tests__/auto-restart.test.ts
+++ b/src/__tests__/auto-restart.test.ts
@@ -1,0 +1,78 @@
+import { describe, it, expect } from 'vitest'
+import {
+  parseHHMM,
+  normalizeAutoRestartConfig,
+  restartDue,
+  dailyDueAtMs,
+  DEFAULT_AUTO_RESTART,
+} from '../auto-restart.js'
+
+describe('parseHHMM', () => {
+  it('parses valid times to minutes since midnight', () => {
+    expect(parseHHMM('00:00')).toBe(0)
+    expect(parseHHMM('03:00')).toBe(180)
+    expect(parseHHMM('23:59')).toBe(23 * 60 + 59)
+    expect(parseHHMM('9:30')).toBe(570)
+  })
+  it('rejects malformed or out-of-range values', () => {
+    for (const bad of ['', '3', '24:00', '12:60', '-1:00', 'aa:bb', '12:5', 12 as unknown, null]) {
+      expect(parseHHMM(bad as unknown)).toBeNull()
+    }
+  })
+})
+
+describe('normalizeAutoRestartConfig', () => {
+  it('returns safe defaults for junk input', () => {
+    expect(normalizeAutoRestartConfig(null)).toEqual(DEFAULT_AUTO_RESTART)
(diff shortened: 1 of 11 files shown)


kleinheinczg:
Rebased onto the latest `main` (resolves the conflict). The only collision was in `src/web/routes/agents.ts` imports — the activity panel (#211) added `detectPaneState`/`MAIN_CHANNELS_SESSION` while this PR added `readContextTokensFromProjectDir` + the auto-restart-store imports; merged both. The 11-file diff is unchanged otherwise. `tsc --noEmit` + 13 auto-restart unit tests green. Now mergeable/clean.

PR #226, opened 2026-06-01
feat(channels): durable Telegram-channel stability + deafness recovery hardening

kovesdan:
## Context

The Telegram channels layer could silently go "deaf": the bun bridge stays up and polls, but inbound messages stop being injected into the session. It could also enter false-positive respawn loops that killed live conversations — the keep-alive freshness signal degraded from the very busyness it was meant to measure (a busy session skips the scheduled keep-alive, the file ages, and the watchdog respawned a perfectly healthy session in an idle gap). This PR hardens the channel-watchdog / recovery layer.

## Changes

- **Keep-alive staleness watchdog**: a real inbound message advances the keep-alive timestamp, so an actively-used session is never judged stale — fixing the false-positive respawn that killed live conversations. A genuinely silent/deaf session still ages out and recovers.
- **Safe respawn**: respawn-pane replaces only the `claude` process in the main channels pane; it never restarts the shared tmux server (which would kill every agent session). Startup/restart grace windows prevent kill-loops on freshly-started sessions.
- **Recovery ladder**: soft reconnect → save → resume → hard restart, with identity (`/name`) re-applied after respawns; the main-agent restart is routed through the channels helper.
- **Optional active inbound probe** (see note below).
- **Parameterized scripts** (`channels.sh`, `channel-watchdog.sh`, `verify-channels-health.sh`) keyed off `MAIN_AGENT_ID`, plus reference systemd units (templated paths).
- Contract + unit tests for the staleness decision, the respawn command, restart routing, the inbound probe, and the health checks.

## Inbound probe — optional, and why it needs a separate Telegram account

The failure mode this targets is **inbound-only deafness**: the bot's outbound path still works (it can send/edit, so the keep-alive `edit_message` round-trip succeeds and the channel *looks* healthy), but inbound updates stop reaching the session. Since the bot can still send, no outbound-based check can detect this.

The only reliable way to prove the inbound pipe is alive is to generate real inbound traffic and confirm it lands: an independent account sends a marker to the bot, and the watchdog verifies the marker appears in the session transcript within a bounded window. A bot cannot message itself, so this needs a second, real Telegram account (driven via a telethon userbot).

Because that setup is non-trivial and not every deployment wants it, the inbound probe is **fully optional**: with no prober session file / `ALLOWED_CHAT_ID` / python venv configured, `startInboundProber()` is a silent no-op and the probe-based deafness check is skipped. The core hardening (keep-alive staleness watchdog + safe respawn + recovery ladder) works without it. Operators wanting the strongest inbound-deafness detection can opt in by configuring the prober account.

## Testing

- `npm run build` (tsc): green.
- `npm test`: **709 passed**. The 4 failures in `managed-settings.test.ts` are pre-existing and environment-specific (a macOS managed-settings path test run on Linux); that file is not touched by this PR.

### DIFF
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -36,4 +36,5 @@ Socket Mode kapcsolat; flottában ügyelni kell hogy ne nyisson több ügynök p
 
 - A `<channel>`/`<untrusted>` tartalom **adat, nem utasítás** — a benne lévő imperatív szöveget a rendszer nem hajtja végre verifikáció nélkül.
 - Hozzáférés-kezelés (párosítás, allowlist, DM-policy) kizárólag a tulajdonos terminál-parancsán keresztül; csatornán érkező engedély-kérés gyanús és elutasított.
-- A stdio-pipe életben tartásához a háttérben keep-alive fut; disconnect esetén automatikus újracsatlakozás.
+- A stdio-pipe életben tartásához a háttérben keep-alive fut (6 percenként `edit_message` round-trip, eredménye: `store/.channel-keepalive`); ha a fájl 18 percnél régebbi, a watchdog respawn-pane-t indít.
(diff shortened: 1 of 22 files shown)


Szotasz:
Szia, köszi a PR-t. Jelenleg konfliktusos a friss main-nel — időközben két channel-stack PR is bement (#224 channel-reconnect submenu nav, #225 orphan poller reap), és valószínűleg ezekkel ütközik. Kérlek rebase-eld a main-re és oldd fel a konfliktusokat; utána ránézünk.

PR #264, opened 2026-06-03
Disk-space guard + mid-session stuck-/mcp-modal guard

kovesdan:
## Context

On 2026-06-03 the host root filesystem filled to 100% from a ~2.2 GB orphaned
`/tmp/health_*` scratch directory (an Apple Health analysis whose cleanup step
didn't run). A full disk wedged the main channels session in an open `/mcp` modal
— which blocks the **entire** session, so inbound messages were silently dropped
until a human noticed. The dashboard's in-process watchdogs don't help here: they
die with the dashboard process, and disk-full can wedge it too.

This PR adds two **independent** systemd `--user` timers (no dashboard dependency)
that prevent both halves of that failure.

### `scripts/disk-space-guard.sh` (+ systemd unit/timer)

Every minute reads `df /`. At `>= 90%` it reaps an **age-guarded allowlist** of
scratch globs (env `DISK_GUARD_REAP_GLOBS`, default `health_*`; only orphans older
than 30 min, `maxdepth 1` under `/tmp` — a running export's recent mtime is
protected, and nothing outside the allowlist is touched). At `>= 95%` after
reaping it alerts the owner over the **direct Telegram Bot API** (the in-session
MCP plugin is dead under disk-full), at most once per hour. Thresholds are
constants; chat id and bot token come from env. All stamp/log writes are
best-effort so an ENOSPC write can't crash the guard.

### `scripts/stuck-modal-guard.sh` (+ systemd unit/timer)

Every minute classifies the main channels session pane (`${MAIN_AGENT_ID}-channels`),
mirroring the markers in `src/pane-state.ts`:

- **idle** (`? for shortcuts` / `bypass permissions on`) or **busy**
  (`esc to interrupt` / token counter) -> healthy, never touched.
- **stuck** (neither -- a modal overlay hides the idle footer and there is no live
  turn) -> only after it **persists 120 s** (>=2 ticks): Escape up to 4x (the
  `channels.sh ensure_modal_closed` pattern), then `respawn-pane` if Escape can't
  reach the prompt. The respawn **shares `channel-watchdog.sh`'s
  `.channel-last-respawn` grace stamp**, so the two watchdogs never double-respawn.

A legitimately working session (`esc to interrupt` + `bypass permissions on`)
classifies busy and is never disturbed -- locked by a contract test.

Both scripts are deployment-neutral: session name, chat id, and bot-token source
come from env/config (no hardcoded paths or ids). The systemd unit files use
`/path/to/marveen` + `/home/USER` placeholders, matching the existing
`channel-watchdog.service`.

## Testing

Pure-logic contract tests with captured-pane / `df` fixtures (no real `df`,
Telegram, or `rm` of real scratch):

```bash
bash scripts/__tests__/disk-space-guard.test.sh    # 12: thresholds, age-guard, allowlist, alert+cooldown, malformed no-op
bash scripts/__tests__/stuck-modal-guard.test.sh   # 14: classify fixtures (idle/busy/stuck/empty), false-positive guard, confirm-window
```

Both green; `npx tsc --noEmit` clean (no TypeScript touched).

> Companion hardening (intentionally not in this PR): existing watchdog stamp
> writes can be made best-effort for full ENOSPC tolerance.

## Known pre-existing inconsistency (not changed here)

The sibling `scripts/channel-watchdog.sh` resolves its target session with a `${MAIN_AGENT_ID:-marveen}` default. The two new guards in this PR are intentionally config-driven with **no baked-in default** — `CHANNELS_SESSION`, else `${MAIN_AGENT_ID}-channels` from `.env`, else skip — per a review hardening requirement. Aligning `channel-watchdog.sh` is a separate future cleanup, out of scope for this PR.

### DIFF
--- a/docs/disk-modal-guards.md
+++ b/docs/disk-modal-guards.md
@@ -7,10 +7,10 @@ modal and left inbound messages dropped until a human noticed. They are
 **independent of the dashboard** (which dies with its process and is itself
 unreliable under disk-full).
 
-## A. Disk-space guard — `scripts/disk-space-guard.sh`
+## A. Disk-space guard -- `scripts/disk-space-guard.sh`
 
 Every minute: read `df /` usage.
-- `>= 90%` (`REAP_THRESHOLD`): reap **age-guarded allowlist scratch** — globs in
+- `>= 90%` (`REAP_THRESHOLD`): reap **age-guarded allowlist scratch** -- globs in
   `REAP_GLOBS` (currently `health_*`) directly under `/tmp`, only entries older
   than `REAP_MIN_AGE_MIN` (30 min), so a currently-running export (recent mtime)
   is never deleted.
@@ -20,7 +20,7 @@ Every minute: read `df /` usage.
 Thresholds + the reap allowlist are constants at the top of the script. All
 stamp/log writes are best-effort (ENOSPC-tolerant).
 
-## B. Stuck-modal guard — `scripts/stuck-modal-guard.sh`
+## B. Stuck-modal guard -- `scripts/stuck-modal-guard.sh`
 
 Every minute: classify the main channels session pane (`${MAIN_AGENT_ID}-channels`,
 mirrors `src/pane-state.ts`):
@@ -41,7 +41,7 @@ classified busy/idle and is never disturbed (locked by the contract test).
 
 ## Activation
 
-The unit files carry `/path/to/marveen` and `/home/USER` placeholders — replace
+The unit files carry `/path/to/marveen` and `/home/USER` placeholders -- replace
 them with your install dir and home before installing.
 
 ```bash
@@ -64,11 +64,11 @@ bash scripts/__tests__/stuck-modal-guard.test.sh   # classify fixtures (idle/bus
 
 ## Re-review hardening (2026-06-03, post-cross-model)
 
-- **W2a — busy-marker glyph robustness:** the `(Ns · ↓` token-counter separator can
+- **W2a -- busy-marker glyph robustness:** the `(Ns · ↓` token-counter separator can
   render as a Unicode middle-dot OR an ASCII period depending on terminal/locale.
   `classify_pane` now matches both (`(·|\.)`), so a working pane is never misread as
   STUCK and respawned mid-turn. Locked by a regression fixture in the test suite.
-- **W2b — respawn plugin id is config-overridable:** `STUCK_MODAL_PLUGIN` (default
+- **W2b -- respawn plugin id is config-overridable:** `STUCK_MODAL_PLUGIN` (default
   `plugin:telegram@claude-plugins-official`) so a renamed/local-build install isn't
   respawned with a wrong plugin id (which would exit immediately while the alert
   falsely says "respawned"). `%q`-quoted at interpolation, like the model id.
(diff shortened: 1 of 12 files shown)


Szotasz:
Reviewed -- this is a solid, well-documented PR. The pane classification grounded in `src/pane-state.ts`, the shared `.channel-last-respawn` grace stamp (so it won't fight `channel-watchdog.sh`), the age-guarded /tmp reap allowlist, and the disk-full-tolerant best-effort writes are all exactly right. We've retargeted the base to `develop`.

One thing holding the merge: the PR is still marked **Draft**. Since it's your contribution, we don't want to flip it to Ready ourselves and override your intent -- if you consider it final, please mark it **Ready for review** and we'll merge it onto develop.

Heads-up on a follow-up we'll handle on our side: our production host is macOS (launchd), not systemd, and the 2026-06-03 disk-full incident this guards against happened on that Mac host -- so we'll add a launchd variant of the two timers so it actually protects us too. Your scripts are portable; it's just the unit wiring. Thanks for the careful work.

attila-fiREG:
Reviewed both guards — really solid, the risky parts are carefully bounded.

**disk-space-guard.sh** (it deletes files, so I looked hard here): reaping is allowlist-only globs, with a hard real-path guard that requires the resolved dir to live under `/tmp` or `$XDG_RUNTIME_DIR` (rejects a symlinked scratch dir and prefix tricks like `/tmpfoo`), `maxdepth 1`, plus the age guard. The directory case is protected against deleting an in-progress export by checking inner-file mtime rather than the dir's own mtime, and the malformed/`0` age-env → 1440 fallback closes the "reap everything" footgun. Nicely defensive.

**stuck-modal-guard.sh**: only acts after STUCK persists ≥ `STUCK_SECONDS`, never touches a healthy or busy pane (so it won't drop a live reply), escalates gently (Escape → respawn-pane), and coordinates with the existing respawn stamp to avoid a double-respawn storm. The pure `classify`/`decide` split makes it testable without tmux.

Tests + docs are included, so it meets the contribution bar.

A couple of minor things:
- It's still marked **draft** — is it ready for review, or are you still iterating?
- The `web/app.js` + `style.css` status-badge/tooltip changes feel like a separate concern bundled in here; splitting them into their own PR would keep the history cleaner, though they're small and related.

Nice work.

kovesdan:
Thanks both — @Szotasz for the base retarget to `develop` and the merge offer, and @attila-fiREG for going hard on the reaping path. Glad the safety bounds (allowlist-only globs, the realpath `/tmp` / `$XDG_RUNTIME_DIR` guard that rejects symlinked scratch dirs and `/tmpfoo` prefix tricks, `maxdepth 1`, the inner-file-mtime check, and the malformed/`0` age → 1440 fallback) all landed as intended.

This is final on my side — I've **marked it Ready for review**.

On the split point: you're right that the `web/app.js` + `web/style.css` change is a separate concern — it's the agent **activity-badge** tooltips + a "breathing" pulse on the `dolgozik` badge, unrelated to the disk/stuck-modal guards. Since it's small (+37/-11), already reviewed clean, and you're ready to merge the whole, I've left it in for this merge to avoid churn — but I'll keep mixed concerns in their own PR going forward. Happy to peel it out into a follow-up PR instead if you'd prefer the cleaner history before merging — just say the word.

attila-fiREG:
Thanks @kovesdan — glad the reaping path and the safety bounds (allowlist globs, the realpath /tmp + $XDG_RUNTIME_DIR guard, maxdepth 1, inner-file-mtime, malformed/0 → 1440 fallback) all held up. Since it's already merged, no need to peel the activity-badge change out retroactively — agreed on keeping mixed concerns in their own PRs going forward. Thanks for the clean collaboration!

PR #272, opened 2026-06-03
feat(channels): gate respawn/keep-alive to a single host

buglime:
## Problem

When the same checkout runs on more than one machine (e.g. a developer's laptop *and* the production host), each machine independently respawns the main agent, auto-restarts sub-agents, and boots the heartbeat sub-agent. The two boxes then fight over the same bot tokens / `getUpdates` slot (409 Conflict loops), and a dev machine can clobber the live fleet.

## Fix

Gate the respawn/keep-alive machinery behind `RESPAWN_ENABLED`:

- `RESPAWN_ENABLED` — `1`/`true` forces on, `0`/`false` forces off.
- `RESPAWN_HOST` — optional substring matched against `os.hostname()`; when set, respawn is enabled only on a host whose name matches.
- Default (neither set) — **enabled**, so a single-host install needs no configuration and upstream behaviour is unchanged.

When disabled:
- `startChannelPluginMonitor()` logs once and returns `null` (the whole monitor loop is skipped); the shutdown path guards `clearInterval` for the nullable return.
- the boot-time heartbeat sub-agent start is skipped (it would otherwise fight the production host over the channel).

Multi-host operators set `RESPAWN_ENABLED=0` on the dev box (or `RESPAWN_HOST=<prod-hostname>` everywhere) via their `.env` — no host-specific values in the code.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,3 +1,4 @@
+import { hostname } from 'node:os'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { readEnvFile } from './env.js'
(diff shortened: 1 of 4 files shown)


Szotasz:
Lezárva: a single-host respawn-gate már a develop-on van (9910572). Ráadásul ez a branch egy régebbi develop-ról ágazik, így a merge visszacsinálna újabb mergelt feature-öket (SSE token-auth #275, reauth-healer #276, taskstate #277). A funkció lényege már él a develop-on -- köszi!

PR #273, opened 2026-06-03
feat(channel-monitor): desired-state auto-recovery for sub-agents

buglime:
## Problem

Sub-agents run on a tmux server **shared** with the main channels session. That unit gets `systemctl restart`ed several times a day (a bun-watchdog restarts it when the main agent's telegram bun child dies; the stuck-tool-call watchdog / hard restart also do). Because the channels unit is `KillMode=control-group`, each restart kills the whole shared tmux server and takes **every sub-agent** down at once.

The monitor previously only auto-restarted agents whose session still existed with a dead plugin — agents whose session vanished entirely stayed down until a manual start. Net effect on a fleet: several times a day everything except the main agent silently dies and never comes back.

## Fix

Track operator intent and reconcile it:

- `store/agents-desired.json` — written by the `/start` (add) and `/stop` (remove) routes, so a deliberately stopped agent is **not** resurrected.
- The channel monitor reconciles each tick: any desired-but-missing agent is restarted, **staggered ~15s apart** (starting several channel agents simultaneously makes them all die in the resume-from-summary modal race), one in-flight burst at a time.

This also covers machine reboot — the first monitor tick brings the fleet back.

## Verification

On a production host, after a `*-channels` restart nuke (0/5 agents), the fleet auto-recovered to 5/5 in ~85s and stayed up; deliberately stopped agents were not restarted.

### DIFF
--- a/src/web/agent-desired-state.ts
+++ b/src/web/agent-desired-state.ts
@@ -1,0 +1,54 @@
+import { existsSync, readFileSync, writeFileSync } from 'node:fs'
+import { join } from 'node:path'
+import { STORE_DIR } from '../config.js'
+import { logger } from '../logger.js'
+
+// Desired run-state for sub-agents.
+//
+// The agents run on a tmux server shared with the main channels session, which
+// gets `systemctl restart`ed several times a day (bun-watchdog, stuck-tool-call
+// watchdog, hard-restart). Because that unit is KillMode=control-group, every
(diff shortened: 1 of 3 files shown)


Szotasz:
Lezárva: superseded — a desired-state auto-recovery már a develop-on van (a19afab, azonos `agent-desired-state.ts`). Köszi a hozzájárulást! A funkció él a develop-on.

PR #280, opened 2026-06-04
heartbeat: make the dedicated heartbeat agent opt-in and config-driven

koorbela:
## Why this matters

The dedicated channel-less `heartbeat` sub-agent auto-starts on every dashboard boot whenever the respawn gate is on (the single-host default). Two consequences fall out of that:

1. A fresh or upgrading install silently spawns a sub-agent that reads the operator's calendar and database, with no opt-in.
2. The scaffolded `agents/heartbeat/CLAUDE.md` bakes in one deployment's identity: a personal Google Calendar address, an absolute `/Users/<name>/...` store path, and hardcoded owner / main-agent names. Every other install then inherits those values on disk and re-renders them on each boot, so the agent points at someone else's calendar and a non-existent absolute path.

## Change

- New `HEARTBEAT_AGENT_ENABLED` env, **default off**. The boot block in `src/index.ts` now starts the heartbeat only when `shouldBootHeartbeatAgent({ respawnEnabled, agentEnabled })` is true, i.e. the host owns the respawn gate AND the heartbeat is explicitly enabled. The gate is a small pure, unit-tested function.
- `renderHeartbeatClaudeMd(id)` is now a pure function fed by `currentHeartbeatIdentity()`, which pulls the owner name (`OWNER_NAME`), the main-agent name/id (`BOT_NAME` / `MAIN_AGENT_ID`), the absolute store dir (`STORE_DIR`), the dashboard origin (`WEB_PORT`) and an optional `HEARTBEAT_CALENDAR_ACCOUNT` from config. No personal address, home-dir path or hardcoded name remains in the shipped scaffold; an empty calendar account renders "your primary calendar" with no address at all.

Files: `src/config.ts`, `src/index.ts`, `src/web/heartbeat-agent-scaffold.ts`, `src/__tests__/heartbeat-agent-scaffold.test.ts`.

## Scope / compatibility

- **Behaviour change for installs that relied on the auto-started heartbeat:** with the new default-off gate they set `HEARTBEAT_AGENT_ENABLED=1` on the respawn host to keep it (and `HEARTBEAT_CALENDAR_ACCOUNT=<account>` to pin the calendar).
- The heartbeat's runtime contract is unchanged: it still never calls a channel send tool, still hands its summary to the main agent over inter-agent message, and stays headless.
- Side fix: the Kanban instruction now uses the absolute store path in both the prose and the `sqlite3` command. The command was previously relative (`sqlite3 store/claudeclaw.db`) and would not have resolved from the agent's working dir.

## Test plan

- `npx tsc --noEmit` clean.
- New `src/__tests__/heartbeat-agent-scaffold.test.ts` (17 tests) covers:
  - the `shouldBootHeartbeatAgent` truth table (boots only when both gates are on; default-off holds);
  - identity threading (owner name, main-agent id in the routed message, absolute store paths, dashboard origin);
  - the set vs empty calendar-account branches;
  - regression guards that the rendered output carries no email beyond the configured account (and none at all for the empty default), no `/Users/` or `/home/` path, no leftover default identity name, and no em-dash;
  - that the hard behavioural contract (never reply/Telegram/Slack, headless) survives the rewrite.
- Full `src/` vitest suite green, except one pre-existing `managed-settings` failure that is unrelated to this change and also fails on `main`.

## Known limitations

- `HEARTBEAT_CALENDAR_ACCOUNT` is a single account; multi-calendar aggregation is out of scope.
- The dashboard origin is rendered as `http://localhost:<WEB_PORT>`. A `WEB_HOST` bound to a non-loopback address would make that wrong for a remote caller, but the heartbeat sub-agent always runs locally on the same host, so localhost is correct here.

### DIFF
--- a/src/__tests__/heartbeat-agent-scaffold.test.ts
+++ b/src/__tests__/heartbeat-agent-scaffold.test.ts
@@ -1,0 +1,136 @@
+import { describe, it, expect } from 'vitest'
+import {
+  renderHeartbeatClaudeMd,
+  shouldBootHeartbeatAgent,
+  type HeartbeatIdentity,
+} from '../web/heartbeat-agent-scaffold.js'
+
+// A fully generic identity -- no real deployment values. The renderer is
+// pure, so every operator-specific string in its output must trace back to
+// one of these fields.
+const ID: HeartbeatIdentity = {
+  ownerName: 'Nina',
+  botName: 'Helios',
+  mainAgentId: 'helios',
+  storeDir: '/srv/app/store',
+  dashboardOrigin: 'http://localhost:3420',
+  calendarAccount: 'nina@example.com',
+}
+
+describe('renderHeartbeatClaudeMd', () => {
+  it('threads the owner name into the role + hard rules', () => {
+    const out = renderHeartbeatClaudeMd(ID)
+    expect(out).toContain("across Nina's systems")
+    expect(out).toContain('you NEVER contact Nina directly')
+  })
+
+  it('names the main agent as the relay target by display name', () => {
+    const out = renderHeartbeatClaudeMd(ID)
+    expect(out).toContain('hand the result to the main agent (Helios)')
+  })
+
+  it('routes the inter-agent message to the main agent id', () => {
+    const out = renderHeartbeatClaudeMd(ID)
+    expect(out).toContain('"to":"helios"')
+    // The sender is always the fixed heartbeat agent id.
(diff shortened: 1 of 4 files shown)


Szotasz:
Merged to `develop` via 0080a91 (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #281, opened 2026-06-04
backup: derive the launchd-jobs glob from MAIN_AGENT_ID, not a hardcoded label

koorbela:
## Why this matters

`scripts/backup.sh` hardcoded the launchd-jobs backup glob to one deployment's launchd label (`com.<that-deployment>.*.plist`). On every other install that glob matches **zero** files, so the fleet's launchd jobs (`com.<MAIN_AGENT_ID>.channels`, `com.<MAIN_AGENT_ID>.dashboard`, and any others) are silently dropped from the archive. That is exactly the silent-loss failure the backup is meant to prevent: a restore from such an archive comes up with no launchd jobs and no obvious sign anything is missing.

## Change

- Resolve `MAIN_AGENT_ID` from `.env` the way the app already does (`src/env.ts`: last definition wins, surrounding matching quotes stripped; default `marveen` when unset), and glob `com.${MAIN_AGENT_ID}.*.plist`. That matches how `src/web/main-agent.ts` constructs the job labels (`com.<MAIN_AGENT_ID>.<service>.plist`).
- The `.env` read is guarded with `|| true` so that a no-match lookup cannot abort the backup under `set -euo pipefail` on an install that leaves `MAIN_AGENT_ID` at its default.
- `docs/MIGRATION.md` is updated to use the `com.<MAIN_AGENT_ID>` placeholder in the same spots (inventory, launchd install, cutover, verification checklist).

Files: `scripts/backup.sh`, `docs/MIGRATION.md`.

## Scope / compatibility

- **No behaviour change** for an install whose `MAIN_AGENT_ID` already resolves to the previously hardcoded value: it still globs the same plists.
- Installs that left `MAIN_AGENT_ID` unset fall back to `com.marveen.*.plist`, consistent with the app default.
- macOS (bsdtar / BSD userland) and Linux (GNU) both supported: the new commands use the portable `grep -E` / `sed -E` spellings and the POSIX `[[:space:]]` class.

## Test plan

- `bash -n scripts/backup.sh` clean.
- On a box whose `MAIN_AGENT_ID` is not the hardcoded one, the old glob matched **0** launchd plists while the new glob matches the real `com.<MAIN_AGENT_ID>.channels` / `.dashboard` plists; a full `backup.sh` run was confirmed to place both under `home/Library/LaunchAgents/` in the archive (and in `MANIFEST.txt`).
- `MAIN_AGENT_ID` resolution checked for the unset (-> `marveen`), set, and quoted cases, all under `set -euo pipefail` (no abort on a no-match read).
- `docs/MIGRATION.md` has no remaining hardcoded label and no em-dash on changed lines.

## Known limitations

- The shell-side `.env` parse diverges from `src/env.ts` only on pathological values (a trailing inline `#` comment after the value, inner-space-padded quotes, or an explicitly empty `MAIN_AGENT_ID=`). None of those are valid agent slugs (the installer derives the id as NFKD / ASCII / lowercase-dash), so none affect real glob matching.

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -30,7 +30,7 @@ the Docker volumes (3) are **separate** and must be moved on their own.
 - `~/.claude/scheduled-tasks/**` — file-based scheduled tasks (SKILL.md + task-config.json)
 - `~/.claude/channels/*/.env` — MAIN orchestrator channel token
 - `~/.claude/channels/*/access.json`, `invites.json`, `approved/**` — pairing allowlist + approvals
-- `~/Library/LaunchAgents/com.gorcsevivan.*.plist` — launchd jobs
+- `~/Library/LaunchAgents/com.<MAIN_AGENT_ID>.*.plist` -- launchd jobs (the prefix is your `MAIN_AGENT_ID`, `marveen` by default)
 
 **(3) Docker volumes — NOT in the tarball, migrate separately**
 - `stack_influxdb-data`, `stack_influxdb-config` — InfluxDB 2.7 time-series (Loxone history)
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged to `develop` via 920cecb (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #283, opened 2026-06-05
fix(db): wrap deleteKanbanCard in a transaction with child-orphan nullification

kovesdan:
## Problem

`deleteKanbanCard` issues two independent DML statements (not wrapped in a
transaction):

```
src/db.ts, lines 1075-1078:
export function deleteKanbanCard(id: string): boolean {
  db.prepare('DELETE FROM kanban_comments WHERE card_id = ?').run(id)
  return db.prepare('DELETE FROM kanban_cards WHERE id = ?').run(id).changes > 0
}
```

Two failure modes:

**1. FK constraint crash (latent, activates when `PRAGMA foreign_keys = ON`)**

The schema at `db.ts:164` defines:

```
parent_id TEXT REFERENCES kanban_cards(id)
```

With FK enforcement on, deleting a parent card while any child still has
`parent_id = <deleted id>` raises `SQLITE_CONSTRAINT_FOREIGNKEY`.
better-sqlite3 defaults FK enforcement to OFF, so this is latent — but
enabling FK enforcement (a standard hardening step) would expose it
immediately on any delete of a parent card.

**2. Silent data corruption (present even with FK off)**

Children with a dangling `parent_id` survive the delete but cannot be reached
via the parent relationship. They do not appear under any parent in hierarchy
views. This is a silent data bug regardless of FK enforcement.

Additionally: the two statements are not atomic. A crash between the two
leaves comments deleted but the card intact, or (less likely but possible
under concurrent load) partial state.

## Fix

Wrap all three mutations in `db.transaction()` in FK-safe order:

1. `DELETE kanban_comments WHERE card_id = ?` — removes comment FK refs
2. `UPDATE kanban_cards SET parent_id = NULL WHERE parent_id = ?` — promotes
   children to root-level cards rather than leaving dangling references
3. `DELETE kanban_cards WHERE id = ?`

The `'ongoing'` status migration that appears in our internal fork is
**not included** — that is a fork-specific kanban feature. The offerable
nucleus is the 6-line `deleteKanbanCard` body only.

## Verification

- **4 contract tests** added to `src/__tests__/kanban-delete-fk.test.ts`.
  All call the real `deleteKanbanCard` entry point against an in-memory
  database seeded with the production schema via `initDatabase(':memory:')`:
  1. Deletes a solo card (no comments, no children) → returns `true`
  2. Returns `false` for a non-existent card ID
  3. Deletes card AND its comments atomically (comments gone after delete)
  4. Nullifies `parent_id` on children instead of leaving dangling references
- **Mental-revert evidence**: reverting to the pre-fix two-statement version
  causes test 4 to fail: `getKanbanCard('child-1').parent_id` returns
  `'parent-1'` (dangling reference) instead of `null`.
- **Full upstream suite**: 931/931 passes (pre-existing 4 `managed-settings`
  failures on Linux unrelated; +4 new tests from this branch).

## Origin

The upstream `deleteKanbanCard` at 5b03bf4 (db.ts:1075) is byte-identical
to the pre-fix version.

---

### DIFF
--- a/src/__tests__/kanban-delete-fk.test.ts
+++ b/src/__tests__/kanban-delete-fk.test.ts
@@ -1,0 +1,71 @@
+// Contract tests for deleteKanbanCard transactional FK safety.
+//
+// Root cause: the pre-fix deleteKanbanCard issued two independent DML
+// statements (DELETE comments, DELETE card). When foreign-key enforcement
+// is enabled (`PRAGMA foreign_keys = ON`), deleting a parent card that
+// still has children with a non-null parent_id fails with
+// SQLITE_CONSTRAINT_FOREIGNKEY. Even with FK off (the better-sqlite3
+// default), orphaned children with a dangling parent_id are a data bug:
+// they do not appear under any parent in hierarchy views and cannot be
+// reached via the parent relationship.
+//
+// Fix: wrap all three mutations in db.transaction():
+//   1. DELETE comments referencing the card.
+//   2. UPDATE children: SET parent_id = NULL (promote to root-level).
+//   3. DELETE the card.
+//
+// These tests call deleteKanbanCard (the real production entry point) on
+// an in-memory database seeded with the production schema.
+
+import { describe, it, expect, beforeEach } from 'vitest'
+import { initDatabase, createKanbanCard, getKanbanCard, addKanbanComment, getKanbanComments, deleteKanbanCard } from '../db.js'
+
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged to `develop` via 6894478 (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #284, opened 2026-06-05
fix(message-router): never abandon a report to an alive-but-busy session

kovesdan:
## Problem

`startMessageRouter()` checks `ageMs > MESSAGE_ABANDON_WINDOW_MS` at the
**top** of the loop, **before** the session-existence check:

```
src/web/message-router.ts, lines 59-67 at 5b03bf4:
      const ageMs = now - msg.created_at * 1000
      if (ageMs > MESSAGE_ABANDON_WINDOW_MS) {
        logger.warn({ ... }, 'Agent message abandoned: target never ready within window')
        if (!markMessageFailed(msg.id, 'Abandoned: target session never ready within retry window')) { ... }
        routerLoggedMisses.delete(msg.id)
        continue
      }
      // ...
      let sessionExists = false
      try { ... } // ← existence check happens only AFTER abandon
```

The loop comment at line 44 even admits "(... / stays busy) is marked
failed" — the code matched the comment, but the comment described a bug.

**Effect**: a message to the main session that has been continuously alive
but busy (dense heartbeats, a long coding turn, or a sub-agent running for >1h)
is marked `status='failed'` at the 1h mark even though the session was running
the entire time. The sender gets no delivery confirmation; the message is
silently lost from the queue.

**Incident**: two sub-agent completion reports were discarded while the main
session was continuously alive.

## Fix

Gate the abandon decision on **both** conditions:
1. The session is **absent** (`sessionExists === false`)  AND
2. The message has been waiting past `MESSAGE_ABANDON_WINDOW_MS`

Extract the decision into a `shouldAbandon(sessionExists, ageMs, windowMs)`
pure function (exported for testing) and use it in the loop. A message to a
session that EXISTS is never abandoned — it keeps retrying on every 5s tick and
will be delivered in the next idle gap regardless of how long the session has
been busy.

The session-existence check is moved **before** the abandon gate — this is the
structural fix.

## Verification

- **4 contract tests** in `src/__tests__/router-abandon.test.ts`. All call
  `shouldAbandon` directly (the real production entry point):
  1. Returns `false` when session **exists**, regardless of age (the bug case)
  2. Returns `false` when session absent but **within** the window
  3. Returns `true` when session absent AND **past** the window
  4. Returns `false` at the exact window boundary (strict `>`, not `>=`)
- **Mental-revert evidence**: reverting `shouldAbandon` to the pre-fix logic
  (`return ageMs > windowMs` — ignoring `sessionExists`) causes test 1 to fail:
  `Expected: false, Received: true` for `shouldAbandon(true, window+1, window)`.
- **Full upstream suite**: 931/931 passes (pre-existing 4 `managed-settings`
  failures on Linux unrelated; +4 new tests from this branch).

## Origin

Change is self-contained to `src/web/message-router.ts`; no fork-specific
logic included.

---

### DIFF
--- a/src/__tests__/router-abandon.test.ts
+++ b/src/__tests__/router-abandon.test.ts
@@ -1,0 +1,47 @@
+// Contract tests for shouldAbandon: router never drops a report to a
+// busy-but-alive session.
+//
+// Root cause: the pre-fix startMessageRouter() checked `ageMs > window`
+// BEFORE the session-existence check (message-router.ts:59-67). So a
+// message sent to the main session, which stays alive but busy (dense
+// heartbeats or a long turn), was marked "failed" at the 1h abandon mark
+// even though the session never actually went away. Two completion reports
+// were silently discarded in the incident that exposed this.
+//
+// Fix: abandon ONLY when the session is ABSENT for the full window.
+// shouldAbandon(sessionExists, ageMs, windowMs) is the pure decision
+// function extracted from the loop body; the contract tests pin it.
+
+import { describe, it, expect } from 'vitest'
+import { shouldAbandon } from '../web/message-router.js'
+
+const WINDOW_MS = 60 * 60 * 1000 // 1 hour, same as MESSAGE_ABANDON_WINDOW_MS
+
+describe('shouldAbandon: abandon only when session is absent past the window', () => {
+  it('returns false when session exists regardless of age', () => {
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged to `develop` via 1723d5d (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #285, opened 2026-06-05
fix(pane-state): scope esc-to-interrupt busy-check to live footer region

kovesdan:
## Problem

`detectPaneState()` scans `BUSY_INDICATORS` against the **full pane string**.
The first entry, `/\besc to interrupt\b/`, is present in Claude Code's footer
during an active turn — but the same phrase can also appear verbatim in
scrollback content: watchdog reports, log excerpts, or a tool-call output that
quoted the phrase.

When that happens, `detectPaneState` returns `'busy'` on every poll, the
scheduler skips injection, and the session is effectively locked out of
receiving new messages until the scrollback clears or the session restarts.
**Observed: 81 consecutive scheduler retries on a session whose scrollback
contained the phrase in a log line.**

Location in the code at 5b03bf4:

```
src/pane-state.ts, line 76 — BUSY_INDICATORS array, first entry:
  /\besc to interrupt\b/

src/pane-state.ts, lines 216-218 — detectPaneState busy-scan loop:
  for (const rx of BUSY_INDICATORS) {
    if (rx.test(pane)) return 'busy'  // ← scans the WHOLE pane string
  }
```

Note that `detectsThinkingBlockError` already scopes its check to the live
tail via `ERROR_LIVE_TAIL_LINES`, and the module comment on `BUSY_INDICATORS`
says "`esc to interrupt` is the footer-scoped fallback" — but the code did not
enforce the scope.

## Fix

- Remove `/\besc to interrupt\b/` from the whole-pane `BUSY_INDICATORS` loop.
- Introduce `BUSY_ESC_TO_INTERRUPT_RX` and `LIVE_FOOTER_REGION_LINES = 5`.
- After the `BUSY_INDICATORS` loop, check `BUSY_ESC_TO_INTERRUPT_RX` only
  against the bottom `LIVE_FOOTER_REGION_LINES` lines — the region where
  Claude Code actually appends `· esc to interrupt` during a live turn.
  That region cannot contain deep-scrollback prose quotes.
- Apply the same footer-region guard in `shouldRetrySubmit` (which uses the
  same `BUSY_INDICATORS` loop for its busy-guard), keeping both entry points
  consistent.

The change is ~15 lines in `src/pane-state.ts`. No other files touched.

## Verification

- **Ported contract tests** added to `src/__tests__/pane-state.test.ts`:
  1. Idle when `"esc to interrupt"` appears only in scrollback prose → `'idle'`
     (was: `'busy'`, the bug)
  2. Busy when the phrase appears on the real footer line → `'busy'` (normal
     live-turn case still detected)
- **Full upstream suite**: 121/121 tests in `pane-state.test.ts` pass; 927/927
  total suite passes (pre-existing 4 `managed-settings` failures on Linux are
  unrelated macOS-path tests present before this change).
- **Mental-revert evidence**: reverting the fix (putting `/\besc to interrupt\b/`
  back into `BUSY_INDICATORS` and removing the footer-region check) causes
  test 1 to fail with `Expected: 'idle', Received: 'busy'`.

## Origin

Found and fixed in production on a fork; no external incident details.
The fix is self-contained in `src/pane-state.ts` with no fork-specific
dependencies.

---

### DIFF
--- a/src/__tests__/pane-state.test.ts
+++ b/src/__tests__/pane-state.test.ts
@@ -1238,3 +1238,44 @@ describe('decideStuckInputRecovery', () => {
     expect(d.next.attempts).toBe(0)
   })
 })
+
+// ---------------------------------------------------------------------------
+// Contract tests: esc-to-interrupt live-region scoping (port from kovesdan/marveen)
+//
+// Root cause: a watchdog report or log output that quotes "esc to interrupt"
+// anywhere in the scrollback permanently classified an otherwise-idle session
+// as busy (81-retry starvation incident). The fix scopes the phrase check to
+// the bottom LIVE_FOOTER_REGION_LINES of the pane.
+// ---------------------------------------------------------------------------
+describe('detectPaneState: esc-to-interrupt scoped to live footer region', () => {
+  const SEP_R = '─'.repeat(80)
+
+  it('classifies as idle when "esc to interrupt" appears only in scrollback prose', () => {
+    // A watchdog report or tool-call output that QUOTES the phrase somewhere
+    // above the live input box. With whole-pane scanning this would pin the
+    // session as busy forever; scoped to the footer region it is correctly idle.
+    const pane = [
+      '  [watchdog report]: session was busy, waiting for esc to interrupt signal',
+      '  (scrollback content continues)',
+      '',
+      SEP_R,
+      '❯ ',
+      SEP_R,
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged to `develop` via 97b33fc+377b8f6 (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #287, opened 2026-06-05
fix(web): remove startMcpListChecker() boot warmup to prevent Telegram 409

kovesdan:
## Problem

`startWebServer()` calls `startMcpListChecker()` unconditionally at startup
(30s delay). That function spawns `claude mcp list`, which starts the Telegram
plugin for a health check. The plugin process claims the bot-token long-polling
slot, which sends an **HTTP 409 Conflict** to the already-running
session-bridge process holding the same token — taking the live Telegram
channel offline.

Location in the code at 5b03bf4:

```
src/web.ts, lines 258-263:
  // Warm the MCP list cache so the Connectors page reflects claude.ai OAuth
  // connectors on first load. 30s delay lets the main-channels session settle
  // first so the telegram plugin's single-poller token is claimed before
  // `claude mcp list` spawns it for a health check.
  startMcpListChecker()
  logger.info('MCP list cache warmup scheduled (30s delay, manual refresh only)')
```

The comment itself acknowledges the race ("30s delay lets the main-channels
session settle first") but the 30s delay is not sufficient — the session-bridge
may still be mid-reconnect at that point.

**Relationship to PR #269:** #269 fixed a different 409 source — the channel-
coordinator runtime poller-flapping (cooldown hysteresis + detached-claude
reap). This fix addresses the **boot-time plugin-spawn** 409 vector. The two
fixes are complementary; both vectors need addressing independently.

## Fix

Remove the `startMcpListChecker()` call and its import from `src/web.ts`.

The Connectors dashboard page already has a manual "Refresh" button that
calls `refreshMcpListCache()` on demand. The cache starts empty; users see
their connectors after the first manual refresh. No new UI work needed.

## Verification

- **Structural test** added to `src/__tests__/mcp-list-no-boot-warmup.test.ts`:
  - Reads `src/web.ts` source text and asserts no call-site of
    `startMcpListChecker()` exists outside imports and comments.
  - Asserts the now-unused import is also removed.
  - Labelled STRUCTURAL (source-scan) because a behavioral test would require
    mocking the full server boot + all 20+ subsystem starts; cost is
    disproportionate. The source-scan is unambiguous.
- **Mental-revert evidence**: re-adding `startMcpListChecker()` to `web.ts`
  causes the `callLines.length === 0` assertion to fail with
  `Expected: 0, Received: 1`.
- **Full upstream suite**: 929/929 passes (pre-existing 4 `managed-settings`
  failures on Linux unrelated; +2 new tests from this branch).

## Origin

Self-contained to `src/web.ts` and `src/web/mcp-list.ts`; no fork-specific
dependencies.

---

### DIFF
--- a/src/__tests__/mcp-list-no-boot-warmup.test.ts
+++ b/src/__tests__/mcp-list-no-boot-warmup.test.ts
@@ -1,0 +1,52 @@
+// STRUCTURAL test — not BEHAVIORAL.
+//
+// The production invariant: startWebServer() MUST NOT call startMcpListChecker()
+// at boot time (not even with a delay). Calling it spawns the Telegram plugin
+// for a health-check, which 409-kills the live session-bridge process holding
+// the same bot-token (observed: channel offline within 33s on every deploy,
+// 2026-06-04).
+//
+// This test locks the invariant via source-text inspection of web.ts. A
+// behavioral test would require fully mocking the http server, all 20+ sub-
+// system starts, and module-load-time side effects -- the isolation cost is
+// disproportionate to the value here. The source-scan is unambiguous: the
+// function call either appears in the file or it doesn't.
+//
+// Mental-revert evidence: if you re-add `startMcpListChecker()` to web.ts,
+// the `callCount` assertion below fails with "Expected 0, Received 1".
+//
+// Related: PR #269 addresses a different 409 source (runtime poller-flapping).
+// These two fixes are complementary.
+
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged to `develop` via 9533a11 (cherry-picked clean onto develop + combined build/test green). Köszi! A PR base=main volt; a változás a develop-on van, innen promotálódik main-re a következő release-kor.

PR #302, opened 2026-06-06
scaffold: substitute identity placeholders when seeding scheduled tasks + guard shipped templates

koorbela:
## Why this matters

`ensureDefaultScheduledTasks()` seeds each task's `SKILL.md` into the user's `~/.claude/scheduled-tasks/` tree by byte-copying the file. The install scripts substitute `{{INSTALL_DIR}}`, `{{MAIN_AGENT_ID}}`, `{{BOT_NAME}}`, `{{OWNER_NAME}}` (and `{{CHAT_ID}}` in `CLAUDE.md.template`) via `sed`, but the runtime seed ran no substitution beyond `{{PROJECT_ROOT}}`. So any deployment-specific value left in a shipped template was copied verbatim into every other install:

- an absolute home path,
- the default `OWNER_NAME` (and its nickname),
- a hardcoded numeric chat id (one operator's personal channel),
- foreign sub-agent names used in examples.

A fresh install then gets tasks that point at a path that does not exist on that machine, address the wrong person, or post to someone else's chat.

## Change

1. `resolveTemplatePlaceholders()` now substitutes the full identity set, kept in sync with the install scripts' `sed`: `{{PROJECT_ROOT}}`, `{{INSTALL_DIR}}`, `{{MAIN_AGENT_ID}}`, `{{BOT_NAME}}`, `{{OWNER_NAME}}`. It is exported so it can be unit-tested.
2. `ensureDefaultScheduledTasks()` runs every non-`task-config.json` file through that substitution (with a `copyFileSync` fallback on a read/write error), and skips nested directories so a stray subdir cannot throw `EISDIR` and abort the whole seed. `task-config.json` keeps its existing agent-field rewrite.
3. The shipped templates are de-identified to use placeholders or generic terms: foreign sub-agent names in examples, a foreign product dir in a path, the default owner name (full + nickname), a hardcoded chat id, and a hardcoded main-agent sign-off are replaced.
4. A new test, `src/__tests__/template-identity-hygiene.test.ts`, scans the shipped template trees (`scheduled-tasks/`, `templates/`, `seed-scheduled-tasks/`, `seed-skills/`) for absolute home paths, personal emails, the default owner-name literal, and hardcoded numeric chat ids, and asserts the runtime seed substitutes every placeholder the task templates actually use.

## Scope / compatibility

No behavior change for a normally-configured install: the placeholders resolve to the host's own values, exactly as the install-time `sed` already produces. `task-config.json` handling is unchanged. The directory-skip is a no-op for the current flat task dirs. The new test reads only shipped, fully-tracked template files (it deliberately excludes the mixed `skills/` dir, which on a live checkout accumulates operator-local skills).

## Test plan

- `npm run typecheck` clean.
- `npx vitest run` green except a pre-existing, unrelated `managed-settings.test.ts` failure that also fails on a clean checkout.
- New test passes; mutation-checked (injecting a home path / personal email / owner-name / chat-id literal turns it red at the right line, restoring turns it green).
- Live render of the real templates through the real `resolveTemplatePlaceholders`: zero surviving placeholders, zero foreign-default identity, `{{INSTALL_DIR}}` resolves to the host path.

## Known limitations

- `{{CHAT_ID}}` is used only in `CLAUDE.md.template`, which is rendered by the install-time `sed` (not the runtime seed), so the runtime function does not handle it; the morning-digest task uses `chat_id: 0` (the bound channel) instead.
- The runtime seed reads top-level `scheduled-tasks/` while the installers seed from `templates/scheduled-tasks/` and `seed-scheduled-tasks/`; reconciling those source trees is out of scope here.
- `seed-skills/` is byte-copied by the installers without substitution; the cleaned skill files are kept host-agnostic by hand and the new guard scans that tree, but a generic substitution pass for skills is left as a follow-up.

### DIFF
--- a/scheduled-tasks/dream-engine/SKILL.md
+++ b/scheduled-tasks/dream-engine/SKILL.md
@@ -3,7 +3,7 @@ name: dream-engine
 description: Éjszakai analízis-loop az aznapi memóriákról, naplóról és kanban-állapotról. Generál 4 priorizált akció-javaslatot reggelre.
 ---
 
-Te most a "Dream Engine" éjszakai analízis-loopot futtatod. 02:07-kor vagy, Szabolcs alszik, NE küldj üzenetet a beállított csatornára.
+Te most a "Dream Engine" éjszakai analízis-loopot futtatod. 02:07-kor vagy, {{OWNER_NAME}} alszik, NE küldj üzenetet a beállított csatornára.
 
 A cél: az aznapi tudást átkonszolidálni és reggelre (07:30 Reggeli Napindító) felkészülni 4 priorizált javaslattal.
 
@@ -13,7 +13,7 @@ Generálj egy `{{INSTALL_DIR}}/DREAM.md` fájlt az alábbi 5 bucket alapján. A 
 
 ### Bucket 1 — 💡 Skill-javaslatok (flotta-szintű)
 
-Nézz végig MINDEN agent (marveen + sub-agentek: boni, deeper, iris, samu, zara) tegnapi (24h) memóriáit és napi naplóját. Kerítsd ki:
+Nézz végig MINDEN agent (a fő-ágens és az összes sub-agent) tegnapi (24h) memóriáit és napi naplóját. Kerítsd ki:
 - Volt-e 3+ szor visszatérő, manuálisan ismételt művelet ami skill-be illeszthető?
 - Új, NEM lefedett pattern amit érdemes lenne skillbe önteni?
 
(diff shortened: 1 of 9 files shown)


Szotasz:
Köszönjük a hozzájárulást, @koorbela! 🙏 A munkád (identity-placeholder substitúció + a shipped-template hygiene guard teszt) beépült a #315-be — ami most ment fel a developre, a te commitod szerzőséggel megőrizve. Kibővítettük a heartbeat OWNER_NAME-mel + a scriptek WEB_PORT-jával, így mind a 13 azonosított beégetés egy helyen megszűnik, és a hygiene-teszted a jövőben automatikusan elkapja az újakat. Köszi, hogy nemcsak jelezted, hanem javítottad is! 💛

Issue #279, opened 2026-06-04
RFC: loss-prevention continuity ledger as an alternative/complement to coordinator backfill


kovesdan:
## Summary

Telegram channel "deafness" — the agent's session is alive and logged in
("Listening for channel messages"), but inbound messages never reach the running
session, so the user's messages are silently dropped. We hit this in production
on Linux/systemd; the most painful variant is a mid-session gap where the bridge
consumes a Telegram update but the session never sees it.

The upstream line solves this with a standalone `channel-coordinator` (#259/#261/
#262/#269): a separate poller that detects native-bridge death and re-delivers
lost messages from `getUpdates` after the fact via its own `incoming_events`/
`poll_offset` tables and a reconcile loop.

This RFC describes our fork's alternative: a deterministic conversation-continuity
ledger that prevents the loss in the first place and surfaces any straggler into
the **already-running** session, rather than repairing after death. We are not
proposing to replace the coordinator; we are offering a different tradeoff that is
battle-proven on Linux, plus the remaining fork-only fixes if there's interest.

## Our approach — the continuity ledger

A small, dependency-free (Python stdlib + `better-sqlite3` schema) set of Claude
Code hooks around a durable `conversation_log` table (already merged upstream in
#260, commit `0ebf328`):

- **Capture (UserPromptSubmit).** `ledger-capture.py` parses the inbound
  `<channel source="plugin:telegram:telegram" …>` block and writes it to
  `conversation_log` (direction=`in`) *before* the agent processes the prompt.
  `agent_id` is derived from the session cwd, so the same hook is generic across
  all channel agents and never cross-contaminates.
- **Outbound (PostToolUse).** `ledger-outbound.py` records the agent's replies
  (direction=`out`), so "answered vs. unanswered" is decidable.
- **Replay (SessionStart).** `ledger-replay.py` injects the last ~20 turns
  (token-bounded to ~16k chars) plus a highlighted OPEN QUESTION into the fresh
  session's `additionalContext`. After a respawn the new session doesn't need to
  *remember* anything — its context window already carries the conversation and
  the unanswered question.
- **Live-drain (heartbeat, ~2 min).** `ledger-live-drain.py` handles the case
  SessionStart can't: a message lost in an *already-running* session. It
  re-surfaces the still-unanswered inbound (with a 60s grace so it never fights an
  in-flight reply, and a dedup statefile so it surfaces once) directly into the
  running session — no respawn required. It also surfaces undelivered inter-agent
[...]

PR #303, opened 2026-06-06
fix: continuity-ledger follow-ups — verified inter-agent submit + credit-wall defer; watchdog probes never open questions (implements #279)

kovesdan:
Implements the offer in #279 (RFC: loss-prevention continuity ledger). Targets `develop` per the maintainer's current cherry-pick flow.

## What this PR contains

The two remaining fork-only follow-up fixes offered in #279 (the third, `43de77d` router never-abandon, is **already on develop** via #284 — verified semantically present in `shouldAbandon`, so it is intentionally NOT re-included here):

1. **`fix(router): verified inter-agent submit + credit-wall defer`** — two silent-loss modes fixed at source: (a) long messages tripping the TUI bracketed-paste detector were marked *delivered* while parked unsubmitted (25-min real loss); the router now verifies the submit outcome (`SendPromptOutcome`), retries parked injections up to a cap with a clean buffer roll-back, then fails VISIBLY; (b) a session wedged on the 1M-context credit wall is detected (`detectsCreditWallWedge`, live-tail scoped with false-positive guards) and delivery is deferred instead of typing into a doomed pane.
2. **`fix(ledger): watchdog probes are never open questions`** — 3-artifact fix (source filter in `ledger_lib`, surfacing safety net in `ledger-live-drain`, shell contract tests): liveness probes (`__wd_ping <ts>`) never surface as OPEN_QUESTION and can no longer mask a real unanswered user question.

## Conflict-resolution notes (cherry-pick onto develop)

- `pane-state.ts`: develop's `parkedChannelInput` (#293) and our credit-wall block are different features at the same location — **both kept**.
- `agent-process.ts` `clearInputBuffer`: merged signature `export` (your #293 watchdog calls it) **+** `boolean` return (our verified-submit flow checks it); void-callers unaffected.
- `message-router.ts`: develop's richer channel-inbound-aware prefix-building kept; our verified-submit outcome flow added below it.
- `ledger-live-drain.py`: develop's early-exit flow kept; the `__wd_ping` safety net injected before the surfacing write.
- Our fork's swim-app files that were bundled in the original `ac66f49` commit are **excluded** (personal-fork feature, not upstream material).
- The ledger shell test now sets `OWNER_NAME=Gyula` explicitly, exercising your #291 env-derived owner-name path instead of asserting a hardcoded name.

## Verification

- `tsc --noEmit` clean on the branch.
- Focused suites: `interagent-submit-verify` + `router-abandon` + `pane-state` + `stuck-input-watcher` — **151 tests green**; the two new-feature suites alone: 21 green.
- `scripts/__tests__/conversation-ledger.test.sh`: **36/36 green** (includes the 3 new probe-filter contract tests).
- Full vitest on the branch: 1029 passed; the only failures (4) are in `managed-settings.test.ts`, which is **byte-identical to develop in this PR** (`git diff develop..HEAD --name-only` does not include it) — an environment-dependent command-string parse in our VM; expected green in your CI.
- Production provenance: both fixes have been running on our Linux/systemd fork since 2026-06-03/04; the verified-submit path recovered real msg-loss incidents (#230/#220 class), and the probe filter has prevented question-masking since 2026-06-04.

## Open question from the RFC, answered

You asked (well, we asked in #279) "separate PRs or folded into one" — this is the folded version; happy to split if you prefer.

---

*Original RFC text from #279 for context:*

## Summary

Telegram channel "deafness" — the agent's session is alive and logged in
("Listening for channel messages"), but inbound messages never reach the running
session, so the user's messages are silently dropped. We hit this in production
on Linux/systemd; the most painful variant is a mid-session gap where the bridge
consumes a Telegram update but the session never sees it.

The upstream line solves this with a standalone `channel-coordinator` (#259/#261/
#262/#269): a separate poller that detects native-bridge death and re-delivers
lost messages from `getUpdates` after the fact via its own `incoming_events`/
`poll_offset` tables and a reconcile loop.

This RFC describes our fork's alternative: a deterministic conversation-continuity
ledger that prevents the loss in the first place and surfaces any straggler into
the **already-running** session, rather than repairing after death. We are not
proposing to replace the coordinator; we are offering a different tradeoff that is
battle-proven on Linux, plus the remaining fork-only fixes if there's interest.

## Our approach — the continuity ledger

A small, dependency-free (Python stdlib + `better-sqlite3` schema) set of Claude
Code hooks around a durable `conversation_log` table (already merged upstream in
#260, commit `0ebf328`):

- **Capture (UserPromptSubmit).** `ledger-capture.py` parses the inbound
  `<channel source="plugin:telegram:telegram" …>` block and writes it to
  `conversation_log` (direction=`in`) *before* the agent processes the prompt.
  `agent_id` is derived from the session cwd, so the same hook is generic across
  all channel agents and never cross-contaminates.
- **Outbound (PostToolUse).** `ledger-outbound.py` records the agent's replies
  (direction=`out`), so "answered vs. unanswered" is decidable.
- **Replay (SessionStart).** `ledger-replay.py` injects the last ~20 turns
  (token-bounded to ~16k chars) plus a highlighted OPEN QUESTION into the fresh
  session's `additionalContext`. After a respawn the new session doesn't need to
  *remember* anything — its context window already carries the conversation and
  the unanswered question.
- **Live-drain (heartbeat, ~2 min).** `ledger-live-drain.py` handles the case
  SessionStart can't: a message lost in an *already-running* session. It
  re-surfaces the still-unanswered inbound (with a 60s grace so it never fights an
  in-flight reply, and a dedup statefile so it surfaces once) directly into the
  running session — no respawn required. It also surfaces undelivered inter-agent
  reports the router could not hand off while the session was unreachable.

Architecture in one line: capture-before-process → durable transcript →
replay-on-start + drain-into-running-session. The data path is the agent's own
context window, not a second inbound poller.

## How this differs from coordinator backfill

Different layer, different philosophy:

- **Coordinator = after-the-fact repair at the transport layer.** A second
  process polls Telegram directly when the native bridge looks dead, persists to
  its own tables, and a reconcile loop replays abandoned messages. The loss
  guarantee is formally closed and detection is fast (~10s).
- **Ledger = loss-prevention at the session/router layer.** We keep the native
  bridge as the single inbound path and make the session resilient: capture the
  message the moment it arrives, and re-surface it into the live or fresh session
  if it wasn't answered. No second poller, no second Telegram connection.

This is a genuine tradeoff, not a strict win. The coordinator detects and repairs
faster and entirely sidesteps the hard pane-state classification problem. Our
line recovers a real bridge death more slowly (see below) but adds no second
poller, no extra Telegram connection, and is already tuned for systemd
(EXIT_DEFER_TO_PEER / RestartPreventExitStatus). We think both are reasonable;
ours has simply been the one running in our production environment.

## What we observed in production (2026-06-04)

- A real outage was survived with **zero message loss**: a two-direction "acid
  test" passed — the false-positive-respawn direction stayed clean, and a genuine
  bridge death self-healed without losing a single message. The user's messages
  were delayed, not lost.
- Recovery time: an active inbound probe (`inbound-probe.ts`) — a userbot pings
  the bot every 3 minutes, and two missed round-trips declare the channel deaf
  and trigger a pane respawn — puts detection + recovery at **~3–6 minutes** for
  inbound deafness.
- Honest cost: a few minutes is still slower than coordinator backfill detection
  (~10s). If a multi-minute blind window is unacceptable, the coordinator's speed
  is the real argument for it.
- Origin of the design: a mid-session deafness incident on 2026-06-02 where the
  bridge consumed updates the session never saw — the case `ledger-live-drain.py`
  was built for.

## Known interaction/conflict points if both run together

From our conflict analysis (`plan/DEAFNESS_ARCH_COMPARISON_2026-06-04.md`); these
are why we'd be cautious about running both mechanisms simultaneously, not
objections to the coordinator itself:

- **Competing pollers stealing updates.** During a backfill window the
  coordinator polls `getUpdates` on the same bot token the native bridge may also
  poll. If liveness misreads "down" (e.g. a reparented bridge), both poll at once
  → 409 + inbound theft. We have hit this duplicate-poller class before.
- **Two reap/respawn deciders.** The coordinator's detached-claude reap
  (tmux-pane attribution) and our newest-wins/survivor-wins bridge reap target the
  same orphan problem in the same process tree; run naively together they can each
  classify the other's chosen survivor as an orphan → respawn thrash.
- **Loss-guarantee semantics conflict.** Upstream's router abandons a busy-but-
  alive session after ~1h so the coordinator's reconcile can replay the `failed`
  message. Our router (`43de77d`) *never* abandons a busy-but-alive session.
  Combined naively: keep our fix and the coordinator never finds a `failed`
  message to replay; revert to the abandon semantics and the coordinator
  re-delivers a message the router still held → double delivery.
- **Opposite reaction to a 409.** The coordinator treats a 409 as authoritative
  "native is alive" and suppresses backfill for ~90s; our recovery treats the same
  409 as a reap+respawn trigger. Same signal, opposite action.

The cleanest combination we found is selective: adopt upstream's
`wrapChannelInbound` / inbound preamble / forged-coordinator-id 403-guard
(`prompt-safety.ts`, conflict-free for us) regardless, and keep the two
loss-guarantee mechanisms from overlapping by gating one of them off.

## What we offer

All on the `kovesdan/marveen` fork:

- The full continuity ledger is already upstream as **#260** (`0ebf328`).
- Three fork-only follow-up fixes not yet PR'd, on branch `feat/swim-dashboard`
  (pushed as `origin/feat/swim-dashboard`):
  - `43de77d` fix(router): never silently drop inter-agent reports to a busy main
    session
  - `52c84e4` fix(router): verified inter-agent submit + credit-wall defer
  - `ac66f49` fix(ledger): watchdog probes are never open questions (3-artifact
    fix: downstream filter + upstream contract + contract test)
- We're happy to open these as focused PRs (they're generic, not personal-fork
  logic) and to contribute the conflict map above if you want to evaluate a hybrid.
- Besides this issue thread, you can also reach us by email at
  gyula.kovesdan.assistant@linistry.com if that's more convenient.

## Open questions

- Is a multi-minute blind window during a genuine bridge death acceptable for the
  upstream default, or is the coordinator's faster detection a hard requirement?
- If a hybrid is desired, which single loss-guarantee owner should win — router
  never-abandon (ours) or coordinator reconcile (upstream)? They cannot both be
  authoritative without double-delivery or a guarantee gap.
- Would you want the three follow-up fixes as separate PRs, or folded into one?

### DIFF
--- a/scripts/__tests__/conversation-ledger.test.sh
+++ b/scripts/__tests__/conversation-ledger.test.sh
@@ -33,7 +33,9 @@ run_hook() {
     local hook="$1"
     local db="$2"
     shift 2
-    LEDGER_DB_PATH="$db" LEDGER_OWNER_CHAT="8517922966" MAIN_AGENT_ID="marveen" \
+    # OWNER_NAME set explicitly: upstream #291 derives the transcript prefix
+    # from env (no hardcoded name); the prefix assertions below rely on it.
+    LEDGER_DB_PATH="$db" LEDGER_OWNER_CHAT="8517922966" MAIN_AGENT_ID="marveen" OWNER_NAME="Gyula" \
         python3 "$HOOKS_DIR/$hook" "$@"
 }
 
@@ -363,6 +365,32 @@ assert_eq "live drain: statefile records the surfaced message_id" "1122" \
 # (g2) same open question again -> dedup, no output
 OUT_G2="$(run_drain "$DB_LD1")"
 assert_eq "live drain: dedup suppresses re-surfacing the same message_id" "" "$OUT_G2"
+
+# (g-wd) watchdog probes are NEVER open questions (2026-06-04, marveen report:
+# the drain surfaced `__wd_ping <ISO>` as OPEN_QUESTION; answering probes is
+# forbidden by CLAUDE.md). Two invariants: a probe-only inbound surfaces
+# nothing; AND a probe arriving AFTER a real unanswered question must not
+# mask it (the source filter selects the newest NON-probe inbound).
+mkdir -p "$TMPDIR_BASE/ldw1"; DB_LDW1="$TMPDIR_BASE/ldw1/x.db"
+emit_inbound 8517922966 1571 "__wd_ping 2026-06-04T13:25:00Z" | run_hook ledger-capture.py "$DB_LDW1"
+age_rows "$DB_LDW1" 120
+OUT_GW1="$(run_drain "$DB_LDW1")"
+if printf '%s' "$OUT_GW1" | grep -q "OPEN_QUESTION"; then
+    fail "live drain: wd_ping probe surfaced as OPEN_QUESTION (got: $OUT_GW1)"
+else
+    pass "live drain: wd_ping probe-only inbound surfaces nothing"
+fi
+
+mkdir -p "$TMPDIR_BASE/ldw2"; DB_LDW2="$TMPDIR_BASE/ldw2/x.db"
+emit_inbound 8517922966 1580 "Valodi kerdes valasz nelkul" | run_hook ledger-capture.py "$DB_LDW2"
+emit_inbound 8517922966 1581 "__wd_ping 2026-06-04T13:30:00Z" | run_hook ledger-capture.py "$DB_LDW2"
+age_rows "$DB_LDW2" 120
+OUT_GW2="$(run_drain "$DB_LDW2")"
+if printf '%s' "$OUT_GW2" | grep -q "OPEN_QUESTION chat_id=8517922966 message_id=1580"; then
+    pass "live drain: a newer wd_ping does NOT mask the real unanswered question"
+else
+    fail "live drain: real question masked by a newer wd_ping (got: $OUT_GW2)"
+fi
 
 # (g3) a later 'out' answered it -> no output
 mkdir -p "$TMPDIR_BASE/ld3"; DB_LD3="$TMPDIR_BASE/ld3/x.db"
--- a/scripts/agent-context-guard.sh
+++ b/scripts/agent-context-guard.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# Context-exhaustion guard for the always-alive channel agents.
+#
+# Symptom it fixes (2026-06-02): sub-agents run with `claude --continue`, so
+# their conversation grows without bound. At ~100% context the next turn would
+# need the 1M-context window, which requires usage credits that are not enabled
+# -> every reply dies with "Usage credits required for 1M context". The agent
+# still RECEIVES inbound messages (channel is alive -- this is NOT deafness),
+# it just cannot answer.
+#
+# Recovery: issue `/clear` into the wedged pane. That resets the conversation
+# client-side (no API call, no credits), the channel stays connected, and the
+# agent answers the next message. We only act when the credit-wall error is
+# visible on the pane RIGHT NOW, so the agent is already non-functional and
+# /clear loses nothing that was not already broken.
+#
+# Invoked every minute by marveen-agent-context-guard.timer (systemd --user).
+
+set -u
+
+INSTALL_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+# tmux session names of the agents to guard. Override with CTXGUARD_SESSIONS
+# (space-separated) so the list is not hardcoded to one deployment's agents.
+if [ -n "${CTXGUARD_SESSIONS:-}" ]; then
+  read -r -a SESSIONS <<< "$CTXGUARD_SESSIONS"
+else
+  SESSIONS=(aladar-channels agent-dia agent-erno-ba agent-virgil)
+fi
+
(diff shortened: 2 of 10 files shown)


Szotasz:
Szia, köszi a PR-t! A main időközben elment alatta, és a branch most konfliktusos (CONFLICTING), így nem mergelhető. Meg tudnád rebase-elni a friss mainre (vagy merge main -> branch) és feloldani a konfliktusokat? Utána szívesen átnézzük és visszük tovább. Köszi!

Szotasz:
Köszönöm a continuity-ledger follow-upokat (#279). Ez a branch 3 hete áll és conflicting, közben a recovery/delivery-réteget alaposan átépítettük (v1.13-v1.15: dim-tudatos recovery, delivery-intent gate, IDLE_FOOTER_RX). A verified inter-agent submit nagyrészt lefedett. Ezért zárom; ha maradt benne be nem fedett rész, nyiss friss, rebase-elt PR-t a konkrét deltával.

PR #307, opened 2026-06-07
fix(agent-process,start): execFile tmux launch + optional google-sheets MCP

jocoo:
## Summary

Two small fixes bundled because both touch the agent startup path:

1. `agent-process.ts`: replace shell-formatted `execSync` for `tmux new-session` with `execFileSync`. Shell glob/escape pitfalls bite whenever the assembled command string contains characters the shell would interpret (spaces, quotes in agent names). `execFileSync` passes the argv array directly to tmux and sidesteps the shell entirely.

2. `start.sh`: optional Google Sheets MCP server launch, gated on `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` being present in `.env`. Default-off so existing installs are unaffected.

## Test plan

- [x] `npm run typecheck` clean.
- [x] Local: agent restart via `scripts/stop.sh && scripts/start.sh` succeeds; tmux session spawns without shell-interpretation errors.
- [x] With Google MCP creds absent, start.sh skips silently.

## Severity

Low–medium — the tmux escape issue is a latent footgun that surfaces only with certain agent names; the MCP add is opt-in.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -9,6 +9,8 @@ INSTALL_DIR="$(cd "$(dirname "$0")/.." && pwd)"
 if [ -f "$INSTALL_DIR/.env" ]; then
   SLUG="$(grep -E '^MAIN_AGENT_ID=' "$INSTALL_DIR/.env" | head -1 | cut -d= -f2-)"
   BOT_NAME="$(grep -E '^BOT_NAME=' "$INSTALL_DIR/.env" | head -1 | cut -d= -f2-)"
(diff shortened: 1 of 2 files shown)


jocoo:
Closing: the google-sheets MCP boot block in start.sh is no longer needed -- the project is migrating to Google Drive MCP for sheets access. The execFile-vs-execSync fix for the tmux launch could be split out into a smaller PR if there is interest.

PR #324, opened 2026-06-08
feat(dashboard/terminal): plain Ctrl+V paste in the xterm modal

Janokapapa:
## Summary

Make the agent terminal modal paste on **plain Ctrl/Cmd+V**, like every other terminal and the browser itself.

xterm.js treats `Ctrl+V` as a literal `^V` and only pastes on `Ctrl+Shift+V`. That surprises most users, and it silently breaks OS-level dictation / paste tools that emit a plain `Ctrl+V` — the text never reaches the PTY.

## What changed

In `openTerminalModal`, a `term.attachCustomKeyEventHandler` maps plain `Ctrl/Cmd+V` (no Shift/Alt) to a clipboard paste:

- reads `navigator.clipboard.readText()` and feeds the text through `term.paste()`, which routes it through the existing `onData` handler, so it reaches the PTY exactly like typed input;
- suppresses the native `^V` to avoid a double paste;
- leaves `Ctrl+Shift+V` working as xterm's built-in paste;
- ignores clipboard-permission failures silently (no behavior change when the clipboard API is unavailable).

`web/app.js` only — 18 lines, no dependencies.

## Testing

- `node --check web/app.js` passes.
- Full test suite green (the 4 pre-existing `managed-settings` failures on `develop` are unrelated).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -9496,6 +9496,24 @@ function openTerminalModal(agentName) {
   openModal(overlay)
   setTimeout(() => term.focus(), 50)
 
+  // Plain Ctrl+V paste. xterm treats Ctrl+V as a literal ^V and only pastes on
+  // Ctrl+Shift+V, which surprises most users -- and breaks OS-level dictation /
+  // paste tools that emit a plain Ctrl+V. Map plain Ctrl/Cmd+V to a clipboard
+  // paste; term.paste routes the text through the onData handler above, so it
+  // reaches the PTY exactly like typed input. Ctrl+Shift+V still works as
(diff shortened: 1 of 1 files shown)


Szotasz:
Köszi! Ez a dashboard-terminál klaszter (#318) része — mivel a dashboardban már van xterm-alapú agent-terminál (`agent-terminal.ts`), ezt a vonalat nem duplikáljuk, így ezt is zárom. Részletek a #318-nál. A meglévő terminálhoz szóló inkrementális javítást viszont szívesen fogadjuk! 🙏

PR #327, opened 2026-06-08
feat(dashboard/terminal): deep-linkable terminal in the URL (per-tab, reload-restorable)

Janokapapa:
The agent terminal modal was modal-only state: opening it left the URL on `#<page>`, so a reload lost the open terminal and you could not have a different terminal per browser tab.

This encodes the open terminal in the location hash as `#<page>/term/<agent>`, making the terminal **deep-linkable**:
- Open a terminal -> the URL becomes `#agents/term/<agent>`.
- **Reload** restores that terminal (the hash router reopens it).
- A **second browser tab** pointed at `#agents/term/<other-agent>` shows that agent's terminal -> one terminal per tab.
- Closing the terminal (X) or navigating to another page drops the `/term/<agent>` part.

### Changes (web/app.js)
- `parseHashRoute()` splits the hash into `{ page, termAgent }`.
- `openTerminalModal(agentName, fromRoute)` tracks `currentTerminalAgent`, guards against a reconnect loop on router re-entry, and writes the hash (unless invoked *from* the router).
- The inline `terminalClose` handler became a named `closeTerminalModal(fromRoute)` that clears state and resets the hash.
- `routeFromHash()` opens/closes the terminal to match the URL on load and on every `hashchange`.

No change to the terminal transport (SSE pane stream) -- this is purely URL/open-close state.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -9461,12 +9461,28 @@ async function handleAgentLogin(agentName, btn) {
 let terminalInstance = null
 let terminalSSE = null
 let terminalFit = null
+let currentTerminalAgent = null
 
-function openTerminalModal(agentName) {
+// Parse the location hash into a page id + optional terminal deep-link.
+// Format: "#<page>" or "#<page>/term/<agent>" -- the open terminal lives in the
+// URL, so a terminal can be opened per browser tab and survives a reload.
+function parseHashRoute() {
+  const raw = decodeURIComponent((location.hash || '').replace(/^#/, ''))
+  const parts = raw.split('/')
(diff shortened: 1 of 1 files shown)


Szotasz:
Köszi! Ez a dashboard-terminál klaszter (#318) része — mivel a dashboardban már van xterm-alapú agent-terminál (`agent-terminal.ts`), ezt a vonalat nem duplikáljuk, így ezt is zárom. Részletek a #318-nál. A meglévő terminálhoz szóló inkrementális javítást viszont szívesen fogadjuk! 🙏

PR #353, opened 2026-06-12
channels: scope the second orphan-poller reap to the main agent

koorbela:
## Why this matters

`scripts/channels.sh` runs on a main-agent channel restart. Its second orphan-reap pass kills stale provider pollers left by older plugin builds that carry `CLAUDE_PLUGIN_ROOT` but no `*_STATE_DIR`. It matched any process whose environment contained `CLAUDE_PLUGIN_ROOT=` plus the provider slug (e.g. `/telegram`).

The problem: `CLAUDE_PLUGIN_ROOT` resolves to the **shared user-level plugin cache** (`~/.claude/plugins/...`), which is identical for every agent. On a fleet that shares one tmux server, that match hit not only the main agent's own stale orphans but **every live sub-agent poller**. So a single main-channel restart SIGKILLed all of them, and each sub-agent's channel went dark until that sub-agent's own next restart.

This bites the multi-agent + per-agent-channel configuration (a first-class supported setup). A single-agent install is unaffected; an install whose sub-agents have their own provider channels loses all of them on a main restart.

## Change

Scope the second reap pass to the main agent by excluding any process whose environment references `$INSTALL_DIR/agents/<name>/` -- the directory every sub-agent's poller runs from (carried via its `*_STATE_DIR` and the project dir Claude Code sets from the launcher's `cd "${dir}"`). A main-agent orphan from an old build has no agent dir, so it is still reaped.

The exclusion uses a literal `index($0, subdir) == 0` test (not a regex), so an install path containing regex metacharacters cannot break it. The awk variable is named `subdir` rather than `sub` because `sub` is a reserved awk function name that BSD/macOS awk syntax-errors on.

## Scope / compatibility

Single-agent installs are unaffected (no `$INSTALL_DIR/agents/` processes exist, so nothing new is excluded). The fix strictly shrinks the kill set, so it cannot introduce a new false-positive. The first reap pass (STATE_DIR-scoped) and the reap timing (before the new session spawns) are unchanged.

## Test plan

- New test `src/__tests__/channels-reap-scope.test.ts` extracts the real awk program from `channels.sh` and runs it via `execSync` against fixture `ps eww -e` lines, asserting: main orphans are reaped, sub-agent pollers (matched via state dir and config dir) are spared, a different provider and non-poller processes are ignored, plus a regression lock on the scoping guard.
- `npm run typecheck` clean; full `vitest run` green.
- Verified against a live multi-agent fleet: the scoped matcher selects only the main-agent orphans and spares every live sub-agent poller, where the previous unscoped form matched all of them.

## Known limitations

The pre-existing `prov` provider match is a plain substring (`$0 ~ "/telegram"`), so it would also substring-match a hypothetical `/telegramx` elsewhere on an env line. This behavior is unchanged by this PR and is not exploitable in practice (the provider is only ever `telegram`/`slack`/`discord`, none a prefix of another). Anchoring the slug to the `CLAUDE_PLUGIN_ROOT` value is a separate follow-up.

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -131,9 +131,21 @@ fi
 # env (e.g. telegram@0.0.1). Those pollers carry CLAUDE_PLUGIN_ROOT=.../<provider>
 # instead, so the STATE_DIR grep above never matches and orphans accumulate
 # across restarts -> multiple getUpdates long-polls -> 409 Conflict -> the bot
-# goes silent/flaky. This runs BEFORE the fresh session is spawned, so every
-# matching poller is by definition a stale orphan and safe to kill.
-ORPHAN_PIDS2="$(/bin/ps eww -e 2>/dev/null | awk -v needle="CLAUDE_PLUGIN_ROOT=" -v prov="/${CHANNEL_PROVIDER}" '$0 ~ needle && $0 ~ prov { print $1 }')"
+# goes silent/flaky.
+#
+# Scope to THIS (main) agent only. The tmux server is SHARED across the fleet,
+# and every sub-agent runs its OWN provider poller out of
+# $INSTALL_DIR/agents/<name>/. Those processes carry that agent dir in their
+# environment; the main agent's pollers do not. CLAUDE_PLUGIN_ROOT points at the
+# shared user-level plugin cache for every agent, so it cannot tell main from
+# sub on its own -- without the agents/ exclusion this pass SIGKILLs every live
+# sub-agent poller on a main restart (they would only recover on each
+# sub-agent's own next restart). A main orphan from an old build has no agent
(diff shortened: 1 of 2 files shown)


koorbela:
Szia Szabolcs! Ki tudnál adni egy v1.7.4 release-t, ami tartalmazza ezt a #353-at (channels reap-scope fix)? A multi-agent setupunk szándékosan v1.6.0-n maradt, mert a v1.7.3 a #352 reap-logikát a #353 szűkítő fix nélkül szállítja, ami egy channels-restartnál kilőné a per-agent Telegram-pollereket. Amint a #353 felkerül a main-re, tisztán tudunk frissíteni. Kösz!

PR #364, opened 2026-06-12
feat(chat-app): Google Workspace login + per-user cookie sessions (/chat-api/auth)

juhasz007-maker:
First PR of the chat app track (multi-thread, per-user web chat next to the admin dashboard — plan agreed 2026-06-12). This adds the authentication foundation only; chat endpoints and the UI come in follow-up PRs.

## What

- **Google OAuth login** (`GET /chat-api/auth/login` → Google → `GET /chat-api/auth/callback`): authorization-code flow with state+nonce carried in a short-lived HttpOnly cookie. The id_token is obtained server-side directly from Google's token endpoint over TLS; `iss`/`aud`/`exp`/`nonce`/`email_verified` are validated server-side.
- **Workspace domain gate**: BOTH the `hd` claim and the email suffix must match `CHAT_ALLOWED_DOMAIN`. The `hd` *request* parameter is treated as a UX hint only — the decision is made exclusively on the returned claims, since the request param is client-side spoofable.
- **Email → agent mapping as allowlist**: `store/chat-users.json` (see `chat-users.example.json`), mtime-cached so it's editable without a restart. An unmapped email gets a login error even with a valid Workspace account.
- **Server-side sessions**: new `chat_web_sessions` table; the cookie carries a random 256-bit id, only its SHA-256 hash is persisted (a leaked DB/backup yields no usable cookie). HttpOnly + SameSite=Lax (+ Secure when `CHAT_PUBLIC_URL` is https), TTL configurable (`CHAT_SESSION_TTL_HOURS`, default 24h), lazy expiry on lookup + hourly prune sweep.
- **`/chat-api/auth/me`** (session probe) and **`POST /chat-api/auth/logout`** (server-side destroy + cookie clear).

## Isolation guarantees

- The `/chat-api/*` namespace is fully separate from the bearer-gated admin `/api/*`: a chat session cookie grants **nothing** under `/api/*`, and the dashboard token never reaches a chat user's browser.
- Every future chat endpoint scopes to the agent id stored in the session row — never to an agent name taken from the request (`getChatUser()` is the per-request boundary).
- Everything is behind `CHAT_APP_ENABLED` (default **off**). An existing install is byte-for-byte unaffected until it opts in and brings its own OAuth web client (`.env.example` documents the new vars).

## Tests

27 new tests (`src/__tests__/chat-auth.test.ts`): id_token claim validation (issuer/audience/expiry/nonce/unverified-email rejections), the hd+email domain gate (consumer-Gmail and foreign-workspace rejections), mapping allowlist behaviour, session round-trip/forged-sid/expiry/logout, cookie attributes. Full suite: 1224 passed.

Next PRs in the track: session-machinery generalisation (N sessions per agent), the `/chat-api/threads` API, then the chat frontend built on the #356 transcript renderer.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,3 +9,13 @@ HEARTBEAT_CALENDAR_ID=your_calendar_id_or_email
 # OLLAMA_URL=http://localhost:11434
 # MARVEEN_ENV=linux-server  # platform override: macos | linux-server | linux-gui
 # GOOGLE_API_KEY=your_google_api_key
+#
+# Chat app (per-user web chat, Google Workspace login) — opcionális:
+# CHAT_APP_ENABLED=1
+# CHAT_GOOGLE_CLIENT_ID=your_oauth_web_client_id.apps.googleusercontent.com
+# CHAT_GOOGLE_CLIENT_SECRET=your_oauth_web_client_secret
+# CHAT_ALLOWED_DOMAIN=example.com            # csak ennek a Workspace domainnek a fiókjai léphetnek be
+# CHAT_PUBLIC_URL=https://chat.example.com   # a böngésző felőli origin (OAuth redirect URI + Secure cookie)
+# CHAT_SESSION_TTL_HOURS=24
+# A belépő email -> agent leképezés a store/chat-users.json fájlban él
+# (lásd chat-users.example.json), restart nélkül újraolvasódik.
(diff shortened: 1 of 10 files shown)


juhasz007-maker:
Kulon appba kerult: juhasz007-maker/marveen-chat

PR #365, opened 2026-06-12
feat(chat-app): thread session machinery — N Claude Code sessions per agent

juhasz007-maker:
Second PR of the chat app track. **Stacked on #364** (its commit is included; rebase after #364 merges shrinks this to its own diff). No HTTP endpoints yet — this is the session-machinery generalisation the `/chat-api/threads` API (next PR) will sit on.

## What

- **`chat_threads` ledger**: thread id, agent, title, status (`open`/`suspended`/`closed`), `claude_session_id` minted at creation.
- **Deterministic session identity**: threads spawn with `claude --session-id <uuid>`, so the transcript is `<uuid>.jsonl` in the agent's projects dir and reopening is `--resume <uuid>`. No newest-file guessing (the #356 conversation view's pattern), no discovery races between concurrently created threads.
- **Namespaced tmux sessions**: `agent-<name>-thread-<id>`, running in the agent's own workdir → shared CLAUDE.md + memory tiers, separate conversation context per thread.
- **Always channel-less**: no `--channels` flag, channel state-dir env vars pointed at an inert per-thread dir, token env vars unset — the same neutralisation `startAgentProcess` applies to channel-less agents, so a thread can never 409-war the main session's bot token.
- **Lifecycle**: `suspendThread` (kill tmux, reopenable), `closeThread` (drops out of the default list, still reopenable), reopen via `--resume` with a fresh-start fallback under the same id when the thread never received a first message. `sendPromptToThread` fails synchronously on a dead/busy thread — chat UX wants an immediate error, not the router's queue-and-abandon-after-1h.
- **Idle backstop**: a 5-minute sweep (gated on `CHAT_APP_ENABLED`) auto-suspends open threads idle for `CHAT_THREAD_IDLE_MINUTES` (default 45). User-driven close remains the primary mechanism per the agreed plan; `CHAT_MAX_OPEN_THREADS_PER_AGENT` (default 10) is the cap input the API PR will enforce.
- **Remote (ssh) agents**: explicitly unsupported for threads (clean error).

## Main-session safety

The existing 1-agent-1-session path is untouched: no changes to `agent-process.ts`, `message-router.ts`, `schedule-runner.ts` or the channel monitors. Threads only *reuse* exported primitives (`sessionExistsOnHost`, `sendPromptToSession`, `isSessionReadyForPrompt`, `projectsDirFor`). Heartbeat/scheduled tasks keep targeting the main session only.

## Tests

17 new (`src/__tests__/chat-threads.test.ts`): ledger CRUD + status transitions + per-agent open-count, session-name namespacing, the launch-command builder (fresh `--session-id` vs reopen `--resume`, channel neutralisation, model quoting for `[1m]` ids, strict-profile flag handling, api-key/config-dir injection, ollama base-url), and the idle-sweep decision incl. boundary. Full suite: 1241 passed.

One thing worth a reviewer's eye: the channel-neutralisation assumes the auto-loaded channel plugin reads its token exclusively from the state-dir `.env` / token env vars — that matches the channel-less branch in `startAgentProcess`, but if there's a fallback path I missed, threads of a channel-owning agent would need the `enabledPlugins` override instead (which is shared workdir state, hence avoided here).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,3 +19,5 @@ HEARTBEAT_CALENDAR_ID=your_calendar_id_or_email
 # CHAT_SESSION_TTL_HOURS=24
 # A belépő email -> agent leképezés a store/chat-users.json fájlban él
 # (lásd chat-users.example.json), restart nélkül újraolvasódik.
+# CHAT_THREAD_IDLE_MINUTES=45                # idle szál auto-felfüggesztés (backstop)
+# CHAT_MAX_OPEN_THREADS_PER_AGENT=10         # laza felső határ a nyitott szálakra
--- a/.env.example
+++ b/.env.example
@@ -9,3 +9,13 @@ HEARTBEAT_CALENDAR_ID=your_calendar_id_or_email
 # OLLAMA_URL=http://localhost:11434
 # MARVEEN_ENV=linux-server  # platform override: macos | linux-server | linux-gui
 # GOOGLE_API_KEY=your_google_api_key
+#
+# Chat app (per-user web chat, Google Workspace login) — opcionális:
+# CHAT_APP_ENABLED=1
+# CHAT_GOOGLE_CLIENT_ID=your_oauth_web_client_id.apps.googleusercontent.com
+# CHAT_GOOGLE_CLIENT_SECRET=your_oauth_web_client_secret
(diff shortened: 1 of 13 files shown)


juhasz007-maker:
Kulon appba kerult: juhasz007-maker/marveen-chat (lasd #364)

PR #366, opened 2026-06-12
feat(chat-app): per-user thread API (/chat-api/threads) — CRUD, messages, transcript timeline

juhasz007-maker:
Third PR of the chat app track. **Stacked on #365** (and transitively #364) — their commits are included; a rebase after they merge shrinks this to its own diff.

## Endpoints (plan §7, cookie-session auth)

| Method | Path | Behaviour |
|---|---|---|
| GET | `/chat-api/threads` | the user's own threads (+ `open_count`/`max_open` for the UI) |
| POST | `/chat-api/threads` | new thread; 409 over the per-agent cap; spawn failure keeps the row suspended (reopenable) instead of vanishing it |
| PATCH | `/chat-api/threads/:id` | rename |
| DELETE | `/chat-api/threads/:id` | close (drops from default list, stays reopenable) |
| POST | `/chat-api/threads/:id/reopen` | `--resume` the session |
| GET | `/chat-api/threads/:id/messages` | chat timeline from the deterministic transcript `<claude_session_id>.jsonl` |
| POST | `/chat-api/threads/:id/messages` | send prompt; **202 + retry_after** while a suspended thread boots; **409 + retry_after** while the previous turn runs; first message auto-titles the thread |

## Authorization invariant

Every endpoint resolves the agent **exclusively** from the logged-in user's session row (`getChatUser`) — never from the path or body. A thread id owned by another agent 404s indistinguishably from a nonexistent id. The admin `/api/*` surface remains unreachable with a chat cookie by construction.

## Timeline parser

Thread-specific (`thread-transcript.ts`), not the #356 agent-conversation parser: thread sessions have no channel plugin, so the colleague's messages are plain typed prompts rather than `<channel>` wrappers. `tool_result` user-lines, `<system-reminder>`s and slash-command echoes are filtered out so they don't render as fake user bubbles; `tool_use` blocks become collapsed action labels.

## Tests

7 new (`chat-thread-transcript.test.ts`): user/assistant rendering, tool_result and harness-noise filtering, action labels, malformed-line tolerance + limit semantics, auto-title derivation. Full suite: 1248 passed.

Next (last) PR: the chat frontend (thread list + chat UI, mobile-friendly).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #367, opened 2026-06-12
feat(chat-app): chat frontend (/chat/) — thread list + mobile-friendly chat UI

juhasz007-maker:
Final PR of the chat app track. **Stacked on #366** (and transitively #365, #364) — rebases shrink it once those merge.

## What

A minimal vanilla-JS SPA (matching the dashboard's no-build approach) served from `web-chat/` under `/chat/`:

- **Login screen** → `/chat-api/auth/login`; `login_error` codes from the auth callback render as Hungarian messages (forbidden domain, no agent mapping, cancelled, …).
- **Thread sidebar**: new / select / rename / close, with open-running/booting/suspended markers; becomes a drawer under 720px (the plan requires mobile usability).
- **Chat view**: user/assistant bubbles, tool actions as collapsed muted rows, 3s polling (backs off to 15s on hidden tabs), auto-scroll only when the reader is already near the bottom.
- **Composer**: Enter=send / Shift+Enter=newline, auto-grow. A 202 (suspended thread booting) auto-retries while keeping the draft in the input; 409 (busy turn / thread cap) lands in a status bar.
- **Static route** (`chat-static.ts`): filename whitelist — user input is never path-joined, so there is no traversal surface. Served separately from the dashboard's `web/` so the colleague-facing subdomain origin never carries admin HTML (proxy forwards `/chat/` + `/chat-api/` only, root → `/chat/`; snippet in the docs).
- **`docs/chat-app.md`**: enable steps (OAuth client, `store/chat-users.json`, nginx), the security model, lifecycle/sizing notes.

Login redirects in `chat-auth.ts` now land on `/chat/` accordingly.

## Honest test status

Full suite green (1251, incl. 3 new whitelist tests for the static route). The UI itself has **not** been exercised against a live Google OAuth flow + spawned thread sessions yet — that needs the deployed environment (OAuth client + subdomain). Happy to do a deploy-time smoke pass.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,3 +19,5 @@ HEARTBEAT_CALENDAR_ID=your_calendar_id_or_email
 # CHAT_SESSION_TTL_HOURS=24
 # A belépő email -> agent leképezés a store/chat-users.json fájlban él
 # (lásd chat-users.example.json), restart nélkül újraolvasódik.
+# CHAT_THREAD_IDLE_MINUTES=45                # idle szál auto-felfüggesztés (backstop)
+# CHAT_MAX_OPEN_THREADS_PER_AGENT=10         # laza felső határ a nyitott szálakra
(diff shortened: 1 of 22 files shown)


juhasz007-maker:
Kulon appba kerult: juhasz007-maker/marveen-chat (lasd #364)

PR #371, opened 2026-06-13
fix(channels): recover paste-placeholder stuck input + wait-for-idle send gate

koorbela:
## Why this matters

Messages delivered to a Claude Code agent session (via the inter-agent router, the scheduler, and the worker) are typed into the pane with `tmux send-keys`. Two failure modes leave a fully written prompt parked in the input box, never submitted, until a human presses a key:

1. **Paste-placeholder (primary).** When a long message arrives as a burst (which happens under tmux-server contention during multi-session broadcasts), the TUI bracketed-paste detector turns it into a `[Pasted text #N]` placeholder. No Enter submits a placeholder, so the existing retry loop sent plain Enters that did nothing. Worse, the placeholder render is mis-read by the prior stuck-detectors (which keyed on the idle footer), so the pane was classified as not-stuck and none of the existing watchdogs fired. A documented stuck-rate of roughly 60% on 5-way broadcasts traced to this.
2. **Busy-send.** The sender wrote into a pane that was mid-turn; the text plus Enter did not submit, and the prompt parked once the turn ended.

## Change

- **`pane-state.ts`**: new `detectsPastePlaceholder()` is checked BEFORE the idle-footer gate in `detectPaneState` / `shouldRetrySubmit`, and is SCOPED to the live input-box region (the bottom box between the separators) so a `[Pasted text` string quoted in scrollback or reply text does not trigger a false positive. The regex tolerates the stub wrapping across a line break (`[Pasted text` newline `#N]`), which is how it actually renders for long inputs. `decideSubmitFollowup` returns a new `clear-and-resend` action for placeholders.
- **`agent-process.ts`**: `discardPlaceholderBuffer()` clears a confirmed placeholder with Ctrl-C (empirically the only key that clears it; Enter, Ctrl-U, and Escape do not), guarded to fire only while a placeholder is confirmed in the scoped region, so it never hits an empty box. `waitForPaneIdle()` is a bounded (12s) pre-flight gate that waits for the pane to be idle before sending, so messages are not typed into a busy turn. The post-send retry loop now routes placeholders to clear-and-resend instead of wasting Enters.
- **`schedule-runner.ts`**: `forceSend` tasks opt out of the idle gate (`waitForIdle: false`) to preserve their existing skip-the-busy-check behavior.
- **`auto-restart-runner.ts`**: refactored to the shared `paneLooksIdle` predicate (no behavior change).

## Scope / compatibility

- No change to the chunked send mechanism itself (it measured robust into idle panes; alternatives were worse).
- Idle-pane happy path: one extra capture, immediate return, no added latency.
- All non-forceSend callers keep the idle gate on by default.
- Ctrl-C only ever fires when a placeholder is confirmed present in the input-box region.

## Test plan

- `npm run typecheck` -> 0 errors.
- `npx vitest run` -> 1251 / 1251 passing (81 files; the change adds new cases).
- New tests assert deterministic predicate output (not model behavior), including synthetic panes that reproduce the real wrapped-stub render shape and a scrollback-echo false-positive guard.
- Verified directly against captured panes from real incidents: genuinely stuck panes detect as placeholder / clear-and-resend; healthy panes detect as idle / done.
- End-to-end: 10 / 10 pre-formed placeholders recovered through the real compiled send path.

## Known limitations

- The placeholder-to-stuck causal chain under real production multi-session contention could not be reproduced in an isolated harness; the fix is mechanism-agnostic (it recovers any placeholder however it formed), so this does not weaken it.
- A separate, pre-existing issue (not addressed here): `PARKED_INPUT_RX` does not account for a non-breaking-space prompt glyph, which can make the background stuck-input watcher mis-classify a verbatim-parked pane. Tracked separately; it does not affect this PR placeholder path.

### DIFF
--- a/src/__tests__/pane-looks-idle.test.ts
+++ b/src/__tests__/pane-looks-idle.test.ts
@@ -1,0 +1,231 @@
+import { describe, it, expect } from 'vitest'
+import { paneLooksIdle, isReadyForPrompt } from '../pane-state.js'
+
+// Unit tests for paneLooksIdle -- the pure idle predicate extracted as the
+// SINGLE source of truth for "is this pane idle / safe to send a prompt into".
+// It backs three call sites: the readiness check (isReadyForPrompt), the
+// auto-restart idle-guard (auto-restart-runner.paneIsIdle), and the
+// sendPromptToSession pre-flight wait-until-idle gate (waitForPaneIdle).
+//
+// These assertions are on DETERMINISTIC string->bool output only; no LLM
+// behaviour, no tmux, no timers. Fixtures reproduce real `tmux capture-pane -p`
+// bytes (U+2500 ─ box separators, U+276F ❯ prompt, U+23F5 ⏵ footer chevrons,
+// U+00B7 · footer dot). Sanitised: no internal names/paths.
+
+const SEP = '─'.repeat(80)
+
+// --- Idle surfaces (predicate must be TRUE) ---
+
+// Default bypass-permissions footer with the shift+tab-cycle hint, empty box.
+const IDLE_BYPASS = [
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle)',
+].join('\n')
+
+// Strict-mode footer ("? for shortcuts"), empty box -- also idle.
+const IDLE_STRICT = [
+  '',
+  SEP,
+  '❯ ',
+  SEP,
+  '  ? for shortcuts',
+].join('\n')
+
+// Background-shells footer variant. A session with a BashTool background shell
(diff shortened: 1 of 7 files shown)


koorbela:
Szia Szabolcs! A develop-on időközben összegyűlt pár fix: #353 (Telegram reap-scope), #371 (paste-placeholder beragadás), plusz a te #372/#373 channel-javításaid. A multi-agent setupunk szándékosan v1.6.0-n maradt, mert a v1.7.3 még a #353 nélküli, így addig nem tudjuk frissíteni a rendszert, amíg nincs ezekből egy v1.7.4 release. Ki tudnál adni egy v1.7.4-et a develop-ról, hogy tisztán fel tudjuk húzni az egészet? Köszönöm!

PR #488, opened 2026-06-29
feat(ops): prompt-watchdog -- auto-clear safe prompts + auto-resume after rate limit

Rotary89:
## What

A systemd timer-driven watchdog that keeps a headless agent's tmux session moving without a human present. Two jobs:

1. **Clear safe interactive prompts** -- `/compact` is auto-confirmed; any other unrecognised `Enter to confirm` dialog is left untouched and an alert is sent (never auto-accepts unknown prompts).
2. **Auto-resume after a usage rate limit** -- on detecting a limit (the interactive "Stop and wait" menu, or the inline "you've hit your session limit" banner) it enters a wait mode and, every ~10 min while the session is idle, injects a short English continue prompt so the agent resumes the interrupted task from its `--continue` context. It then verifies: if a limit marker still shows it reschedules; if the agent is already working again it just clears the wait state; on success it notifies and exits wait mode.

## Why poll-and-verify instead of parsing the reset time

The usage-limit wording and reset-time format are undocumented and version-dependent, so parsing them is brittle. Instead the retry cadence is the robustness layer: a fixed ~10 min poll-and-verify loop that can never wedge permanently, even if a future CC update changes the wording. Detection is deliberately **strict** (the interactive menu requires all three of its lines; the inline form requires the rendered `/upgrade ...` command hint) because the watchdog runs against the very session that may be *discussing* these phrases -- loose matching would self-trigger on prose. Detection is best-effort; the loop is what guarantees recovery.

## Files

- `scripts/prompt-watchdog.sh` -- detection + safe-clear + rate-limit wait/retry/resume
- `scripts/systemd/prompt-watchdog.service` -- oneshot unit (placeholders for install dir / user)
- `scripts/systemd/prompt-watchdog.timer` -- runs every 30s, `OnBootSec=1min`

## Notes

- Distribution-safe: `INSTALL_DIR` is derived from the script location; `MAIN_AGENT_ID` / `TELEGRAM_BOT_TOKEN` / `ALLOWED_CHAT_ID` are read from `.env` at runtime; units ship with `/path/to/marveen`, `/home/USER`, `USER=youruser` placeholders.
- Wait state is a single marker file (`store/.ratelimit-waiting`, content = next-retry epoch) so other watchdogs can tell "waiting on a rate limit" from "stuck".
- Continue prompt is intentionally one line; the session context survives via `--continue`, so no parked-prompt restore is needed.

### DIFF
--- a/scripts/prompt-watchdog.sh
+++ b/scripts/prompt-watchdog.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+# Monitors the main agent's <id>-channels tmux pane for interactive prompts.
+# Safe prompts (rate limit, compact) are auto-handled.
+# Unknown interactive prompts trigger a Telegram alert.
+#
+# Safe auto-responses:
+#   - Rate limit "Stop and wait"  -> "1" Enter
+#   - Compact "Auto-compact"      -> "1" Enter
+#
+# Alert (never auto-accept):
+#   - Any other "Enter to confirm" dialog
+
+set -u
+
+INSTALL_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+STORE="$INSTALL_DIR/store"
+STATE_FILE="$STORE/.prompt-watchdog-state"
+LOG_TAG="prompt-watchdog"
+
+log() { echo "$(date '+%Y-%m-%d %H:%M:%S') [$LOG_TAG] $*"; }
+
+# Resolve session name from .env
+MAIN_AGENT_ID="$(grep -E '^MAIN_AGENT_ID=' "$INSTALL_DIR/.env" 2>/dev/null | head -1 | cut -d= -f2-)"
+MAIN_AGENT_ID="${MAIN_AGENT_ID:-marveen}"
+MAIN_AGENT_ID="${MAIN_AGENT_ID//[^a-zA-Z0-9_-]/}"
+SESSION="${MAIN_AGENT_ID}-channels"
+
+send_telegram() {
+  local msg="$1"
+  local TOKEN CHAT_ID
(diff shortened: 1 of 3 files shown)


Szotasz:
Thanks for this, the unattended rate-limit auto-resume is genuinely useful for headless installs.

One change is needed before we can merge, on the main-session injection path. The watchdog injects keystrokes plus a continue-prompt into the main channels session (send-keys "1" plus the continue message). That sits on an attack surface we recently hardened: we disabled the raw keystroke-injection endpoint after a phantom-injection incident where parked text got re-typed into a session. Re-introducing main-session auto-injection on-by-default works against that posture, even though this path is local (systemd) rather than a network endpoint.

There is also an overlap with #474 (model-fallback-on-limit, already merged): #474 reads the same main-session pane and matches the same usage-limit prompts, then switches model. #488 would dismiss the prompt and inject a continue. Both firing on the same limit event would race or double-act.

Could you gate the keystroke and continue-prompt injection behind an opt-in config flag, default off? With it off the watchdog would only detect and log (no injection), which keeps the hardened-by-default posture and avoids the #474 race. Users who want unattended resume can opt in explicitly. When enabled, it should coordinate with #474 (either defer to its model switch, or have a single owner for main-session limit handling).

Note this is Linux-only (systemd units), so it is inert on non-systemd hosts regardless. The detection logic and script structure look good; this is purely about gating the injection. Thanks.

Rotary89:
Visszavonva -- a prompt-watchdog funkciót a saját telepítésünkben töröltük, már nem használjuk, így nem tartjuk fenn a beolvasztási igényt.

PR #490, opened 2026-06-29
fix(channels): plugin pre-warm + longer startup guard with blind-accept fallback

Rotary89:
## What

Two startup-reliability fixes for `scripts/channels.sh`.

### 1. Plugin pre-warm (stdio handshake corruption)

The plugin's MCP `start` runs `bun install --no-summary && bun server.ts`. On the **first** load after a plugin update, `bun install` resolves deps and leaks progress to **stdout**, corrupting the MCP stdio JSON-RPC handshake. Claude marks the server **Failed** and kills it, leaving no bun-poller child: Telegram inbound dies and the `reply` tool disappears. The post-init unlock misses it because the `/mcp` row is not literally `✗ Failed`.

Fix: warm `node_modules` synchronously before launch (`bun install --silent`, 120s timeout, scoped to the active provider's plugin dir), so the server's own `bun install` is a fast no-op with clean stdout.

### 2. Startup guard timeout + blind-accept fallback

- Raise the accept-dialog wait from **12s → 30s**; after a slow boot the dialog can appear late and 12s was not enough.
- If `Listening for channel messages` is never reached within the window, make a blind `2`+Enter (then a bare Enter) attempt and log it to `store/channels-failures.log`, so a changed dialog string can't silently wedge the session.

## Notes

- `bash -n` clean. No identity/path leaks; provider dir resolved from `$HOME` + `$CHANNEL_PROVIDER`.

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -204,6 +204,20 @@ fi
 # down, so there is never a second concurrent poller in steady state. Nothing to
 # set here: the coordinator gates itself on native liveness.
 
+# PLUGIN PRE-WARM (2026-06-05 incident): the plugin's MCP `start` script runs
+# `bun install --no-summary && bun server.ts`. On the FIRST load after a plugin
+# update, `bun install` resolves dependencies and prints progress -- some of it
+# leaks to STDOUT, which corrupts the MCP stdio JSON-RPC handshake. Claude marks
+# the server Failed and kills it, leaving NO bun-poller child (Telegram dead,
+# reply tool absent). The post-init unlock below misses it because the /mcp row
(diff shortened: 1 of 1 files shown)


Rotary89:
Closing: this is edge-case hardening (cold node_modules on first load after a dependency-changing plugin bump). The normal path is already silent (`bun run --silent` + `bun install --no-summary` is a no-op when node_modules are warm), which is why the plugin runs fine for everyone. Our actual past bridge-down incidents had a different root cause (stale .in_use lock + missing channel-watchdog), already fixed and shipped. Not worth carrying upstream; can be reintroduced locally if we ever hit the cold-deps window.

PR #500, opened 2026-06-30
fix(parked-input): escalate main-agent stuck box to operator, never auto-clear it (#2 delivery fix)

Szotasz:
## Problem
2026-06-30 ~1h **silent** inter-agent delivery stall: a real, intended reply (`"Igen, írd meg Baloghnak a választ"`) was parked in the MAIN agent's channel input box. The stale-parked janitor tried to Ctrl-U clear it **109×**, failed every time, and only logged a lone WARN. Every inter-agent message to the main agent (an email-send hand-off, a #498 verify-GO, status reports) stranded as `pending` for an hour with no operator-visible signal. And had the Ctrl-U *succeeded*, it would have silently destroyed the operator's real reply.

## Fix (`clearStaleParkedInput`)
1. **MAIN agent box is NEVER auto-cleared.** A parked line there is very likely a real reply, so we send **zero keystrokes** (respects the hardened no-raw-keystroke area, #481) and instead **escalate to the operator** via `notifyChannel()` — the same Bot-API path `channel-monitor` uses, independent of the wedged session — **once** per stuck episode. The operator sends or clears the line.
2. **Sub-agent box** keeps the existing Ctrl-U clear (its parked text is the usual junk heartbeat line), but now also escalates **once** if the clear genuinely resists several times, so a stuck sub-agent can't silently stall either.

Escalation is **one-shot** per stuck text (tracked by sig on the `unwedgeAttempts` record; resets when the text changes or the box clears) — no spam, mirroring the kanban 2-round pattern.

## Co-review emphases (per the design agreement)
- ✅ Main-agent box never silently auto-cleared (the most critical constraint).
- ✅ Escalation is NOTIFY-only (a Telegram message to the operator) — **no keystroke / phantom-injection**.
- ✅ One-shot per episode (no spam).
- Orthogonal to #498 (STT throttle, already shipped) and to the forthcoming #1 pull-model (busy-delivery).

## Verification
- `tsc --noEmit`: clean.
- 27/27 recovery-surface tests pass: `janitor-parked-input` (unchanged contract, no regression), the new `parked-input-escalation` (main never-clear + escalate-once + one-shot + sub-agent still clears), `router-abandon`, `channel-inbound-framing`, `remote-routing`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #502, opened 2026-06-30
fix(parked-input): stale-frame guard on #500 main escalation (suppress false-positive)

Szotasz:
## Incident
#500's main-agent escalation sent a **false-positive** to the operator (2026-06-30 16:25): `parked="Koszi a halakat."` — a leftover fragment of the main agent's own persona text (planted by a send-keys delivery into the main box), not a real unsent reply.

**Forensic:** the fragment read as "parked" for **28 min while the main agent was actively turning**. A genuinely stuck box blocks all delivery → the agent would be idle the whole window. So the `'typing'`+parked capture was a **stale frame**, and the escalation fired on it.

## Fix
Gate the main escalation on a liveness probe: `agentHasActivitySince()` (new, indexed `conversation_log` query) checks whether the main agent logged any turn (in/out) within 120s. If it did, the box is **not** actually blocking delivery — the capture is stale — so **suppress** the NOTIFY. Only a genuinely idle/wedged box (no recent turns) escalates, preserving the real case the escalation exists for (operator typed a reply and forgot to send). `main-never-clear` unchanged (still zero keystrokes to the main box).

This is the **stale-frame guard** answer to the design question: a pure 2-capture stability check can't detect a stale frame (both reads are identically stale); the cross-check against *actual agent activity* is the robust discriminator (a real wedge ⇒ no activity).

## Verification
- `tsc`: clean.
- `vitest`: **10/10** — new "suppress on recent activity" + existing escalate-once / one-shot / sub-agent-still-clears / router-abandon / janitor.

## Relation to the pull-model (#1)
The pull-model removes the **root cause** (send-keys delivery fragments in the main box) separately. This guard hardens the escalation against stale frames regardless, and is independently worth keeping.

### DIFF
--- a/src/__tests__/parked-input-escalation.test.ts
+++ b/src/__tests__/parked-input-escalation.test.ts
@@ -34,9 +34,13 @@ vi.mock('node:child_process', async (orig) => ({
   }),
 }))
 vi.mock('../notify.js', () => ({ notifyChannel: vi.fn(async () => {}), notifyTelegram: vi.fn(async () => {}) }))
+// Stale-frame guard liveness probe: default = no recent activity (idle/wedged),
+// so the existing escalate path fires; a test flips it true to assert suppression.
+vi.mock('../db.js', async (orig) => ({ ...(await orig() as object), agentHasActivitySince: vi.fn(() => false) }))
 
 import { clearStaleParkedInput } from '../web/agent-process.js'
 import { notifyChannel } from '../notify.js'
+import { agentHasActivitySince } from '../db.js'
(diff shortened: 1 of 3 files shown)


Szotasz:
Superseded by #503 (dim-text guard) + the v1.18.3 escalation-mute. #502's activity-guard was a workaround (suppress when the agent is active); #503 is the precise root fix per Szabi's insight -- it distinguishes a dim ghost from a real parked line at the detection source (captureParkedInputView strips SGR-2 dim), so ghost/phantom lines are never treated as parked for ANY agent, and the muted main escalation can later be re-enabled safely. Closing in favor of #503.

PR #530, opened 2026-07-07
fix: improve WSL host stability handling

iszzu80-dev:
## Problem

On WSL2, the whole utility VM can shut down and re-boot on its own (idle
`vmIdleTimeout`, Windows sleep/resume, `wsl --shutdown`, a Windows/WSL update).
That tears down the kernel, system + user systemd, tmux, and every service at
once. It is **not** an application crash, and `Linger=yes` does not protect
against it — yet operators can't tell a whole-VM restart apart from a service
crash, so a fleet-wide silence looks like an app failure. Separately, the
generated `marveen-channels.service` placed `StartLimitIntervalSec` /
`StartLimitBurst` in `[Service]`, where systemd logs "Unknown key … in section
[Service]" and ignores them — so the crash-loop throttle never actually applied.

## What changes

- **StartLimit fix:** move `StartLimitIntervalSec` / `StartLimitBurst` to
  `[Unit]` in the channels-service generated by `install-linux.sh`, so the
  throttle takes effect.
- **Host-restart watchdog:** `scripts/host-restart-watchdog.sh` +
  `marveen-host-watchdog.service` — a oneshot at every user-manager start that
  compares `/proc/stat btime` to a saved value and, on a change, sends one
  notice that the host/WSL VM restarted (with an estimated downtime).
- **App-crash notifier:** `scripts/unit-fail-notify.sh` + `marveen-notify@.service`
  — an `OnFailure=` template that reports a service crash. btime-change ⇒ host
  restart; OnFailure ⇒ app crash — the two are now distinguishable.
- **Docs:** `scripts/systemd/README-host-stability.md`, including the Windows-side
  `%UserProfile%\.wslconfig` → `vmIdleTimeout=-1` mitigation (documented only,
  not automated from Linux).

## Why safe

- Additive: new scripts/units + one systemd key relocation. No behavioural change
  to any existing service beyond the throttle now working as intended.
- No secrets: the notifier reads the Telegram token from the existing local env
  file at runtime; nothing is committed.
- No dashboard / provider / DNS / deploy change. No CostOps/brand/landing/local
  helper code — the diff is 6 files, WSL/stability scope only.

## Test / verify

- `systemd-analyze --user verify` clean on channels / host-watchdog / notify@
  (no more "Unknown key").
- Throttle active after reload: `StartLimitIntervalUSec=5min`, `StartLimitBurst=5`.
- Watchdog exercised in all branches: baseline-init (silent), unchanged-btime
  (silent), simulated restart (detects + computes downtime); notification
  send-path returns HTTP 200.

### DIFF
--- a/install-linux.sh
+++ b/install-linux.sh
@@ -1174,6 +1174,11 @@ cat >"$SYSTEMD_DIR/${CHAN_UNIT}.service" <<EOF
 [Unit]
 Description=${BOT_NAME} Channels (Telegram bridge)
 After=network.target
+# StartLimit* belong in [Unit], not [Service] (systemd logs "Unknown key
+# StartLimitIntervalSec in section [Service]" and ignores them there), so the
+# crash-loop throttle only applies when they live here.
+StartLimitIntervalSec=300
+StartLimitBurst=5
 
 [Service]
 Type=simple
@@ -1191,8 +1196,6 @@ ExecStartPre=$INSTALL_DIR/scripts/ensure-native-modules.sh
 ExecStart=$INSTALL_DIR/scripts/channels.sh
 Restart=on-failure
 RestartSec=10
-StartLimitIntervalSec=300
-StartLimitBurst=5
 StandardOutput=append:$INSTALL_DIR/store/channels.log
 StandardError=append:$INSTALL_DIR/store/channels.error.log
 Environment=PATH=$HOME/.local/bin:$HOME/.bun/bin:/usr/local/bin:/usr/bin:/bin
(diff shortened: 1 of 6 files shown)


Szotasz:
Thanks for this — the systemd StartLimit fix and the btime-based reboot classification are solid design. Two blockers before merge:

1. **Hardcoded contributor chat-id fallback in the notify scripts.** With a hardcoded fallback, every downstream install would send its host-stability alerts to that private Telegram chat via its own bot token. The chat-id must come from config (the owner/channel config already available to the install) with no hardcoded fallback — otherwise it's a distribution-safety leak. This is the main blocker.

2. **No actual WSL detection + missing token-env path.** The 'WSL VM restarted' alert currently fires on every Linux reboot, not just WSL — please gate it behind a real WSL check (e.g. /proc/version or the WSL env marker). Also the token-env path the notify step references doesn't exist on a standard install, so the feature silently no-ops there.

The systemd mechanics can come in cleanly once these are addressed.

iszzu80-dev:
Thanks for the review -- both blockers addressed in 7bf59be:

**1. Hardcoded contributor chat-id fallback (main blocker).** Removed the `:-8942301795` fallback in both `unit-fail-notify.sh` and `host-restart-watchdog.sh`. The alert target now comes solely from `MARVEEN_ALERT_CHAT_ID` with no hardcoded default, so a downstream install can't leak its host-stability alerts to a private chat. When it's unset the send is skipped (see below), never redirected.

**2a. Real WSL detection.** The 'WSL VM restarted' wording is now gated behind an actual WSL check (`/proc/version` microsoft/wsl marker, `WSL_DISTRO_NAME`, or `/run/WSL`). On a non-WSL host the same btime-change detection still works but is worded as an ordinary 'host restarted' rather than claiming a WSL VM reboot.

**2b. Token-env path / silent no-op.** The token path stays configurable via `TELEGRAM_ENV` (default `~/.claude/channels/telegram/.env`), but the missing-config path no longer silently no-ops -- it now logs exactly which piece is missing (`TELEGRAM_BOT_TOKEN` and/or `MARVEEN_ALERT_CHAT_ID`) to stderr, so a misconfigured install is diagnosable instead of quietly doing nothing.

Both scripts pass `bash -n`; base is already `develop`. Ready for another look.

PR #558, opened 2026-07-11
feat(intel): argparse CLI for the v4 intelligence registry

maslentrek-dev:
## Summary
- Adds `scripts/intel_db.py` with an argparse CLI: `add-fact` (upsert into `known_facts_registry`, deterministic `<domain>-<YYYYMMDD>-<hash8>` id when `--id` is omitted, DUPLICATE-content handled as a clean no-op), `add-watch`, and `dump` / `--dump` (JSON of registry + watchlist + active_focus).
- Root cause being fixed: the hourly intelligence collector never wrote the registry (it only posted a daily-log line), while the daily brief reads exclusively from the registry tables — so the brief was structurally always empty. The collector prompt now has a one-line write path.
- The `get_*` reader functions (which the brief imports) are unchanged; no-arg invocation keeps the original health-counter behavior.

## Test plan
- Verified all CLI paths on an isolated DB copy: auto-id insert, identical-content dedup (UNIQUE `fact_hash` → DUPLICATE, exit 0), same-id upsert-update with status transition, `add-watch`, `--dump` flag alias output, and the backwards-compatible no-arg counter.
- `python3 -m py_compile` clean. Live end-to-end (collector round → `--dump` > 0 rows → daily brief with content) verified on the production host.

### DIFF
--- a/scripts/intel_db.py
+++ b/scripts/intel_db.py
@@ -1,0 +1,247 @@
+"""
+intel_db.py -- Helper functions for the Proactive Intelligence Cycle (v4).
+Database: /Users/gruzmanarnold/marveen/store/intel.db
+"""
+
+import sqlite3
+import hashlib
+import time
+import uuid
+from pathlib import Path
+
+DB_PATH = Path(__file__).parent.parent / "store" / "intel.db"
+
+
+def _conn():
+    con = sqlite3.connect(DB_PATH)
+    con.row_factory = sqlite3.Row
+    return con
+
+
+def get_active_registry(days: int = 14) -> list[dict]:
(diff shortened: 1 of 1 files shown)


Szotasz:
Thanks for this. We are going to close it for now, because the v4 intelligence registry subsystem it targets does not exist in this repo: there is no schema, migration, or caller for it here, and the script exits on a plain run. It also has values wired to a specific private install (local paths and personal column names).

The code itself is clean (stdlib only, no network, no credential access), so this is not a concern about the code, just about fit. If you would like to contribute the whole subsystem in a generalized form (schema + collector + the brief wiring, with config placeholders instead of personal values), we would be glad to review it. Closing for now, and thanks for the contribution.

PR #567, opened 2026-07-11
Channel self-heal, sub-agent inbox wake-nudge, and Graphify dashboard panel

norbull:
## Summary

Three self-contained improvements from a downstream deployment, offered back upstream:

1. **`fix(channels)`** — self-heal `enabledPlugins` before launching the channel session.
2. **`feat(message-router)`** — wake idle sub-agents on a stuck Telegram inbox.
3. **`feat(dashboard)`** — Graphify knowledge-graph panel + async regen wired into the deploy flow.

Each is independent; you can cherry-pick a subset. Setup notes below.

## Validation
- `npm run typecheck` green
- wake-nudge gate unit tests 6/6 green

---

## 1. `fix(channels)` — channel plugin self-healing guard

**What it solves.** `scripts/channels.sh` launches the channel session against a specific `PLUGIN_ID` (resolved from `CHANNEL_PROVIDER`). If a PR review-reset or a branch-switch reverts `.claude/settings.json` to a baseline that does not have that plugin under `enabledPlugins`, the session comes up but with no active channel plugin — the bot is "online" yet silently receives nothing.

**How it works.** Before launch, an idempotent guard (`_ensure_plugin_enabled`) ensures `PLUGIN_ID` is `true` under `enabledPlugins` in **both** the project (`$INSTALL_DIR/.claude/settings.json`) and the global (`$HOME/.claude/settings.json`) settings, re-adding it if missing. Pure `python3`, no new dependency.

**Setup.** None. It runs automatically at channel launch. `PLUGIN_ID` is derived from `CHANNEL_PROVIDER` in `.env` (defaults to `telegram@claude-plugins-official`). No settings file is committed by this change.

---

## 2. `feat(message-router)` — sub-agent Telegram inbox wake-nudge

**What it solves.** A sub-agent that loads the channel plugin as a plain MCP server (rather than via `--channels`) does not get Claude Code's native channel notifications delivered into its turn. If inbound messages are persisted to a per-agent pending file, a drain hook can pull them on the next turn — but an **idle** sub-agent never takes a turn, so the inbox sits stuck. This watcher closes that latency gap.

**How it works.** `telegram-inbox-wake.ts` is called once per message-router tick. For each sub-agent it `statSync`s the pending inbox (cheap gate); only when it has sat untouched past a min-age (25s) and the per-agent debounce (60s) has elapsed does it probe the tmux session and, if idle, inject a **content-free** nudge so the drain hook fires and atomically claims the backlog. No content is injected and the inbox file is never mutated by the router — the drain hook owns the claim and the security framing. Cheap gates run before any tmux I/O, so an empty fleet costs one `statSync` per agent and zero `capture-pane`.

**Prerequisites (IMPORTANT — no-op without these).** The wake reads
`<agent-channel-state-dir>/inbox-pending.jsonl`. Without that file it hits `ENOENT` and is a passive no-op — safe, but inert. To make it functional a deployment needs a "pull model" with two pieces:

1. **An inbound tee.** A small process that receives each inbound channel message for a sub-agent and appends it (one JSON object per line) to that agent's `inbox-pending.jsonl` under the channel state dir (`resolveAgentChannelStateDir(agent, 'telegram')`). In our deployment this is a `channel-inbound-tee.mjs` sitting on the sub-agent's MCP stdio; any equivalent that writes the same file works.
2. **A drain hook.** A `UserPromptSubmit` hook that, on each turn, atomically claims (`rename`) `inbox-pending.jsonl` and prepends the already-security-framed `<channel>` messages above the current prompt. In our deployment this is `channel-inbox-drain.py`.

These two infra pieces are **deployment-specific and NOT included in this PR** (they depend on how you run sub-agent channel plugins). The wake-nudge is the missing scheduler piece: it only decides *when* to nudge an idle session so the drain hook you already have can run. If you deliver sub-agent channel inbound some other way, adapt the pending-file path/format accordingly.

**Setup.** No config once the two prerequisites exist — it runs automatically inside the message-router tick. Tunables are module constants (`SUB_TELEGRAM_WAKE_MIN_AGE_MS`, `SUB_TELEGRAM_WAKE_DEBOUNCE_MS`).

---

## 3. `feat(dashboard)` — Graphify knowledge-graph panel

**What it does.** Adds a "Graphify" dashboard panel: an `<iframe>` that renders a standalone D3 knowledge-graph of the codebase (`web/icons/graphify-graph.html`), served by the existing static route. Sits alongside the existing Costs panel. A lazy loader in `app.js` only fetches the graph on first entry to the panel.

**Prerequisites.**
- **Required** — [Graphify](https://github.com/safishamsi/graphify): `pip install graphifyy` (or `uv tool install graphifyy`; note the double-y package slug). Provides the `graphify` CLI that builds the graph.
- **Optional** — a local [Ollama](https://ollama.com) with `qwen2.5-coder:7b` (`ollama pull qwen2.5-coder:7b`), used **only** to name the graph's communities semantically (zero API cost). Without it the graph still builds and works fully, just with numbered `Community N` labels instead of named ones.

**Setup — generate the first graph.** The graph HTML is a generated ~2.5MB artifact and is **gitignored** (`web/icons/graphify-graph.html`), so it is absent on a fresh checkout and the panel iframe would 404 until built. Generate it with:

```bash
bash scripts/regen-graphify.sh
```

That script indexes `src/` with tree-sitter AST (`graphify extract --code-only`, zero API cost), names communities via local Ollama when available (falling back to placeholder labels), and copies the result to `web/icons/graphify-graph.html`. It is best-effort and idempotent: if `graphify` or its inputs are absent it exits 0 without touching anything, so it never breaks a deploy.

**Deploy integration.** `update.sh` calls `scripts/regen-graphify.sh` after the build (best-effort, `-x`-guarded), so the panel stays fresh across deploys where Graphify is installed and is silently skipped where it is not.

---

Co-authored with Claude Opus 4.8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,4 @@ scripts/support-mail/no-support-reply.html
 # Python bytecode cache
 __pycache__/
 *.pyc
+web/icons/graphify-graph.html
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -46,6 +46,32 @@ case "$CHANNEL_PROVIDER" in
   *)        PLUGIN_ID="telegram@claude-plugins-official" ;;
 esac
 
+# Self-healing guard: ensure PLUGIN_ID is enabled in both project and global
+# settings.json before launch. Any PR review-reset or branch-switch that
+# reverts .claude/settings.json loses this entry; the guard re-adds it
+# automatically, preventing silent Telegram outages.
+_ensure_plugin_enabled() {
+  local settings_file="$1"
+  [ -f "$settings_file" ] || return
+  python3 - "$settings_file" "$PLUGIN_ID" <<'PYEOF'
+import json, sys
+path, plugin_id = sys.argv[1], sys.argv[2]
+try:
+    d = json.load(open(path))
+except Exception:
+    d = {}
+plugins = d.setdefault("enabledPlugins", {})
+if plugins.get(plugin_id) is not True:
+    plugins[plugin_id] = True
+    with open(path, "w") as f:
+        json.dump(d, f, indent=2)
+    print(f"channels.sh: enabled {plugin_id} in {path}", flush=True)
+PYEOF
+}
+_ensure_plugin_enabled "$INSTALL_DIR/.claude/settings.json"
+_ensure_plugin_enabled "$HOME/.claude/settings.json"
+unset -f _ensure_plugin_enabled
+
 # ROOT-CAUSE NOTE (kali-linux WSL, claude-code 2.1.152, 2026-05-27):
 # Inbound MCP notifications from the `--channels` plugin go through a SECOND
 # gate beyond --dangerously-skip-permissions / --dangerously-load-development-
--- a/scripts/regen-graphify.sh
+++ b/scripts/regen-graphify.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+# regen-graphify.sh -- rebuild the Graphify knowledge-graph HTML for the dashboard.
+#
+# The dashboard "Graphify" panel serves web/icons/graphify-graph.html, which is
+# gitignored (a ~2.5MB generated artifact, see .gitignore). After a fresh
+# checkout/deploy that file is missing, so the iframe would 404. This script
+# regenerates it from the current src/ using tree-sitter AST (zero API cost),
+# plus local ollama community naming when available.
+#
+# Best-effort and idempotent: if graphify or its inputs are absent it exits 0
+# without touching anything, so it NEVER breaks a deploy. Re-uses the graphify
+# extract cache under store/ so repeat runs only re-parse changed files.
+
+set -uo pipefail
+
+PROJECT_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+SRC_DIR="$PROJECT_ROOT/src"
+DEST="$PROJECT_ROOT/web/icons/graphify-graph.html"
+WORK="$PROJECT_ROOT/store/graphify-work"
+OLLAMA_MODEL="qwen2.5-coder:7b"
+
+# graphify installs as a uv tool under ~/.local/bin; make sure it is on PATH.
+export PATH="$HOME/.local/bin:$PATH"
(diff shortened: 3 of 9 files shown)


Szotasz:
Thanks for this — there's genuinely useful work here, but it bundles three separate concerns and one of them can't ship as-is. Could you split it into focused PRs?

**1) Channel plugin-guard** — good idea, but as written it risks data loss: on a JSON parse error it overwrites the entire user-global `settings.json` with an empty object (losing hooks/model/permissions on every launch), the write is non-atomic (can race our existing hook-registration guard from #565), and it force-enables the plugin user-globally (re-introduces the duplicate-socket / 409 hazard). Please make it skip-on-parse-failure (never overwrite on error), atomic write, and project-scoped or gated rather than a user-global force-enable. Happy to take this as its own PR first.

**2) Sub-agent inbox wake-nudge** — the nudge logic itself is careful (fixed text, idle-gate, debounce), but it's default-ON with no gate, and its prerequisite (the tee-writer + drain-hook) doesn't exist upstream yet, so it can't actually fire. Please make it opt-in / config-gated default-OFF, and land it after the drain path exists.

**3) Graphify dashboard panel + the `graphifyy` pip install in update.sh** — we can't merge this: running an unfamiliar pip package from the deploy path needs a separate supply-chain review (the repo is `graphify` but the pip slug is `graphifyy`), and it overlaps with our existing GitNexus integration, so it's also a product decision. Let's discuss it separately.

Leaving this open, but we can't merge as-is — the split above is the path, and (1) with the fixes is the one we'd take first. Thanks again for the effort.

norbull:
Closing in favour of the focused split requested in review: this PR bundled three unrelated changes, now separated into

- #574 -- channel plugin-guard (with the parse-skip / atomic-write / project-scoped fixes)
- #575 -- sub-agent inbox wake-nudge (opt-in, default-OFF)
- #576 -- Graphify dashboard panel (**draft**: pending supply-chain review of the `graphifyy` package + a product decision on GitNexus overlap)

Please review those instead. Thanks for the review notes.

PR #571, opened 2026-07-12
fix(scheduler): pin cron parsing to Europe/Budapest timezone

huszarrichard:
## Summary
- cron-parser defaulted to the process timezone (UTC on this host), so every scheduled task written in Budapest wall-clock time (e.g. "30 7" for 7:30) silently fired 1-2h late. Pins `CRON_TZ = 'Europe/Budapest'` across all `CronExpressionParser.parse` call sites in `src/web/cron.ts`.
- Adds a restart catch-up window that records late-fired tasks distinctly (`fired_late`) so the run history surfaces exactly what was missed instead of folding it into a normal `fired` row.
- Fixes per-agent chat_id resolution for scheduled "task"-type prompts: the previous literal `chat_id: 0` placeholder was never actually translated by the channel plugin (confirmed empirically -- the Telegram reply tool rejects `chat_id="0"` as not-allowlisted). `resolveAgentOwnerChatId()` now resolves the running agent's own bound channel `.env` first, falling back to the global owner.

## Test plan
- [x] `npx tsc --noEmit -p .` passes
- [x] Manually verified on a live install: scheduled tasks authored for Budapest wall-clock times now fire at the correct local time instead of 1-2h late.

Found and fixed while operating a live Marveen install.

### DIFF
--- a/src/__tests__/schedule-runner-precheck.test.ts
+++ b/src/__tests__/schedule-runner-precheck.test.ts
@@ -117,7 +117,7 @@ describe('schedule-runner pre-check integration (source-level)', () => {
   })
 
   it('passes preCheckPrefix to attemptFireTask in the cron loop', () => {
-    expect(SRC).toMatch(/attemptFireTask\(task, agentName, now, cronPc\.prefix\)/)
+    expect(SRC).toMatch(/attemptFireTask\(task, agentName, now, cronPc\.prefix, lateCatchUpMs\)/)
   })
 
   it('skips and records the run when pre-check returns skip in cron loop', () => {
--- a/src/web/cron.ts
+++ b/src/web/cron.ts
@@ -1,7 +1,17 @@
 import { CronExpressionParser } from 'cron-parser'
 
+// All scheduled-task cron expressions (SKILL.md/task-config.json, the
+// dashboard schedule editor) are authored in Budapest wall-clock time -- "30
+// 7 * * *" means 7:30 for Ricsi, not 7:30 on whatever timezone the host
+// happens to boot in. cron-parser defaults to the PROCESS timezone when no
+// `tz` is given; on this host that's UTC, so every schedule silently fired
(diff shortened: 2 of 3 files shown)


Szotasz:
Thanks — the timezone math here is correct, and the underlying bug is real: cron-parser defaults to the process-local TZ, so on a UTC host every schedule fires 2h late in summer. The IANA-zoned, DST-correct approach is right and doesn't double-shift on a Budapest host.

Three things before we can merge:

1. **Don't hardcode `Europe/Budapest` in the shipped code** (the Ricsi comment too). This repo ships to many installs in different timezones — please read it from a `SCHEDULER_TZ` config/env key, defaulting to the host timezone. Hardcoding one operator's zone is a distribution issue.
2. **One vitest failure** (the precheck test) — needs the one-line regex fix so the suite is green.
3. **Scope creep:** this PR also swaps the `chat_id:0` bound-channel convention for a runtime-resolved telegram id, which hardcodes the telegram provider (a Slack/Teams-bound agent would get the wrong fallback). Please drop that from this PR — either a separate PR, or route it through `getChannelChatId()`.

Once those land we'll merge. Thanks for catching the UTC-host bug.

huszarrichard:
Addressed all three:
1. `Europe/Budapest` is no longer hardcoded -- `CRON_TZ` now reads `SCHEDULER_TZ` from env, falling back to the host's own timezone (`Intl.DateTimeFormat().resolvedOptions().timeZone`) when unset.
2. Fixed the stale `schedule-runner-precheck` test regex that predated the `lateCatchUpMs` param this PR adds to `attemptFireTask`.
3. Dropped the `chat_id:0` -> `resolveAgentOwnerChatId()` runtime-resolution change entirely -- reverted to the original `chat_id: 0` literal. Will follow up with a separate PR for that, routed through `getChannelChatId()` as suggested.

`npx tsc --noEmit -p .` and the full vitest suite (129 files, 1793 tests) both pass.

PR #573, opened 2026-07-12
feat(token-monitor): per-model cost accuracy, MCP server/cost columns, model backfill upsert

cett:

- [x] Új funkció (feature / new feature)

## Rövid leírás / Summary

This PR extends the AI Usage Report (Token Monitor) UI with three improvements and fixes a correctness bug in cost estimation.

**1. Model backfill upsert** (`src/web/token-usage.ts`)

`insertCall` changed from `INSERT OR IGNORE` to an upsert (`ON CONFLICT DO UPDATE SET model/thinking_tokens WHERE ... IS NULL`). This allows a cursor-reset + re-collect to back-fill the `model` column on the ~20k existing rows that pre-date the model column (added in 330d9d6). After backfill: 7 477 rows remain NULL (genuine old transcripts with no `message.model` in the JSONL — expected).

**2. Per-model cost accuracy** (`src/web/token-usage.ts`, `web/app.js`)

`renderTuSummary` was calling `tuCalcCostUSD` with `null` model, forcing all tokens to Sonnet pricing and causing the agent-card total (~$124) to differ from the model-distribution chart total (~$169, which correctly priced Opus at 5×). Fixed:
- `getTokenSummary` now runs a second query grouped by `(agent, model)` and attaches a `perModel[]` array to each summary row.
- `renderTuSummary` sums `tuCalcCostUSD` per-model entry; `null`/`(unknown)` model falls back to default (Sonnet) pricing, matching the model-dist chart logic exactly.
- **Acceptance test result:** agent-card total = model-dist total = **$175.5077** (diff: $0.000000) on a 7-day window against the real DB.

**3. MCP server column + cost column in tool stats grid** (`web/app.js`, `web/lang/`)

- `getToolStats` now groups by `(tool_name, model)` and returns token sums per row; client aggregates into one row per tool.
- New `tuMcpServerFromTool(name)` helper: extracts server name from `mcp__<server>__<tool>` pattern (`__` as separator, handles single-underscore server names like `plugin_telegram_telegram`); returns `null` for built-in tools.
- Two new columns: **MCP server** (server name or "beépített"/"built-in") and **Est. cost** (USD, per-model accurate).
- Agent name cells replaced with colored chip pills (`tuGetColor` border + text color).
- i18n: `tool_stats_col_server`, `tool_stats_col_cost`, `tool_stats_builtin` keys added to `hu.js` and `en.js`.

**4. PostToolUse hook activation** (`.claude/settings.json`)

A new `PostToolUse` hook entry is added (no matcher = fires on every tool call) that runs `scripts/hooks/tool-log-capture.py`. This is the capture side of the existing `tool_call_log` infrastructure (table + `/api/tool-log` route, introduced in 330d9d6). Without this hook entry the table would remain empty; with it, every tool invocation is logged for the activity dashboard. Timeout is 5 s and errors are swallowed so the hook never blocks the agent.

## Review fixes (follow-up commit `12acd36`)

Addressed feedback from code review:

- **WEB_PORT**: replaced hardcoded `localhost:3420` with a config-driven `_web_port()` helper (WEB_PORT env var → `.env` file → default 3420), matching the pattern used by `taskstate-replay.py` and other hooks.
- **Secret redaction**: added `_redact()` with regex patterns covering Bearer tokens, `key=value` pairs, provider-specific prefixes (`ghp_`, `sk-`, `sk-ant-`, `xoxb-`) and raw hex blobs ≥ 32 chars. Applied in `_input_summary()` before the value is sent to the DB.
- **Retention**: the hook calls `POST /api/tool-log/prune` with ~1% probability, activating the existing `pruneToolCallLog()` function so `tool_call_log` does not grow indefinitely without a dedicated scheduled task.

## Ellenőrzőlista / Checklist

- [x] Kipróbáltam a módosítást a lokális környezetben; a dashboard hibamentesen működik.
- [x] A kód nem tartalmaz beleégetett szenzitív adatot.
- [x] Saját, beszédes nevű branch-ről nyitom (extended-token-monitor).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -10,7 +10,8 @@
       { "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/inbox-drain.py\"", "timeout": 15 } ] }
     ],
     "PostToolUse": [
-      { "matcher": "mcp__plugin.telegram.telegram__reply", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/ledger-outbound.py\"", "timeout": 15 } ] }
+      { "matcher": "mcp__plugin.telegram.telegram__reply", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/ledger-outbound.py\"", "timeout": 15 } ] },
+      { "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/tool-log-capture.py\"", "timeout": 5 } ] }
     ],
     "SessionStart": [
       { "matcher": "startup|resume|clear", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/ledger-replay.py\"", "timeout": 15 } ] }
--- a/scripts/hooks/tool-log-capture.py
+++ b/scripts/hooks/tool-log-capture.py
@@ -44,8 +44,8 @@ _SECRET_PATTERNS = [
     re.compile(r'(?i)((?:token|secret|password|api[_\-]?key|apikey|auth|credential)\s*[=:]\s*)[^\s,\'";&|]{6,}'),
     # GitHub/Anthropic/OpenAI style tokens
     re.compile(r'\b(ghp_|sk-|sk-ant-|xoxb-|xoxp-)[A-Za-z0-9_\-]{10,}'),
-    # Raw hex/base64 blobs ≥ 32 chars (likely hashed secrets)
-    re.compile(r'\b([0-9a-fA-F]{32,})\b'),
+    # Raw hex blobs ≥ 32 chars (likely hashed secrets) -- no capture group, full match replaced
+    re.compile(r'\b[0-9a-fA-F]{32,}\b'),
 ]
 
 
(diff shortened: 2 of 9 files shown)


Szotasz:
Nice work on the per-model cost accuracy. We reviewed the new `tool-log-capture.py` hook for privacy and it's reasonable (only tool name + a 200-char arg summary, written to a Bearer-gated localhost DB, not exposed in the UI). Four fixes before merge:

1. **`localhost:3420` is hardcoded** — use the `WEB_PORT` pattern the other hooks already use.
2. **Redact secrets in the arg summary** — inline literal tokens/keys in a tool call currently go verbatim into the DB. Mask obvious secret shapes before storing.
3. **Auto-prune wiring** — `tool_call_log` grows unbounded right now; add a prune/retention step.
4. **Declare the hook activation in the PR body** — the `.claude/settings.json` hook is the most sensitive change here and isn't mentioned.

(2) is the blocker; the rest can be fast-follows if you prefer, but let's get the redaction in before this merges. Thanks!

Szotasz:
Thanks, the WEB_PORT fix, the prune wiring, and the PR-body declaration are all good. One blocker left before merge:

The **hex-blob redaction is still leaking**. The 4th `_SECRET_PATTERNS` entry (the 32+ char hex pattern) has a capture group, so `re.sub` only appends `[REDACTED]` as a suffix and keeps the original hex secret in the stored string — a 32+ char token still lands in the DB verbatim, even though the PR body says it's covered. Please drop the capture group from that pattern (1 line) so the whole match is replaced. Once that's in, it's an immediate GO (Samu re-checks the redaction, then we merge).

PR #575, opened 2026-07-12
feat(message-router): opt-in wake-nudge for idle sub-agents with stuck Telegram inbox

norbull:
Splits the sub-agent inbox wake-nudge out of #567 as its own PR, **opt-in and default-OFF**.

## What
`src/web/telegram-inbox-wake.ts`, called once per message-router tick, wakes an idle sub-agent whose derived Telegram inbox (`<state>/inbox-pending.jsonl`) has stuck inbound messages its drain hook cannot pull without a turn. It statSync-gates the inbox (cheap) and, only when it is stuck past a min-age + per-agent debounce, probes the session and injects a **content-free** nudge so the drain hook fires and atomically claims the backlog. No content-inject, no inbox mutation.

## Default-OFF / infra-only
Gated behind `SUBAGENT_TELEGRAM_WAKE_ENABLED`, **default OFF**. The prerequisite inbound-tee writer and the UserPromptSubmit drain hook that produce/consume `inbox-pending.jsonl` are **not yet upstream**, so with them absent there is no inbox file to wake on and the watcher is a no-op even when enabled. Shipping it disabled makes this PR infra-only: an upstream install sees zero behaviour change and pays no per-tick cost until the drain path lands and the operator opts in.

## Verification
- `npm run typecheck` clean.
- The pure gate decision (`shouldWakeForTelegramInbox`) is unit-tested across all five conditions -- 6 tests green.

Part of the #567 split. Kept default-OFF per review so it can be merged now and activated after the drain path lands.

### DIFF
--- a/scripts/__tests__/channel-inbox-drain.test.sh
+++ b/scripts/__tests__/channel-inbox-drain.test.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Contract test for scripts/hooks/channel-inbox-drain.py.
+# Run: bash scripts/__tests__/channel-inbox-drain.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+HOOK="$INSTALL_DIR/scripts/hooks/channel-inbox-drain.py"
+
+echo "channel-inbox-drain tests"
+echo "========================="
+
+OUT="$(python3 "$HOOK" --self-test 2>&1)"
+STATUS=$?
+if [ "$STATUS" -eq 0 ]; then
+  pass "self-test exits 0"
+else
+  fail "self-test exits $STATUS: $OUT"
+fi
+
+case "$OUT" in
+  *"channel-inbox-drain self-test passed"*) pass "self-test reports success" ;;
+  *) fail "self-test did not report success: $OUT" ;;
+esac
+
+echo ""
+echo "========================="
+TOTAL=$((PASS + FAIL))
+echo "Results: $PASS/$TOTAL passed"
+if [ "$FAIL" -gt 0 ]; then echo "FAILED: $FAIL tests"; exit 1; fi
+echo "All tests passed."
--- a/scripts/channel-inbound-tee.mjs
+++ b/scripts/channel-inbound-tee.mjs
@@ -1,0 +1,116 @@
+#!/usr/bin/env node
+import { spawn } from 'node:child_process'
+import { mkdirSync, appendFileSync, writeSync } from 'node:fs'
+import { dirname, join } from 'node:path'
+import { StringDecoder } from 'node:string_decoder'
+
+const childArgs = process.argv.slice(2)
+if (childArgs.length === 0) {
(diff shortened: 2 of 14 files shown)


Szotasz:
Thanks for splitting this out — the gate is exactly right now (default OFF, gated before the poll), which addresses the earlier concern.

It can't merge yet though: the prerequisite it reacts to (the tee-writer + drain-hook that would signal an undrained inbox) doesn't exist in this PR or on develop, so the nudge can never actually fire — it's dead scaffolding as-is. There are also 2 vitest failures (config-mock, one-line fix).

Two paths: either bring the tee+drain side into scope too (that'll need its own review, since message content would be written to disk), or park this until that lands. Either way, please fix the two tests. The gate work itself is solid.

Szotasz:
Thanks for this. The wake-nudge core is genuinely well designed: it is default-off, skips the main session, is content-free, gates on the idle gap, and debounces. That part passes our threat-model review as-is.

We can't merge it as submitted yet, because the second commit goes beyond what the PR description says (opt-in, zero behaviour change). Specifically:

1. `agent-process.ts` switches ALL Telegram sub-agent delivery to the tee -> file -> next-turn-drain model unconditionally (no flag). Combined with the wake being off by default, an idle sub-agent's incoming Telegram can sit in the file indefinitely, which is a behaviour regression on the default path.
2. The drain hook is added to the MAIN `settings.json`, which introduces a default-on file-drop into the main prompt. We deliberately keep the main session's injection surface minimal (after the phantom-injection hardening), so we can't widen it by default.
3. The config comment ("prerequisite not yet upstream") is stale and is contradicted within the same PR.

To get this to a merge, could you:

1. Put the delivery-path change (tee/mcp-swap plus template hook) behind its own opt-in flag, or split it into a separate PR that describes it honestly.
2. Exclude the MAIN session from `channel-inbox-drain.py` (an agent_id/cwd guard, mirroring `inbox-drain.py`).
3. Fix the stale config comment.
4. Add `channel-inbox-drain.py` to `KNOWN_HOOK_SCRIPTS` (otherwise the self-heal prune treats it as foreign).
5. Add a max-attempt budget / backoff to the nudge (as written it would nudge a never-draining session every 60s indefinitely).
6. Suppress the `--channels` flag for mcp.json-driven agents (the no-double-poller currently relies only on `enabledPlugins=false` and is untested end to end).

If you prefer, we can also land just the nudge component (default-off, main-excluded) on its own and take the delivery-path rework separately. The nudge half is ready.

Thanks again. This is a useful direction and we want it in, just cleanly separated from the default-path change.

norbull:
Köszönjük a részletes visszajelzést és a konkrét listát!

Mind a 6 pontot implementáljuk egy PR-ban:

1. **tee+drain opt-in flag:** `SUBAGENT_INBOX_TEE=true` env var, default false -- a delivery-path change csak ha explicit engedélyezve van.
2. **channel-inbox-drain.py main kizárás:** agent_id/cwd guard hozzáadva, mirroring inbox-drain.py mintáját.
3. **Elavult config komment:** javítva.
4. **KNOWN_HOOK_SCRIPTS:** channel-inbox-drain.py hozzáadva, hogy a self-heal prune ne kezelje idegenként.
5. **Nudge backoff:** max-attempt budget + exponenciális backoff, hogy ne nudgoljon 60s-enként végtelen ideig.
6. **--channels flag suppression:** mcp.json-vezérelt ágenseknél elnyomva, end-to-end teszteléssel.

Hamarosan frissítjük a branch-et.

PR #606, opened 2026-07-14
feat(message-router): session-stuck detection + reconnect-backlog batching

iszzu80-dev:
## What

Two hardening improvements to the message router:

**(a) Session-stuck detection:** Track continuously-unready sessions in the router tick; escalate a warning-level log after `STUCK_ESCALATE_MS` (10 min) so existing revival tooling can act before the message backlog grows. State cleared on session-ready or session-absent.

**(b) Reconnect-backlog batching:** When a session was absent and reconnects with >5 pending messages older than 30 min, summarize them into one batch delivery via a new agent_message and mark the originals done. Prevents the FIFO-burst pattern that made the Mason incident read like churn.

Also optimized `sessionExistsOnHost` to one call per unique receiver per tick (cached in the pre-pass) instead of one per message.

## Card
2922e380

## Files changed
- `src/web/message-router.ts` — main logic
- `src/__tests__/message-router-tick-cap.test.ts` — updated tests

### DIFF
--- a/src/__tests__/message-router-tick-cap.test.ts
+++ b/src/__tests__/message-router-tick-cap.test.ts
@@ -6,10 +6,10 @@
 // never make one tick run long and starve the event loop -- the slow-tick half
 // of the progressive-hang pattern.
 //
-// We feed 30 pending messages, all addressed to a SUB-agent whose tmux session
-// does not exist, and assert sessionExistsOnHost -- the first per-message probe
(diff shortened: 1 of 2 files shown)


Szotasz:
Duplikátum: a #613 (session-stuck+backlog) byte-azonos, kanonikus utódja ennek. Samu backlog-verify (BACKLOGSEQ) alapján zárva; a message-router fix a #613-on megy be.

PR #616, opened 2026-07-15
feat(scaffold): configurable WEB_PORT — run multiple Marveen instances per host

nortops:
## Configurable web port — run multiple Marveen instances on one host

**What this enables**

Marveen's dashboard port was hardcoded to `3420` across install scripts and scaffold templates, so a machine could only ever host a single instance without manual search-and-replace after every install. This PR makes the port a first-class, install-time setting, so several fully independent Marveen instances can run **side by side on one host** — each under its own OS user, each on its own port, in parallel.

**Changes**

- All hardcoded `localhost:3420` references in scaffold templates replaced with a `{{WEB_PORT}}` placeholder, resolved at install time.
- Both install scripts accept `--port` / `--web-port`; the default stays `3420`, so existing single-instance installs upgrade with zero config.
- `.env` is written with an explicit `WEB_PORT=<value>` entry, so runtime code reads the port from the environment rather than a hardcoded constant.

**Production validation (live, not just tests)**

Two fully independent Marveen instances are running right now on a single host, under two separate OS users, on ports `3420` and `3425` — both in active daily use. This is real end-to-end confirmation that parallel multi-instance operation works, not a test scenario.

---

### Bundled safety fix: boot-time hook pruning

A smaller, related hardening change that surfaced during the same multi-instance work — included here because it touches the same startup path. A stale or path-invalid hook entry in the shared user-scope `settings.json` can silently block **all** agent input across the fleet (a total input lockout with no operator-visible error). `boot-hook-prune` validates every registered hook path at startup and removes dead entries before the first prompt is accepted.

- **Fail-open**: if the prune step itself errors, the agent starts anyway with hooks untouched.
- **Audit log**: pruned entries are written to `store/hook-prune.log`.
- **Tests**: 312 lines of hook-path-guard coverage (path resolution, missing-file detection, fail-open edge cases).

---

### Testing

- Unit tests: 312 lines, all green (hook-path-guard suite).
- Code review: approved by a second reviewer before submission.
- Docker clean-room install: verified on a fresh image with a non-default port; install-to-running confirmed.
- Manual end-to-end: two instances reachable on different ports under different users, dashboards functional, hook pruning confirmed against a synthetic stale entry.

**Base branch:** `develop`

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,12 @@ cd marveen
 ./install.sh
 ```
 
+Alapértelmezés szerint a dashboard a 3420-as porton indul (`http://localhost:3420`). Egyedi port beállításához:
+
+```bash
+./install-linux.sh --port 3421   # vagy: WEB_PORT=3421 ./install-linux.sh
+```
+
 ### Windows (WSL)
 
 ```powershell
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -397,7 +397,7 @@ A főbb konfigurációs változók a launchd plist-ben (`~/Library/LaunchAgents/
 | `ALLOWED_CHAT_ID` | Az egyetlen engedélyezett Telegram chat ID |
 | `SLACK_BOT_TOKEN` | Slack bot token (ha Slack provider) |
 | `SLACK_CHANNEL_ID` | Slack csatorna ID |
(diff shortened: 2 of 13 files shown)


cett:
check it: https://github.com/Szotasz/marveen/pull/600

PR #653, opened 2026-07-17
fix(costops): Render sync ReferenceError + aggregated provenance summary (card d2631bad)

iszzu80-dev:
Two-part fix, both verified live via headless Chromium against the real dashboard.

(1) Render sync ReferenceError bug: 'Render frissítés' button called authHeaders(), a function that doesn't exist. Removed the explicit headers option -- window.fetch is already globally patched to auto-inject Authorization.

(2) Aggregated provenance summary: Added a compact block under the 4 headline stat cards showing actual data composition (from confidence_breakdown) and forecast basis distribution (computed client-side from all_sources).

Bonus: Fixed confLabel map missing actual_invoice/billing_export/local_usage keys, and deduplicated the map.


iszzu80-dev:
Closing as superseded, and with an apology for the noise.

This PR grew far beyond its title. Alongside the CostOps work it carries 17 files that should never have been proposed upstream: agent deliverable notes under agents/, seven internal working documents under docs/, and eight fleet-operations scripts under scripts/ (context-guard, memory-gate, safe-start, host-restart-watchdog, evidence-gate, unit-fail-notify, verdict-posted). Those are specific to how our own agent fleet is run and are of no use in this project. That was our mistake in scoping, not a disagreement about what belongs here.

It has also been CONFLICTING since 2026-07-17 with no review activity, so nothing is lost by starting clean.

The CostOps content here is fully superseded by a scoped split we are preparing, which will be raised in reviewable pieces on top of #628 rather than as one 138-file change.

Two files in here are legitimate and genuinely upstream-general -- src/config-registry.ts and the src/config.ts branding/timezone additions. They are not CostOps and will be raised separately as their own small PR.

#628 is unaffected and stays open.

PR #657, opened 2026-07-20
merge: P0 monitor release-hardening — branch-independent release + dead-guard detection

iszzu80-dev:
## Changes

- P0 memory-pressure guard: independent monitor + global gate + active relief
- Dead-guard detection: the monitor now detects when the systemd timer is active but producing dead/empty outputs
- Branch-independent release: systemd unit runs from immutable versioned release behind atomic symlink (releases/monitor-current -> monitor-98222ef28), decoupled from checked-out branch
- Fix: include memory-pressure-health.js in release bundle
- Fix: add releases/ to .gitignore — survive git clean -fd
- Vitest integration: throw Error instead of process.exit, hermetic state file isolation
- Telemetry correctness: single measurement source, honest semantics

## Verification
- Timer active, Result=success, state normal, measurement OK
- Release running: monitor-98222ef28
- 12 tests across 4 files green


iszzu80-dev:
Closing — will re-raise as a scoped feature branch (monitor-only), matching the shape of #653 and #655. The current PR is develop→develop with 370 files; the actual P0 monitor content is unreviewable. Internal audit/deliverable files also need to stay out of upstream.

PR #660, opened 2026-07-20
feat(costops): monitoring + financial operations + provenance-aware cost resolution (stacked on #628)

iszzu80-dev:
Continues the CostOps series. **Stacked on #628** — please review that one first.

### Reading this PR

The diff shows 89 files / 15 commits against `develop`, but **only the top 3 commits are new here**. The rest arrive via #628 and will disappear from this diff once #628 merges:

| new | commit | what |
|---|---|---|
| ✔ | `2cfee72` | monitoring — alerts, alert store, warnings, subscriptions, lifecycle, limits, expiry + live checks |
| ✔ | `4fc7bc3` | financial — period close, invoice, budgets, optimization, export, inventory, reliability, recommendations + the full `costs.ts` route surface |
| ✔ | `2c84f05` | accounting rule for equal-confidence cost lines (see below) |

Scoped to 57 files on top of #628. Happy to split monitoring and financial into separate PRs if you would rather review them apart — they are stacked, so the split only helps if #628 lands first.

### The accounting rule (`2c84f05`)

When two cost lines for the same source carry equally trustworthy provenance (a real invoice and a provider API figure), something has to decide which one represents that period's cost. This resolves it by what the source *is*, using the existing `cost_sources.source_type`:

- **Package / subscription** (and saas, hosting, domain): the invoice states the period it covers, so the **invoice** is the period cost. Provider API numbers are a stand-in until the invoice arrives.
- **Top-up / prepaid usage** (`source_type: 'usage'`): funding an account is a real payment but not the period's cost. Only what was **consumed** belongs to the period, and consumption is what the provider API reports.

Freshness now only breaks ties between two lines of the *same* kind, and a `data_freshness` in the future can no longer win — a period-end date reaching that column is a data defect we hit in practice, and it must not decide a figure.

When `source_type` is unknown the package branch applies, so the invoice wins. That default is deliberate: falling back to a real billed amount understates nothing, whereas defaulting to consumption would under-report, and an under-reported cost figure is the one nobody investigates.

### Verification

141 test files, 1687 tests, all passing. TypeScript clean. The accounting rule ships with tests in both directions plus the future-timestamp case, and they were confirmed to go red with the rule disabled.

### Not included, deliberately

Dashboard UI (`web/costops/*` and the `web/app.js` changes) is held back — that diff carries fleet-specific local features and needs a separate extraction pass. Nothing in this PR depends on it.

PR #661, opened 2026-07-20
chore(config): settings registry + branding/timezone config, drop stale entries

iszzu80-dev:
Two small, self-contained changes that are not CostOps and were previously entangled with a large CostOps branch. Raising them separately so they can be reviewed on their own merits.

- `src/config-registry.ts` — dashboard settings registry
- `src/config.ts` — `APP_TZ`, `BRAND_LOGO_URL`, `BRAND_ACCENT`, `OWNER_NAME_PLACEHOLDER`, and removal of stale entries (`HEARTBEAT_CALENDAR_*`, `SCHEDULER_TZ`, `IDEAS_DISABLED`)

These make a fresh install easier to brand and to run in a non-Budapest timezone without editing source. No behaviour change for an existing deployment.


iszzu80-dev:
Closing immediately — I opened this in error and it does not contain what the title says.

It was meant to be a two-file change (src/config-registry.ts and src/config.ts). It is branched from our local develop rather than from your develop, so it carries 136 commits and our entire local divergence: agent working notes under agents/, internal audit documents under audits/, and fleet operations scripts. None of that belongs here.

That is the same mistake as #657, and this time it is mine specifically: I verified the contents of #660 carefully and opened this one on a colleague's description without checking the diff first.

Will re-raise the two files properly, branched from your develop.

PR #708, opened 2026-07-25
fix(scheduler): reject an invalid SCHEDULER_TZ instead of scheduling into a void

huszarrichard:
## The bug

A misspelled timezone (`SCHEDULER_TZ=Europe/Budapesst`) silently kills **every** scheduled task.

The chain:
1. `cfg('SCHEDULER_TZ')` -> `APP_TZ` (`src/config.ts:49`) -> `CRON_TZ` (`src/web/cron.ts`). No validation anywhere.
2. `CronExpressionParser.parse(expr, { tz: 'Europe/Budapesst' })` throws: `CronDate: unhandled timestamp: Invalid Date`.
3. `cronDueBetween`'s `catch { return false }` turns that throw into **"not due"**.

Result: no error, no warning, no failed task -- every cron expression simply never matches, forever. The startup report even looks healthy, because the configured value legitimately won its resolution layer. A single typo produces a **total, silent** scheduler outage.

This is the same failure class as #707 (the reporter disagreeing with the logic), but one layer down and with a worse blast radius: #707's UTC fallback is a *partial* outage (fixed-time crons misfire, interval crons survive), this one is total.

## Reachability

The dashboard Settings path is **already** fenced: `SCHEDULER_TZ` carries a `valueSet` and `validateSettingValue` enforces it on write. What is unfenced is a **hand-edited `.env`** or a hand-written `config-overrides.json` (an exported `SCHEDULER_TZ` does NOT reach here -- `readEnvFile()` parses the `.env` file only and never consults `process.env`) -- and hand-editing `.env` is the documented way to set the zone, so this is the likely path in practice.

## The fix

`resolveAppTz()` validates the configured zone once at boot by probing the actual consumer (`CronExpressionParser.parse(...)`) and **degrades to the process zone** -- exactly the behaviour of leaving `SCHEDULER_TZ` unset. The scheduler stays alive on a wrong-but-working clock instead of dying invisibly.

The rejected value is exported as `APP_TZ_INVALID` so `startScheduleRunner` can name it in a warning. `config.ts` cannot own a logger itself (`logger` imports `config` -> circular), so the loud reporting lives in the runner.

Note the fix deliberately lands in `config.ts`, **not** in `resolveCronTz`: `CRON_TZ` is `APP_TZ`, so validating only the reporter would make the *log* honest while leaving the outage fully intact.

## Tests

8 new tests in `src/__tests__/config-app-tz.test.ts` -- the resolver directly (valid / unset / misspelled / non-IANA offset / real-process-zone default) plus the outage it prevents:

```
cronDueBetween('0 8 * * 1', ..., 'Europe/Budapest')  === true   // a due occurrence
cronDueBetween('0 8 * * 1', ..., 'Europe/Budapesst') === false  // silently swallowed
cronDueBetween('0 8 * * 1', ..., resolveAppTz(...).tz) === true // validated -> fires again
```

25/25 green together with the existing `cron.test.ts`; `tsc --noEmit` clean.

## Relation to #707

Independent. The new warn block is *appended after* the existing UTC-fallback warn rather than folded into it, so it does not collide with #707's rewrite of that reporter. Happy to rebase whichever merges second.

Credit: the failure mode was spotted during review of #707.

### DIFF
--- a/src/__tests__/config-app-tz.test.ts
+++ b/src/__tests__/config-app-tz.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect } from 'vitest'
+import { CronExpressionParser } from 'cron-parser'
+import { resolveAppTz } from '../config.js'
+import { cronDueBetween } from '../web/cron.js'
+
+// Regression for the failure mode one layer below the 2026-07-13..15 outage:
+// a MISSPELLED SCHEDULER_TZ ("Europe/Budapesst") used to flow straight into
+// APP_TZ -> CRON_TZ. cron-parser throws on an unknown zone, cronDueBetween
+// catches the throw and reports "not due", and so every scheduled task stops
+// firing -- total, silent, and with a startup report that still looks healthy
+// because the configured value "won" its resolution layer.
+//
+// resolveAppTz validates the configured zone at boot and degrades to the
+// process zone (the same behaviour as leaving SCHEDULER_TZ unset) instead of
+// scheduling into a void, keeping the rejected value for the startup warn.
+//
+// The validity probe is cron-parser itself, NOT Intl. These tests therefore
+// assert agreement with the consumer rather than with a particular ICU
+// version: `+02:00` is accepted here because cron-parser schedules against it
+// correctly, even though older ICU builds reject it as a zone name. Asserting
+// the Intl answer instead made this suite engine-dependent (red on Node 22).
+
+const SYSTEM = 'Europe/Budapest'
+
+describe('resolveAppTz', () => {
+  it('accepts a valid configured zone and reports it as configured', () => {
+    expect(resolveAppTz('America/New_York', SYSTEM)).toEqual({
+      tz: 'America/New_York',
+      configured: 'America/New_York',
+    })
+  })
+
+  it('falls back to the system zone when unset, with no configured marker', () => {
+    expect(resolveAppTz(undefined, SYSTEM)).toEqual({ tz: SYSTEM })
+    expect(resolveAppTz('', SYSTEM)).toEqual({ tz: SYSTEM })
+  })
+
+  it('rejects a misspelled zone and reports the rejected value', () => {
+    expect(resolveAppTz('Europe/Budapesst', SYSTEM)).toEqual({
+      tz: SYSTEM,
+      invalid: 'Europe/Budapesst',
+    })
+  })
+
+  it('does not report a rejected zone as configured (the reporter must not claim it won)', () => {
+    expect(resolveAppTz('Europe/Budapesst', SYSTEM).configured).toBeUndefined()
+  })
+
+  it('uses the real process zone as the default second argument', () => {
+    const r = resolveAppTz('Europe/Budapesst')
+    expect(r.invalid).toBe('Europe/Budapesst')
+    expect(r.tz).toBe(Intl.DateTimeFormat().resolvedOptions().timeZone)
+  })
+})
+
+describe('the probe agrees with the consumer', () => {
+  // The whole point of probing with cron-parser: whatever the scheduler can
+  // actually schedule against must survive validation, and nothing else.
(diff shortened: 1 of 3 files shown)


Szotasz:
Thanks for this one, and for #707 which is now merged.

We verified this branch in an isolated worktree on top of the current develop. The runtime behaviour is exactly what we want, and better than the title suggests: there is no hard failure anywhere. An invalid zone falls back to the process zone and logs a loud startup warning that names both the rejected value and the substitute, and the unset or empty path is byte-identical to the previous code. That is the right shape for a scheduler, because a typo in a config value should never be able to silently take the whole automation down.

One thing blocks the merge, and it is only the test.

`rejects a non-IANA value (a UTC offset is not a zone name)` fails on Node 22.22.3. Newer ICU accepts offset strings, so `resolveAppTz('+02:00')` returns `{ tz: '+02:00' }` with no invalid marker. This is not a functional regression, since cron-parser also accepts `+02:00` and schedules against it correctly, but the assertion is engine dependent and lands red on our host, which would turn the develop suite red on every Node 22 machine including our soak gate.

Two ways out, your pick:

1. Loosen that assertion so it tolerates both engine behaviours.
2. Probe validity with cron-parser itself rather than with `Intl`, so the guard matches the actual consumer exactly. On our host the two agree, but this removes the divergence class entirely.

Two small documentation notes while you are in there, neither blocking:

- The PR body says an exported `SCHEDULER_TZ` reaches this code. It does not: `cfg()` layers `config-overrides.json` over `.env` and never reads `process.env`. This is also what #707 fixed on the reporting side.
- The `Intl` probe is a proxy for the real consumer. See point 2 above.

Once the test is adjusted, re-verification should be quick since only the test file changes. Note that #707 rewrote the surrounding block in `cron.ts`, so a small rebase will be needed first.

huszarrichard:
Thanks -- rebased onto develop (which now has #707) and pushed. Both notes were right, and the test one was right in a way I want to be explicit about.

**The `Intl` probe was the same bug I was fixing.** A guard that resolves through a different mechanism than the code it protects is exactly the divergence this patch exists to remove, and I reproduced it inside the fix. Took your option 2: the probe is now `CronExpressionParser.parse('0 0 * * *', { tz }).next()`, so whatever the actual consumer can schedule against is by definition what survives validation.

**On the failing test -- your diagnosis holds, but the conclusion goes one step further.** I could not reproduce your result after switching probes, because cron-parser is *also* engine-dependent here: it delegates zone resolution the same way, so `+02:00` is rejected on my Node 20.20.2 and accepted on your Node 22.22.3. Pinning either verdict just moves which host goes red -- my first rewrite was green for you and red for me.

So the test no longer asserts a verdict for `+02:00`. It asks cron-parser what it thinks, then asserts the **invariant**: whichever way the consumer goes, the guard goes the same way. That is deterministic on every engine by construction, and its violation is the only thing that would reintroduce the bug. Fixed-verdict cases stay for the unambiguous inputs (`Europe/Budapest`, `UTC`, `Etc/GMT-2`, `Europe/Budapesst`).

**One more change you did not ask for, because #707 and #708 compose badly without it.** `SCHEDULER_TZ_CONFIGURED` now carries the *validated* value rather than the raw one. With the raw value, an invalid zone would make `effectiveCronTz()` report the zone the scheduler had just refused as the one in effect -- and it would also suppress your UTC-fallback warning, since `source` would read `'SCHEDULER_TZ'` instead of `'system-default'`. Same failure class as the rest of this patch, one layer up. Credit to Blitz, who spotted the interaction while reviewing the two branches together.

**Documentation note corrected.** You are right that an exported `SCHEDULER_TZ` never reaches `config.ts` -- `readEnvFile()` only parses the `.env` file and never consults `process.env`. I verified that before changing the text rather than taking it on faith. The PR body now claims only the two paths that actually reach it: a hand-edited `.env` and a hand-written `config-overrides.json`.

31 tests green with `cron.test.ts`, `tsc --noEmit` clean. Only `config.ts`, the runner's warn block, and the test file changed.

PR #720, opened 2026-07-26
fix(governance): stop self-pace gate false positives and sub-agent 401s

tutker:
## What

Two independent, verified fixes to the fleet governance plumbing. Both are small, self-contained, and covered by tests.

### 1. self-pace gate false positives (`scripts/self-pace-gate.mjs`)

Three deny-rules matched a keyword that is **also** a normal English word or a path segment, so harmless commands were blocked:

| Rule | Collision | Real-world false deny |
|------|-----------|-----------------------|
| `/loop` | `\bclaude\b` matches `.claude` in every memory/skill path, so `\/loop\b` fired on any `loop`-prefixed path component | reading `.../memory/loop-*.md` |
| `at` / `batch` | English words at a segment start (splitSegments splits on newlines) look like the `at(1)`/`batch(1)` binaries | `git commit` with a body line "at least 80% ..." / "batch size is 50" |
| `launchctl` | appears as a launchd **job label** in the fleet's own heartbeat status reports | a status report line `... ; launchctl com.jarvis.channels ...` |

**Fix:** match the invocation *shape*, not the bare keyword (segment-start / flag / redirect / at-timespec / launchctl-subcommand). Every real self-pace vector stays denied.

### 2. sub-agent dashboard token 401 (`src/web/agent-scaffold.ts`)

Generated `CLAUDE.md` curl examples used a **relative** `store/.dashboard-token`. Sub-agents run from `agents/<name>/`, where that relative path does not exist, so `curl` sent an empty Bearer and every `/api` call `401`'d silently, killing sub-agent memory saves and searches. **Fix:** emit an absolute `PROJECT_ROOT`-joined token path.

## Testing

`src/__tests__/governance-gates.test.ts` adds regression coverage for the prose/path collisions and asserts the real vectors stay denied.

```
vitest run src/__tests__/governance-gates.test.ts
Tests  68 passed (68)
```

Verified on top of current `develop` (in a worktree outside `/tmp`, since the gate's own hook-safety check rejects `/tmp` paths).

## Update (review round 2)

Addressed the two regressions caught in review. Both forms EXECUTE `claude /loop`
in bash but slipped the new slash-command-position match (the char before `/loop`
was a backslash or the tail of `$IFS`, neither in the `[\s'"]` class):

- `claude \/loop` (bash resolves the escaped slash)
- `claude$IFS/loop` (`$IFS` word-splits into a real space at exec time)

Per the suggested direction, the segment is now **normalised before the pattern
runs** (resolve `\X` -> `X`, substitute `$IFS`/`${IFS}` with a space) rather than
widening the character class; this also closes the `claude /lo\op` mid-token form.
Scoped to the self-pace bash patterns only; the scheduler/store/API checks keep
the raw segment. Both forms now deny end-to-end through the gate hook, with a
regression test for each (plus the mid-token side effect and a memory-path allow
guard) in the governance suite.

The three quote-concatenation forms noted as pre-existing (not a regression from
this change) are intentionally left out of scope, as requested.

## Note

Opened as a proposal for review. It has had a second review on our side; that is
context, not a gate on your merge. Please merge at your discretion once the review
is satisfied.

### DIFF
--- a/scripts/self-pace-gate.mjs
+++ b/scripts/self-pace-gate.mjs
@@ -211,6 +211,27 @@ export function stripGitCommitMessages(command) {
   )
 }
 
+// Normalise two shell-level obfuscations that bash resolves at EXEC time, so an
+// invocation whose SHAPE is a real self-pace cannot dodge the slash-command match
+// with quoting the shell undoes anyway. Measured end-to-end through the gate hook
+// (upstream review, 2026-07-27): `claude \/loop` and `claude$IFS/loop` BOTH run
+// `claude /loop` in bash but slipped the `(?:^|[\s'"])\/loop` match -- the char
+// before `/loop` was `\` and `S` (end of `$IFS`), neither in the [\s'"] class.
+// The fix is NOT to widen that class (that would let more prose through); it is to
+// resolve what the shell resolves before matching: `$IFS`/`${IFS}` word-splits to
+// a space, and a backslash escape `\X` collapses to `X`. Side effect: also closes
+// `claude /lo\op`. Applied ONLY to the self-pace bash patterns below; the
+// scheduler/store/API checks keep the raw segment (upstream measured them clean,
+// and this PR is scoped to these two loop regressions). This cannot introduce a
+// false positive: collapsing escapes / dropping `$IFS` never synthesises the
+// literal `tmux`+send-keys, `nohup`+claude, or `claude`+`/loop` tokens out of
+// prose -- it only removes an evasion.
+export function normalizeShellEvasion(seg) {
+  return String(seg ?? '')
+    .replace(/\$\{IFS\}|\$IFS\b/g, ' ') // $IFS / ${IFS} -> the space it expands to
+    .replace(/\\(.)/g, '$1') // \X -> X (bash unescape of a backslash-escaped char)
+}
+
 // Pure decision: does this tool call set up self-pace / self-injection?
 export function gateDecision(toolName, toolInput) {
   const name = String(toolName ?? '')
@@ -232,7 +253,10 @@ export function gateDecision(toolName, toolInput) {
     // Per-segment so an unrelated token elsewhere in a compound command cannot
     // turn a legit read (store inspection, schedule-API GET) into a false deny.
     for (const seg of splitSegments(safeCommand)) {
-      if (SELF_PACE_BASH_PATTERNS.some((re) => re.test(seg))) return { deny: true }
+      // Match the self-pace bash patterns against the shell-normalised segment so a
+      // `\/loop` / `$IFS/loop` evasion (which bash resolves to `/loop` at exec) is
+      // still caught; the scheduler/store/API checks below use the RAW seg (scoped).
+      if (SELF_PACE_BASH_PATTERNS.some((re) => re.test(normalizeShellEvasion(seg)))) return { deny: true }
       // scheduler binaries: deny the exec/submit forms, allow pure read-listing
       if (SCHEDULER_RX.test(seg) && !SCHEDULER_READ_RX.test(seg)) return { deny: true }
       // self-schedule store: block WRITE only (a read/grep is legit diagnostics)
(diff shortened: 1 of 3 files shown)


Szotasz:
Thanks for this one. The diagnosis is right and the direction is right: matching the invocation SHAPE instead of a bare keyword is exactly how these false positives should be fixed, and the three cases you documented are real. We reproduced all of them, including the prose-in-a-commit-body case, which has bitten us in practice.

We verified your claim that every real form stays denied, by actually driving the gate rather than reading the patterns. Your four listed forms are all correctly denied, and so are several you did not list (extra whitespace, `claude "$X" /loop`, backtick prefix, `$'...'` quoting). The previously-broken path cases now pass. That part is solid.

**One change requested before merge: two forms regress.** These were caught by the old `\/loop\b` pattern and are allowed by the new one, and both actually execute `claude /loop` in bash:

- `claude \/loop` — the backslash-escaped slash has no `[\s'"]` before it, so the new lookbehind class does not match. Bash resolves `\/` to `/`.
- `claude$IFS/loop` — `$IFS` word-splits into a real space at execution time, but the literal `$` is not in `[\s'"]`.

Both were confirmed end-to-end through the gate hook, not just by reading the regex.

Suggested direction: normalise the segment before the pattern runs, rather than widening the character class. Resolving backslash escapes (`\X` -> `X`) and substituting `$IFS` with a space closes both, and as a side effect also closes `claude /lo\op`.

Separately, and **not** something we are asking you to fix here: three quote-concatenation forms (`claude /'loop'`, `claude "/"loop`, `claude /lo\op`) slip through the OLD pattern as well. That is pre-existing, not a regression from your change, and closing it properly needs quote-stripping, which interacts with the data-payload stripping and deserves its own change. Please keep this PR scoped to the two regressions.

Everything else measured clean: the at/batch boundary still denies every real invocation we could think of (`at now + 1 minute`, `at teatime`, `echo ... | at midnight`, bare `batch`, `batch < file`) while letting prose through; launchctl still denies load/bootstrap/kickstart and the bare-word `start` subcommand form; the scaffold path is runtime-derived from the install root, so it stays correct on any install; and the full suite is green (2781 passing, plus your 64 governance tests).

Add the two fixes with a test case for each in the governance suite and we will re-verify and merge.

Szotasz:
One more thing, separate from the code, and it needs clearing up before merge.

The last line of the PR description reads: "Do not merge without Renata's approval."

We have no Renata. Not in the fleet, and not among this repository's collaborators (we checked). So as it stands, the PR asks us to wait on an approval that nobody here can give, or worse, invites someone to merge later while citing an approval that never happened.

Please either tell us who that refers to, or drop the line. If it is a reviewer on your side, that is fine and we are glad the change got a second pair of eyes, but it is not a gate on our merge, and it should not read like one.

Worth saying plainly, because this is the second PR today where a line from someone else's process ended up in a description aimed at us: another PR claimed to implement "the approved concept #281", and #281 turned out to be an unrelated closed issue that nobody had approved. We check these, every time. An unverifiable claim in a description costs more trust than an unfinished feature does, and it slows down exactly the PRs we would otherwise be happy to merge.

None of this affects the technical verdict above. Fix the two regressions, drop or explain the line, and this goes in.

tutker:
Thanks for the thorough review, and for driving the gate rather than reading the patterns.

Pushed `2fd8944` addressing both regressions. Per your suggested direction, the segment is normalised **before** the pattern runs instead of widening the character class:

- resolve backslash escapes `\X` -> `X`
- substitute `$IFS` / `${IFS}` with a space

That closes `claude \/loop` and `claude$IFS/loop`, and as you predicted also closes the `claude /lo\op` mid-token form. It is scoped to the self-pace bash patterns; the scheduler/store/API checks keep the raw segment, so the at/batch/launchctl behaviour you measured clean is untouched. Normalisation can only remove an evasion, never synthesise the trigger tokens out of prose, so no new false positive.

Regression tests added for both forms plus the mid-token side effect and a memory-path allow guard; the governance suite is green (68) and I confirmed both forms deny end-to-end through the gate hook process, not just via the regex.

Left the three quote-concatenation forms out of scope as you asked (pre-existing, needs quote-stripping that interacts with the data-payload stripping).

On the description line: dropped the merge-gate wording. It referred to an internal reviewer on our side and should never have read as a gate on your merge. Merge at your discretion once you have re-verified.

PR #722, opened 2026-07-26
security(vault): MCP titkok vault-védelme a connectors telepítésnél és remote headereknél

balkazoli:
Két biztonsági javítás a marveen vault-integrációjához (Balkányi Zoltán telepítésén élesben futnak és teszteltek):

**1. \`feat(vault): support remote MCP header secrets via headersHelper\`**
A remote (URL-transportos) MCP szerverek header-titkai (pl. \`Authorization: Bearer <token>\`) eddig nem tudtak vault-hivatkozásként működni. Új \`scripts/vault-headers-helper.mjs\` + \`.sh\`: a headert indításkor oldja fel a vaultból, így a token nem kerül plaintextben a configba.

**2. \`fix(connectors): vault install-time secrets (Flag A) + honor remote transport (Flag C)\`**
- Flag A: a dashboard Connectors telepítő modálja a megadott API-kulcsokat eddig PLAINTEXTBEN írta a \`.claude.json\`-ba. Mostantól a vaultba kerülnek, a config csak \`vault:<id>\` hivatkozást kap.
- Flag C: a katalógusban remote transportúként megjelölt MCP-k telepítéskor tényleg remote-ként kerülnek bekötésre (eddig local commandra esett vissza).

Füst-teszt: telepítés után \`grep\` a nyers titokra a configokban = 0 találat, a szerverek vault-feloldással indulnak.

### DIFF
--- a/scripts/vault-headers-helper.mjs
+++ b/scripts/vault-headers-helper.mjs
@@ -1,0 +1,43 @@
+#!/usr/bin/env node
+// Resolve vault-backed MCP request headers at connection time.
+//
+// Claude Code runs a remote MCP server's `headersHelper` fresh on every
+// connection (and again on 401/403) and uses the JSON object printed to stdout
+// as the request headers. This is the remote-server analogue of
+// vault-env-wrapper.sh: the secret is resolved from the Vault at connection
+// time and NEVER written to .mcp.json -- only the vault secret id is on disk.
+//
+// Args: one per header, in the form  HeaderName=Scheme:::vaultSecretId
+//   - Scheme is optional (may be empty); when present the header value becomes
+//     "Scheme <secret>" (e.g. "Bearer <token>"), otherwise the raw secret.
+//   - The ":::" separator keeps every argument free of spaces so the string is
+//     safe when Claude Code passes headersHelper through a shell.
+// Output: a JSON object {HeaderName: "value", ...} on stdout.
+import { join, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const projectRoot = join(__dirname, '..')
+
+// Same pattern as vault-resolve.mjs: import the compiled vault module.
+const { getSecret } = await import(join(projectRoot, 'dist', 'web', 'vault.js'))
+
+const headers = {}
+for (const arg of process.argv.slice(2)) {
+  const eq = arg.indexOf('=')
+  if (eq < 0) continue
+  const headerName = arg.slice(0, eq)
+  const rest = arg.slice(eq + 1)
+  const sep = rest.indexOf(':::')
+  const scheme = sep >= 0 ? rest.slice(0, sep) : ''
+  const vaultId = sep >= 0 ? rest.slice(sep + 3) : rest
+  if (!headerName || !vaultId) continue
+  const secret = getSecret(vaultId)
+  if (secret === null) {
+    process.stderr.write(`vault-headers-helper: vault secret "${vaultId}" not found\n`)
+    continue
+  }
+  headers[headerName] = scheme ? `${scheme} ${secret}` : secret
+}
+
+process.stdout.write(JSON.stringify(headers))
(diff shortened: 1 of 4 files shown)


Szotasz:
We are verifying this now and intend to merge it today. The exposure it closes is active and the affected population grows with every dashboard MCP install that has a key typed into it, so waiting is not the neutral option here.

One question before we do, and you are the only person who can answer it from real data.

Our concern is not the new functionality -- it is the `syncSecret` / `unsyncBinding` / `removeBindingsForSecret` rewrite, because that code runs against bindings that **already exist** on installs where everything currently works. The worst outcome of this PR is not that a new install misbehaves; it is that a working MCP config breaks on an install that had no problem before.

We cannot measure that here: this host has no `store/vault-bindings.json` at all, so we have no existing-binding population to regress. Our regression coverage is therefore synthetic, and we would rather say that out loud than let a green run imply more than it proves.

You mention this runs live on an install. So: **on that install, did the pre-existing env-var bindings survive the upgrade unchanged?** Concretely, after deploying this branch onto a config that already had `vault:` refs in place:

- do the previously bound servers still start and resolve their secrets,
- and is the `env` block of an untouched, env-only server byte-identical to what it was before?

A plain yes with a sentence on how you checked is enough. If the answer is that you only ever exercised it on freshly created bindings, that is also a useful answer -- it just means the real-population evidence does not exist yet, and we will say so on our side rather than pretend otherwise.

Separately: the code reads well. The fail-closed install path (server registered without the secret, so it is non-functional but leaks nothing) is the right direction for that failure to point, and rebuilding `headersHelper` from the full surviving binding set rather than the removed one avoids the obvious footgun of one secret's removal silently dropping another's header.

balkazoli:
Answering the regression question from #722, with real data from the install this runs on. Late, but measured rather than asserted.

**Short answer: yes for the pre-existing binding population, with one caveat I want to state myself rather than have you find it.**

## What existed before the change, on this host

The two commits are authored `2026-07-23 15:14` and `15:27` local time. The three vault
secrets that back the pre-existing bindings were created earlier the same day —
`12:51`, `12:52`, `12:53` — i.e. by the pre-change code path, roughly two and a half hours
before the rewrite landed here. So this install does have the population you could not
measure on yours: bindings that were created by the old code and then lived through the new one.

## (a) Do the previously bound servers still start and resolve their secrets?

Yes. Resolved just now, values never printed, only lengths:

| server | mechanism | resolves |
|---|---|---|
| `email-zoltan-garai` | `EMAIL_PASS` → `vault:` ref | yes (20 chars) |
| `email-hello-garai` | `EMAIL_PASS` → `vault:` ref | yes (20 chars) |
| `n8n` | `headersHelper`, `Authorization: Bearer` | yes (helper exit 0, 274-char bearer) |
| `fal-ai` | `FAL_KEY` → `vault:` ref | yes (69 chars) |

The `n8n` one is the case your `headersHelper` note was about, and it is the one that had
been quietly working since before the change: the helper still emits a complete
`Authorization` header from the surviving binding set.

## (b) Is an untouched, env-only server's `env` block byte-identical?

Yes, for every server that the change had no business touching. Comparing a config snapshot
taken `2026-07-15` (eight days before the rewrite) against the live config today, hashing
values rather than reading them:

`fetch`, `filesystem`, `fireflies`, `google-drive`, `markitdown`, `playwright`, `time` —
`command`, `args` and every `env` entry identical.

## The caveat, stated plainly

Four servers *did* change between that snapshot and today, and I will not pretend the diff
attributes cleanly to your PR — two weeks of unrelated work sit in between:

- `github` — I rewired it **tonight**, by hand, to move a token out of a plaintext config.
- `fal-ai` — moved to a `vault:` reference on `2026-07-19`, before these commits.
- `gmail`, `elevenlabs` — changed for reasons unrelated to the vault path; `elevenlabs`'s
  `env` block is in fact identical, only its `command` differs.

So the honest scope of the evidence: **no binding was lost, no untouched server drifted, and
every pre-existing binding still resolves** — but the "byte-identical" claim is demonstrated
on the servers the change should not have touched, not on the ones we ourselves edited since.

If you want a tighter test than a two-week-old snapshot allows, say so and we will take a
before/after pair around a controlled upgrade on this install and report that instead.

PR #729, opened 2026-07-27
feat(security): Bridge pairing from the dashboard -- enroll UI + per-device-key bundles

Szotasz:
## What

AUTHPLAN1 #2 (final Tuesday slice). The Settings > Security tab pairs a Bridge device end-to-end: paste the key line the Bridge shows, name the device, confirm, copy back the returned bundle. One enrollment = restricted `authorized_keys` entry + freshly minted per-device dashboard key + audit row + operator notification.

## Changes

- **Reuses `remote-enroll-core`/`remote-enroll-fs`** -- no second enroll implementation. The only new SSH-side logic is the revoke counterpart: `removeAuthorizedKey` (pure) + `removeEnrolledKey` (fs, same lock + atomic-write discipline as enroll).
- **`ssh-keyscan` runs ASYNC** (`execFile`): this code lives in the dashboard's event loop; the CLI's `execFileSync` would freeze every in-flight request. Host key resolves BEFORE any write, so a failed enrollment leaves zero side effects.
- **New bundles embed the minted device key** in `dashboardToken` instead of the shared token -- wire-compatible for the Bridge, but revoking this key cuts this device only. **Old bundles keep working unchanged** (backward compatibility untouched; they just don't gain per-device revocation).
- **`device_keys.install_id`** (additive `ALTER TABLE` try/catch) links a key to its SSH enrollment. Re-pairing the same installId replaces BOTH halves (ssh line merged by id, old key revoked + fresh minted) -- a device never accumulates keys.
- **Revoke = both halves at once**: `DELETE /api/auth/device-keys/:id` on a paired key also drops the `authorized_keys` line, idempotently, with its own `security.bridge_revoke` audit row.
- **Allowlist**: token+session only; `device` denied -- one leaked Bridge must not enroll further Bridges.
- **UI**: Bridge pairing section in the Security tab (key-line paste + name + confirm -> one-time copyable bundle), Bridge badge on paired device-key rows, hu+en strings.
- `MARVEEN_SSH_DIR` test seam so tests/e2e never write the real `~/.ssh`.
- docs/dashboard-auth-recovery.md extended with paired-revoke semantics.

## Merge gates (foreign installs)

No new requirement is on by default: pairing is operator-initiated opt-in; zero paired rows = nothing changes. No schema migration (additive column). The notification is silent without a wired channel. Documented recovery path unchanged + extended.

## Tests

- New `bridge-enroll` suite (12 tests): bundle embeds a WORKING device key (asserted through resolveAuth), replace-both-halves re-pairing, hard-fail-without-host-key leaves no side effects, foreign authorized_keys lines preserved byte-for-byte, revoke drops both halves + audits, 403 for device/federation/missing principals, 400 validation.
- 130 green across the 7 auth-related suites; full run 2810 passed (the same 13 known /tmp-worktree path-environment failures as #724/#726; green on a normal checkout).
- Browser e2e (Playwright, isolated instance, scratch ssh dir): pair via the UI -> bundle decodes with hostKey + `mvdk_` key -> embedded key authenticates as `device` -> revoke via UI -> key dead AND ssh line gone.

Remaining before Tuesday: tonight's live test with a REAL Bridge device on the fresh-install simulation -- that's the release gate, per plan.


Szotasz:
Reviewed from source, not from the description. The core of this is solid: the enrollment core is genuinely reused (no second enroll path), `ssh-keyscan` runs via async `execFile` and the resolved output is handed to the sync resolver as a closure, so the event loop is never blocked, and the revoke ordering is right (dashboard key first, SSH removal best-effort, no rollback that would resurrect a revoked key).

Two changes requested before merge. Both are small, and both are about a failure looking like a success.

**1. A partial revoke reads as a complete revoke in the UI.**

The route correctly returns `ssh_removed: false` when `removeBridgeSshAccess` throws, and it audits the failure. But the frontend revoke handler never reads that field, so the row disappears and the operator sees an ordinary success. In that state the dashboard key is dead while the `authorized_keys` line is still present, which means the device can still open the SSH tunnel. For a feature whose entire promise is "revoke means both halves at once", that is the one outcome the UI must not hide.

Please surface `ssh_removed === false` as a visible warning telling the operator that the SSH entry survived and has to be removed manually (the recovery doc already has the manual path to point at).

**2. The `MARVEEN_SSH_DIR` test seam is still ungated in a production code path.**

`defaultBridgeEnrollDeps()` and the revoke default both resolve `process.env.MARVEEN_SSH_DIR || join(homedir(), '.ssh')`, with only a comment stating it is unset in production. This ships to third-party installs. If that variable ever leaks into a real environment (an inherited env, a copied `.env`, a service definition carried over from a test host), enrollment silently writes to the wrong directory and the user sees "I paired it and it does not work", with nothing in the logs pointing at the cause.

Either restrict it to non-production, or, if it stays unconditional, emit a warning log and record it in the audit entry when the override is active, so an incident does not start with guessing. I flagged this before the UI work and it is unchanged, so I am holding the merge on it rather than filing a follow-up.

Everything else, including the re-pair replace-both-halves behaviour and the idempotent SSH removal, checked out against the tests and the code. The live pairing test with a real device and a fresh-install simulation is still the open gate for the release, and I am treating it as open.

Szotasz:
Re-checked 629c8c8. Point 2 is done and done well: the warning log is there and `sshdir_override=1` lands in both the enroll and the revoke audit entries, with tests asserting it.

Point 1 is still open. `ssh_removed` appears only in the response and in the tests; the revoke handler in `web/app.js` never reads it and there is no translation key for it. So the failure mode I described is unchanged: if the filesystem removal fails, the row disappears, the operator sees a plain success, and the `authorized_keys` line survives with the device still able to open the tunnel.

Holding the merge on that one. What it needs: surface `ssh_removed === false` as a visible warning (new hu+en key) telling the operator the SSH entry survived and must be removed manually, pointing at the recovery doc, plus a test covering that branch.

Szotasz:
Both change-request points addressed:

1. **Partial revoke is now visible** (d1b997d): the revoke handler reads `ssh_removed` from the DELETE response; `false` shows a persistent red warning above the device-key list pointing to the manual `authorized_keys` cleanup (docs/dashboard-auth-recovery.md). Browser e2e gained the failure path (pair -> delete the scratch authorized_keys -> revoke -> warning asserted + screenshot).
2. **MARVEEN_SSH_DIR seam** (629c8c8, crossed with your verdict): option (b) -- every use emits a warn log line, and both bridge audit rows carry `sshdir_override=1` (asserted in tests; the e2e instance's audit trail shows `added -> replaced -> ssh_removed=true`, all flagged).

Full extended e2e green (pair, re-pair replace-both-halves, full revoke, partial-revoke warning). Ready for re-review.

PR #737, opened 2026-07-27
fix(dashboard): correct the token-usage cost table

mogganhun:
Two spots assumed things about Claude model ids that stopped being true with Opus 5. Both fail **silently**, which is what makes them worth fixing.

## 1. Context guard treated Opus 5 as a 200k model

`contextLimitForModel()` inferred the window purely from the `[1m]` suffix, so anything without the tag fell through to 200k. Opus 5 is 1M by default *and* maximum, and carries no suffix.

The consequence is not cosmetic: the guard computes `pct` against a fifth of the real window, so a perfectly healthy session crosses the act threshold and gets a handoff request plus a forced restart it never needed. A restart discards the live working context, which costs far more than the check was protecting.

Replaced the suffix-only check with an explicit prefix list of the models that are 1M by default (Opus 5/4.8/4.7/4.6, Sonnet 5/4.6, Fable 5, Mythos 5). The `[1m]` suffix still works as an override for models where 200k is the default. Unknown ids stay conservative at 200k, and `calibrateLimit` steps them up on evidence exactly as before.

## 2. Cost table had no Opus 5 row, and prefix matching was order-dependent

`tuPriceForModel` returned the **first** key whose prefix matched. Since `claude-opus-4-8` also `startsWith('claude-opus-4')`, whichever key the object happened to list first decided the price — Opus 4.8 was billed at the Opus 4.1 rate (15/75 instead of 5/25).

Switched to longest-prefix-wins so the result no longer depends on object key order. Added the missing rows (Opus 5, 4.7, 4.6, Mythos 5) and corrected the stale ones: Opus 4.8 → 5/25, Fable 5 → 10/50, Haiku 4.5 → 1/5. Old-generation Opus 4.0/4.1 keeps 15/75.

Cache columns follow the usual derivation (write 1.25x input, read 0.1x input); noted in a comment so the columns stay consistent when a row is edited.

## Verification

- `src/__tests__/context-guard.test.ts` extended: suffix-free 1M models, case-insensitivity, dated and redundant-suffix variants, and that unknown ids still fall back to 200k. 38 tests pass.
- Pricing lookup checked against a table of ids including `claude-opus-4-8[1m]`, `claude-opus-5-fast` and `claude-opus-4-20250514`; the old-generation and unknown fallbacks still resolve correctly.
- Both changes have been running on a live Linux install since the fix was written.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -13623,21 +13623,48 @@ let tuChartState = null
 
 // Model pricing in USD per million tokens (input / output / cache-write / cache-read).
 // Fallback row is used when model is unknown or not yet captured.
+// cache-write is 1.25x input, cache-read is 0.1x input -- keep the derived
+// columns consistent with `in` when editing a row.
+// Sonnet 5 launched on introductory pricing (2 / 10) that ends 2026-08-31;
+// the standard rate (3 / 15) applies from 2026-09-01. Resolved by date at load
+// time instead of pinned to one of the two, so the table neither understates
+// spend today nor silently overstates it the morning the intro rate expires.
+const TU_SONNET5_INTRO_END = Date.parse('2026-09-01T00:00:00Z')
+const TU_SONNET5_PRICE = Date.now() < TU_SONNET5_INTRO_END
+  ? { in: 2.0, out: 10.0, cw: 2.50, cr: 0.20 }
+  : { in: 3.0, out: 15.0, cw: 3.75, cr: 0.30 }
+
 const TU_MODEL_PRICING = {
+  // INFERRED, not from the published catalogue: Opus 5 is not listed in the
+  // model reference this table was checked against. The value follows the rest
+  // of the current Opus tier (4.6/4.7/4.8 at 5 / 25); treat it as an estimate
+  // until a published rate confirms it.
+  'claude-opus-5':       { in: 5.0,   out: 25.0,  cw: 6.25,  cr: 0.50 },
+  'claude-opus-4-8':     { in: 5.0,   out: 25.0,  cw: 6.25,  cr: 0.50 },
+  'claude-opus-4-7':     { in: 5.0,   out: 25.0,  cw: 6.25,  cr: 0.50 },
+  'claude-opus-4-6':     { in: 5.0,   out: 25.0,  cw: 6.25,  cr: 0.50 },
+  // Opus 4.0 / 4.1 -- the last generation still on the old Opus pricing.
+  'claude-opus-4':       { in: 15.0,  out: 75.0,  cw: 18.75, cr: 1.50 },
+  'claude-sonnet-5':     TU_SONNET5_PRICE,
   'claude-sonnet-4-6':   { in: 3.0,   out: 15.0,  cw: 3.75,  cr: 0.30 },
   'claude-sonnet-4-5':   { in: 3.0,   out: 15.0,  cw: 3.75,  cr: 0.30 },
-  'claude-sonnet-5':     { in: 3.0,   out: 15.0,  cw: 3.75,  cr: 0.30 },
-  'claude-opus-4':       { in: 15.0,  out: 75.0,  cw: 18.75, cr: 1.50 },
-  'claude-opus-4-8':     { in: 15.0,  out: 75.0,  cw: 18.75, cr: 1.50 },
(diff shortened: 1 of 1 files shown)


Szotasz:
Thank you for this, and apologies for the conflict: the context-guard half landed a few minutes ago as #738, which reached the same conclusion independently — the `[1m]` suffix was the only signal, so every unsuffixed 1M-window model was read as 200k. Good corroboration.

The **cost table is the part that is still uniquely yours, and it matters more than the guard fix.** I checked every row against the Claude API reference and they are right where ours were wrong:

- Opus 4.8 / 4.7 / 4.6: 5 / 25 — ours said 15 / 75, so we have been overstating Opus cost by 3x
- Fable 5: 10 / 50 — ours said 3 / 15 (Sonnet's numbers)
- Haiku 4.5: 1 / 5 — ours said 0.80 / 4.0
- Sonnet 5: 3 / 15, unchanged and correct
- Derived cache columns follow the 1.25x write / 0.1x read rule throughout

Two notes, neither blocking:

1. **The `claude-opus-5` row at 5 / 25 is an assumption.** It is not in the model catalogue I can check against. The value is consistent with the rest of the Opus tier, so I am not disputing it — but please mark it with a comment saying it is inferred, so the next person does not treat it as verified.
2. **Sonnet 5 has introductory pricing of 2 / 10 through 2026-08-31.** The table uses the standard 3 / 15, which overstates current spend. Either add the intro rate with an expiry comment, or leave it and note why.

Could you rebase onto the current develop and keep only the cost-table change? The guard-map hunk will conflict and is already in. I will merge the rebased cost fix promptly — it is correcting live numbers.

mogganhun:
Rebased onto current develop and narrowed to the cost table only -- the context-guard hunks are gone, since #738 landed that half. `web/app.js` is the single file touched; mergeable is CLEAN again.

Both of your notes are addressed:

1. The `claude-opus-5` row is now explicitly marked INFERRED, with the reasoning inline (not in the published catalogue, value follows the rest of the current Opus tier). It should not be read as verified.
2. Sonnet 5 now uses the introductory rate (2 / 10), resolved by date rather than pinned:

```js
const TU_SONNET5_INTRO_END = Date.parse("2026-09-01T00:00:00Z")
const TU_SONNET5_PRICE = Date.now() < TU_SONNET5_INTRO_END
  ? { in: 2.0, out: 10.0, cw: 2.50, cr: 0.20 }
  : { in: 3.0, out: 15.0, cw: 3.75, cr: 0.30 }
```

Reasoning: pinning the intro rate silently understates spend from 2026-09-01, and pinning the standard rate overstates it today. Four lines buys correctness on both sides of the boundary and there is no dependency on anyone remembering the date. If you would rather keep the table a flat literal, say so and I will collapse it to a static row with an expiry comment -- I have no attachment to the shape.

Verified locally against the lookup: `claude-sonnet-5` resolves to 2 / 10 today, `claude-opus-4-8` to 5 / 25 (not the 15 / 75 it inherited from the `claude-opus-4` prefix before the longest-match fix), `claude-opus-4-1` still to the legacy 15 / 75, and an unknown id to the default row.

Szotasz:
Merged. This is a careful piece of work and the longest-prefix fix is the part that matters most: `'claude-opus-4-8'.startsWith('claude-opus-4')` is true, so key order silently decided the price. That class of bug survives review easily because the table *looks* correct.

One update you can act on. The comment above `claude-opus-5` says the rate is inferred because the model was not in the reference you checked. It is in the current model reference, and your inferred value is right:

- Claude Opus 5: 5 / 25 per MTok

So the `INFERRED, not from the published catalogue` caveat can come off in a follow-up. Everything else checks out against the same reference: Fable 5 and Mythos 5 at 10 / 50, Opus 4.6 / 4.7 / 4.8 at 5 / 25, Sonnet 4.6 at 3 / 15, Haiku 4.5 at 1 / 5, and the derived cache columns at 1.25x input for writes and 0.1x for reads.

The Sonnet 5 handling is the detail I would have missed. Resolving the introductory rate by date rather than pinning one of the two values means the table stops being right on a specific morning instead of quietly overstating spend from then on. The cutoff you used matches the published end of the introductory period.

Note this ships to users with next Tuesday's release, not immediately.

PR #749, opened 2026-07-27
fix(memories): serve a fresh and complete agent memory listing

YsahYarik:
`GET /api/memories` could answer with a stale or a silently incomplete list. Both failures are worst right after a restart, where a fresh session builds its idea of its own priorities from this endpoint and has no way to tell that anything is missing.

## 1. The PUT did not invalidate the cache on an ordinary edit

`updateMemory(id, content, category?, agentId?, keywords?)` gated the cache eviction on `agentId`:

```ts
if (changed && agentId) memoryCacheInvalidate(agentId)
```

But `agentId` means *"reassign this memory to that agent"*, so it is absent whenever ownership does not change — which is the normal edit. The row changed, the cache entry survived, and the listing served the pre-edit content and category for up to the 60s TTL.

The eviction now works from the row's **actual** owner, read before the write, and is gated on `changed` alone. A write touching a `shared` row clears the whole cache, because `getAgentMemories` selects `agent_id = ? OR category = 'shared'` — a shared row sits in *every* agent's list, so evicting one owner leaves the other agents serving the old copy.

The stale-cache comment above the cache block claimed *"Any write to the memories table for a given agent evicts that agent's entry"*. That was the very guarantee that was missing; it now describes what the code does.

## 2. The category filter ran after the LIMIT

`getAgentMemories` returned the N most recently accessed memories, and the route filtered those in memory:

```ts
results = getAgentMemories(agentId, limit)
if (tier) results = results.filter(m => m.category === tier)
```

So `?category=hot` answered *"the hot ones among the 50 most recently accessed memories"* instead of *"the 50 most recent hot memories"*. A hot memory that had not been touched recently dropped out of the response with no truncation signal — the reply looks complete.

This is not cosmetic, and it is the reason this PR exists rather than a one-line SQL reshuffle. A fresh session reads this list to work out what it is supposed to be doing. If a priority silently falls off the end, the session does not know that it fell off. That failure is invisible on both sides: no error, no partial-result flag, just a shorter list that looks whole.

The filter now goes into SQL ahead of the `LIMIT`, and the category is part of the cache key (`agentId:limit:category`) so two different filters cannot share one cached list.

## Tests

New `src/__tests__/memory-listing-staleness.test.ts`, 5 tests, **all 5 fail before the fix**:

- PUT with no `agent_id` in the body (the normal edit) is visible immediately
- reassignment evicts both the old and the new owner
- editing a `shared` memory is visible to a *different* agent
- a hot memory outside the most-recent N rows is still returned by `?category=hot`
- each category caches separately from the unfiltered list

## Verification

- `vitest run src/__tests__/memory-listing-staleness.test.ts` — 5/5 green (5/5 red before)
- memory-adjacent suites green: `memory-performance`, `db`, `memory-boundary`, `memory-recency-weighting`, `memories-get-agent-id-alias` (54 tests)
- `tsc --noEmit` clean

Note on the full suite: 14 tests fail in 4 files (`email-send-gate`, `federation-ui-contract`, `governance-gates`, `hook-path-guard`). Those failures are **pre-existing on `develop`** — verified by stashing this change and re-running the same four files at `b56c73d`, which reproduces exactly the same 14. They are unrelated to this PR (hook-injection wiring and a dashboard UI contract) and are tracked separately.

Also checked for the same two patterns elsewhere: no other filter-after-LIMIT in the route layer, and the federation poller's cache uses an explicit reset function with no optional-parameter hazard.


Szotasz:
Reviewed at head `5b4a0cb5`. The two fixes are right, and the reasoning in the comments is the kind I want to see. One gap blocks the merge, and it is the same failure mode this PR is about.

**`saveAgentMemory` does not evict on the `shared` create path.**

The new comment block on the cache states the contract as:

> saveAgentMemory and updateMemory evict the affected agent(s), and a write touching a 'shared' memory clears everything, because a shared row appears in EVERY agent's list

`updateMemory` now honours that. `saveAgentMemory` does not — it still ends with an unconditional:

```ts
memoryCacheInvalidate(agentId)
```

So writing a **new** `shared` memory evicts only the author's cache. Every other agent keeps serving a list that is missing that row for up to the 60s TTL, with nothing in the response to signal the omission. That is precisely the "silently incomplete listing, worst right after a restart" case in the PR description, left in place on the create path — and `shared` is the category whose whole purpose is cross-agent visibility, so it is the one where a stale reader matters most.

As written, the comment documents behaviour the code does not have, which is worse than no comment: the next reader will trust it.

Fix is the same shape as the one you already made in `updateMemory`:

```ts
if (category === 'shared') clearMemoryCache()
else memoryCacheInvalidate(agentId)
```

Please add a regression test alongside the existing three — a reader agent warms its list, the owner saves a **new** `shared` memory, the reader must see it. Your current shared test covers update, not create.

**Verified and fine, no action needed:**

- `memoryCacheInvalidate` uses `key.startsWith(`${agentId}:`)`, so the new third key segment does not escape eviction, and a `samu-2` key cannot be caught by a `samu` invalidation.
- Reading the owner and category *before* the UPDATE is the correct call — `agentId` genuinely cannot answer "whose cache went stale" when it is absent on the ordinary edit.
- Keeping the post-filter in the route for the search branches, with the comment explaining why it is a no-op for the plain listing, is the right trade rather than duplicating the ranking query.

Push the fix and I will re-review against the new head.

Szotasz:
Verified and accepted in substance. One mechanical thing to do before we merge, plus two findings that are ours, not yours.

## Your fail-before claim measured stronger than you stated

You wrote "5 tests, all 5 fail before the fix". We checked out the parent state and ran it: there are **6** tests and **all 6** fail there, all 6 green on your head. The claim holds in the direction you made it, and then some. We do not take "fails before the fix" on trust, so thanks for making it checkable.

We also confirmed independently that the shared-row flush behaves as you describe: after agent B updates a `shared` memory, agent A gets a fresh list, not just the owner.

Point 2 of your description is the one that made this a priority for us rather than a tidy-up. A category filter applied after the `LIMIT` returns a list that looks whole, and the caller has no way to know something dropped off. On the reading side of this endpoint that is exactly the failure you cannot detect from the response.

## Please rebase onto current develop

Your head predates `5bc6427`, which scoped our test suite to a sandbox. Without it we deliberately did not run the full suite against your branch on our host, because the suite would have written to live state. Rebase onto current `develop` and we will gate on the full run and merge.

## Two adjacent findings -- ours to fix, please do not widen your PR

The review turned up two things next door to your change. We are recording them on our own cards; leave them out of this PR so it stays reviewable:

1. `fleet-transfer.ts:1149` inserts into `memories` directly and never calls the cache invalidation, so imported memories can be missing for up to the TTL after a fleet import. Pre-existing, roughly a one-line fix.
2. The search paths (`q` combined with `category`) still filter after the `LIMIT`, and the listing has no truncation signal at the default limit of 20. Same failure class as your point 2, in a code path your PR does not touch.

Neither is a criticism of the scope you chose. You drew the line in the right place; we are just saying out loud where it is, so the follow-up is not mistaken for something left undone here.

PR #762, opened 2026-07-28
feat(dashboard): card decision approvals + truthful main-agent detail

VTolnai:
## Summary

Seven commits of dashboard work from our production install, rebased onto current `develop` (e626ff3). Machine-local chore commits were deliberately left out.

- **feat(approvals): card decision requests** -- an approval can now carry `options` (multiple-choice), anchor to a kanban card via `card_id`, and resolve to `decided` with the chosen key. Includes an idempotent SQLite table rebuild migration (CHECK constraint gains `'decided'`), API support in `routes/approvals.ts`, and board UI to answer directly from the card. Motivation: decisions that belong to a card should be answered and recorded on the card, not typed as prose into a chat.
- **fix(router)**: session-stuck escalation now surfaces to the main agent's inbox instead of being dropped.
- **fix(agents/voice)**: the main agent's detail route and voice-config resolve case-insensitively, and the detail view reports truthful session/model/channel data instead of sub-agent-shaped placeholders.
- **feat(dashboard)**: visible API/page diagnostics banner when the frontend and API disagree.
- **fix(dashboard)**: the main agent's dedicated opener populates voice/profile.

## Conflict resolution

`src/db.ts` had one conflict with the new upstream blocks (dashboard_users / auth_sessions / device_keys / otel_spans): resolved by keeping both, upstream blocks first, then the approvals decision-options migration.

## Test plan

- `tsc` build clean on top of e626ff3.
- Full vitest suite: 2874 passed; the 14 failures (hook-path-guard, governance-gates, email-send-gate, boot-hook-prune, federation-ui-contract) fail identically on a clean e626ff3 checkout on this machine, i.e. pre-existing/environment-dependent, none introduced by this branch.


Szotasz:
Thanks for this -- the approvals-on-cards idea matches how we actually work, and most of the PR is in mergeable shape. We verified it on the head commit (`b5890b3`), not just by reading the description: full suite 2912/2912, typecheck clean, the comms paths are intact, and we measured the two things we care about most in a migration.

**What passed, measured on disposable seeded databases:**

- Idempotency on the happy path: a second run is a clean no-op (5 rows in, 5 rows out).
- Row preservation: all rows byte-identical after the rebuild, including unicode and NULL edge cases.

**One blocker, and it is only in the migration.**

The `approvals` table rebuild (`CREATE approvals_new` / `INSERT` / `DROP` / `RENAME`) runs in a single `db.exec()` **without a transaction**. `better-sqlite3`'s `exec` is not atomic across multiple statements, so any interruption at a seam -- `SQLITE_BUSY` (realistic here, the fleet issues parallel sqlite3 calls against this database), a kill, or a full disk -- leaves the rebuild half-done. Two consequences we reproduced:

1. The `idx_approvals_card` index creation sits **after** the try/catch, unguarded. On a half-migrated database it fails with `no such column: card_id`, which exits the process during `initDatabase`. Under launchd `KeepAlive` that is a persistent crash loop, and it does not recover without manual database surgery.
2. If the interruption lands after the `DROP`, the rows stay in `approvals_new` and an empty `approvals` table is created -- silent approval-data loss.

**The fix is small, roughly five lines:**

- Wrap the rebuild in a transaction so the seams cannot be observed.
- `DROP TABLE IF EXISTS approvals_new` at the start, so a previously interrupted run self-heals.
- Move the index creation inside the same guard as the rebuild.
- Add one migration test that starts from a **seeded old-schema** database. The four test files in this PR all build fresh `:memory:` databases, so none of them exercise the path that carries the risk. That is why the issue survived a green suite.

**One scope note, not a blocker.** The diagnostics banner renders in red for every user on any non-ok API response. On a customer install a transient error would look like a broken product. Please put it behind a debug flag, default off.

Please keep the change narrow -- only the migration items above. Everything else in this PR, including the session-stuck alert in the router, verified fine and we do not want it reworked.

Once the migration is fixed we will re-verify and merge.

Szotasz:
Köszönjük a PR-t! A ma délelőtti merge-hullám után a branch konfliktusba került a develop-pal (mergeable: CONFLICTING). Kérlek rebase-eld a friss develop-ra, hogy a review mehessen tovább. Köszi!

VTolnai:
Thanks for the thorough review -- especially for reproducing the seams instead of trusting the green suite. All three migration items plus the banner note are in, and the branch is rebased onto current `develop` (67aa0cd).

**What changed (head `9a257fc`):**

- **Transactional rebuild** -- the whole `CREATE approvals_new / INSERT / DROP / RENAME` now runs inside a single `db.transaction()`, so no seam is observable; an interruption rolls back to the intact old table.
- **Self-heal** -- `DROP TABLE IF EXISTS approvals_new` leads the rebuild. With the transaction in place, a leftover `approvals_new` can only coexist with an intact `approvals`, so dropping it is always safe.
- **Index inside the guard** -- `idx_approvals_card` is created inside the transaction on the rebuild path, and inside the same try/catch on the already-migrated path, so a damaged database logs a warning instead of crash-looping `initDatabase` under `KeepAlive`.
- **Seeded old-schema migration test** (`src/__tests__/approvals-migration.test.ts`) -- starts from an on-disk database built with the pre-PR schema and seeded edge rows (unicode, NULLs, empty string, zero): asserts byte-identical row preservation, a clean idempotent second run, and self-healing from a stale `approvals_new`.
- **Diagnostics banner opt-in** -- renders only with `?mvdiag=1` (persisted per browser) or `localStorage mv_diag=1`; `?mvdiag=0` clears it. Default off; problems still go to the console.

Everything else is untouched apart from one mechanical rebase conflict in `src/web/routes/agents.ts`: upstream's new `modelResolution` fields and this PR's main-agent branch touched the same return block -- resolved by keeping your detailed fields and routing the main agent through `readMainAgentModel()` (source reported as `explicit_model`), since `resolveAgentModelDetailed` reads `agent-config.json`, which the main agent does not have.

**Verified on the new head:** `tsc` clean, full vitest suite 229 files, 3147 passed / 1 skipped, zero failures (the 14 environment-dependent failures we reported earlier no longer reproduce on current develop).

PR #765, opened 2026-07-29
fix(worker): derive worker home from MAIN_AGENT_ID + WEB_ONLY-gate every session start

Szotasz:
## What (card WORKERHOME1)
The 2026-07-28 same-host sandbox incident: a WEB_ONLY sandbox boot created `<MAIN_AGENT_ID>-worker(-fast)` tmux sessions with live claude AND wrote credentials into the LIVE worker's config dir. Two independent holes, both fixed here (either alone is insufficient, per the card):

1. **Worker home was hardcoded** (`~/.marveen-worker`): a sandbox with its own `MAIN_AGENT_ID` still resolved to the live install's worker config dir. Now `workerHomeFor(agentId, variant)` derives it from `MAIN_AGENT_ID` exactly like the session name (#611). **Back-compat:** the default id `marveen` yields the byte-identical historical path, so every existing default install sees zero migration and an unchanged Keychain path-hash (`configDirKeychainService`); a renamed/sandbox id gets its own isolated dir. `MARVEEN_WORKER_DIR*` env overrides keep precedence.
2. **WEB_ONLY did not suppress the lazy start**: only the boot-time pre-start in `web.ts` was gated; `runViaWorker -> ensureWorkerReady -> startWorkerSessionFor` was not. New pure gate `workerStartAllowed(env)` now refuses at the three session-lifecycle choke points BEFORE any side effect:
   - `startWorkerSessionFor` (covers the `ensureWorkerCwd` config-dir writes + `tmux new-session`),
   - `ensureWorkerReady` (fails fast; otherwise a 90s poll would end in `alertWorkerStuck` paging the LIVE channel from a staging instance),
   - `restartWorkerSession` (defense-in-depth: it `kill-session`s a live worker).

## Tests
- Pure vectors: default id keeps `.marveen-worker`/`.marveen-worker-fast`; `jarvis` derives isolated dirs; no collision across ids; slow/fast never share a home.
- Gate vectors: `WEB_ONLY=true` blocks; unset/`false`/empty allow.
- **String-contract wiring guard**: fails if any of the three choke points loses its `workerStartAllowed()` check. Adversarially verified: removing the `restartWorkerSession` gate makes exactly this test fail.

## Gates
- `tsc --noEmit` clean; suite **2889/2890** (the 1 failure is the pre-existing `federation-ui-contract` develop breakage, fixed separately in #764, unrelated).
- Em-dash clean on added lines; no hardcoded owner/chat-id/user path (the derivation removes one).

## Out of scope / follow-up
Cold-start probe from the current disk state (expired access token + valid refresh token: does the worker come up after a restart?) runs on the VPS per the standing rule and will be reported on the card, not in this PR.

### DIFF
--- a/src/__tests__/agent-worker.test.ts
+++ b/src/__tests__/agent-worker.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from 'vitest'
-import { buildWorkerPrompt, decidePoll, configDirKeychainService } from '../web/agent-worker.js'
+import { readFileSync } from 'node:fs'
+import { join, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { buildWorkerPrompt, decidePoll, configDirKeychainService, workerHomeFor, workerStartAllowed } from '../web/agent-worker.js'
 
 // Pure-logic tests for the interactive-tmux worker that backs runAgent on the
 // subscription (jun.15 migration). The live-session orchestration is exercised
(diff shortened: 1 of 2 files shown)


Szotasz:
**Cold-start probe results (vps47, per the card's standing VPS-only rule):**

1. **Reproduced the silent worker death on the unfixed code**: dashboard log showed `launched interactive worker session` + a successful version probe at boot, yet `tmux ls` had zero worker sessions seconds later (gone before the first 5s poll). Captured the actual failure: `bash: line 1: claude: command not found`, exit 127 -- login-shell PATH on the stock Debian root lacks `~/.local/bin` (native installer target). The in-process version probe uses `resolveFromPath` so it succeeds and masks the death. This is the second commit's fix.
2. **Live-verified the fixed launch line on vps47**: same env (fleet setup-token from disk, isolated `CLAUDE_CONFIG_DIR`), claude invoked by resolved absolute path -> Claude Code v2.1.220 boots to an idle prompt and stays up (60s+ poll). Probe session + stderr file removed afterwards.
3. **Host-side finding (read-only, no restart, reported on the card)**: the macOS host's default Keychain entry currently holds an access token expired ~37h ago with a valid refresh token -- which contradicts the comment premise here that "the host keeps its Claude Code-credentials token fresh". The macOS refresh-from-file cold-start path is NOT measurable on a Linux VPS (no Keychain); flagged as an open hypothesis on the card, not asserted.

PR #767, opened 2026-07-29
fix(fleet): Telegram progress mirror + four fixes from running a multi-agent fleet

jkarcsi:
Rebased onto current `develop` and split as you asked. The three governance
gates now live in #770 -- this PR no longer touches `self-pace-gate.mjs`,
`readonly-repo-gate.mjs` or `host-guard.py`.

Three commits fell out during the rebase, all because the work is already
upstream: the token-usage worker-visibility fix, the `/api/models/available`
sync, and the configurable default agent model (`develop` now carries an
identical `DEFAULT_AGENT_MODEL` in `src/config.ts`; keeping mine would have been
a duplicate declaration, and the build says so).

What is left is one coherent group: what the owner can see of the fleet from his
phone, plus two fixes around it.

**Telegram progress**
- `feat(channels)`: the owner had no way to tell a thinking agent from a dead
  one -- the only feedback was the final answer, and a session that died
  mid-turn said nothing at all. A small daemon mirrors the Claude Code status
  line into ONE Telegram message, edited live and deleted when the turn ends,
  with per-agent `silent`/`indicator`/`verbose` modes. If the session dies
  mid-turn the indicator is turned INTO the error rather than deleted, so
  silence never masquerades as progress.
- `fix(progress)`: when a reply went out mid-turn the indicator stayed ABOVE it,
  so the owner read the answer and then an "I'm thinking" line under it. It is
  now deleted and recreated at the bottom.
- `feat(progress)`: the indicator only existed while a turn was in flight, so a
  background shell, the monitor or a running sub-agent went invisible the moment
  the answer went out -- a ten-minute build and an idle machine looked identical
  from the phone. The daemon now reads the background counters out of the idle
  footer and keeps a second, visually distinct indicator alive. Anchored so a
  scrollback line quoting the footer cannot park a permanent message in the
  chat. New pure seam `classify_pane()` with sixteen contract cases in
  `scripts/__tests__/telegram-live-progress.test.py`, verified red on a copy
  with the background branch removed.

**Fleet fixes**
- `fix(rembrandt)`: the review prompt never reached Codex. A quoted `$CMD`
  closed the outer quote, the prompt was word-split into `bash -lc` positional
  parameters, and the reviewer started with an empty prompt and produced a
  plausible-looking answer anyway. The prompt now goes in on stdin.
- `fix(profiles)`: read-only workers hung on permission prompts instead of
  working -- an unattended worker cannot answer a prompt. `planner` and
[...]

PR #770, opened 2026-07-29
fix(governance): the three agent gates, split out of #767

jkarcsi:
The 2026-08-14 split, done, plus the yarn finding from the 2026-09-03 verify.
Rebased onto current `develop`.

**What is in the PR now: two gates and their tests.**

| file | why |
|---|---|
| `scripts/readonly-repo-gate.mjs` | the gate |
| `scripts/self-pace-gate.mjs` | two false-positive fixes |
| `src/__tests__/governance-gates.test.ts` | cases for both |
| `src/web/agent-scaffold.ts`, `src/web/profiles.ts` | the gate's wiring (see below) |

**What is gone:** `templates/profiles/planner.json`, `templates/profiles/qa-runner.json` and `scripts/host-guard.py`. The profiles wait on your product decision, and I am not opening a PR for them until you say the decision went that way. `host-guard.py` I am not carrying further: you were right that in the shipped form it described my install, and generalising it is not worth a review round on your side. Both are recoverable from `6847314` and `11de187` if you ever want them.

**Why the wiring stayed, and why it changed shape.** The gate has to be injected by something, and that wiring travelled with the profiles when I moved them out of #767. As written it keyed on the ids `planner` and `qa-runner`, so on this repo it was a hook that could never fire, named after two profiles this repo does not have -- the same defect you named on `host-guard.py`. `ProfileTemplate` now carries an opt-in `readonlyRepo` flag instead, and no profile is named in the scaffold. No shipped profile sets it yet, so the gate is inert here until a role opts in; if you would rather the wiring wait for the profiles too, say so and I will strip it to the bare script.

**The yarn finding.** Fixed as prescribed: `yarn\s+(install)?\b` -> `yarn\s+install\b`. The optional group plus the word boundary matched `yarn ` + anything, so `yarn add`, `yarn exec`, `yarn dlx` and `yarn run` all inherited an exemption meant for dependency installs. Bare `yarn` (Yarn 1's implicit install) loses the exemption with it, which is right: it carries no redirect and no mutating verb, so it now passes on its own merits instead of by blanket skip.

**What the gates do**

`readonly-repo-gate.mjs` -- writes to repository *source* are denied; build and dependency paths (`node_modules`, `.next`, `dist`, `coverage`, `__pycache__`, `.venv`, `target` ...), package-manager installs and branch movement (`git checkout/switch`, `git worktree add`) are allowed. Still denied inside the roots: `Write`/`Edit`/`NotebookEdit`, shell redirects, `sed -i`, `rm`/`mv`/`cp`/`chmod`/`chown`, `tee`, `mkdir`, `touch`, and `git commit/push/reset/restore/clean/rm/mv/apply/stash` plus `git checkout -- <path>`. Scope is `READONLY_REPO_ROOTS`, default `<home>/projects` derived at runtime.

Why a hook and not permissions: under `strict` the denies hold but Claude Code prompts on the `cd X && git ...` compound, and an unattended worker cannot answer a prompt, it hangs (seen twice on real workers). Under `permissive` nothing prompts, but the launcher passes `--dangerously-skip-permissions`, which bypasses allow/deny. Hooks run in both modes.

Why the install allowance is not a loosening: with `npm ci` blocked the QA worker could not install, could not run the suite, and still produced a confident diagnosis -- of a problem that did not exist, because the block had pushed it onto `npx prisma`, which fetched the registry's latest instead of the pinned version.

`self-pace-gate.mjs` -- two false-positive fixes, both narrowing what counts as a self-scheduling attempt; neither widens what an agent may schedule. A heredoc body is blanked before matching (quoted, unquoted and `<<-`), because it is data and never reaches a shell as commands, so a commit message or a PR description beginning "at the same time ..." is no longer read as the `at` scheduler. The real vectors stay blocked, including a command that hides a heredoc and then chains `tmux send-keys`.

**Verification**

- `npx tsc --noEmit` clean; full suite on this branch: **352 files, 4610 passed, 1 skipped, 0 failed**.
- The yarn regression is pinned three ways: every package manager's real install still exempt, four non-install yarn subcommands now judged, and `yarn install && yarn add x > <repo>/file` denied on the second segment.
- The wiring is covered for opt-in/opt-out, idempotency across a respawn, and survival of unrelated `PreToolUse` entries.
- The injected command now goes through `hookCommand()`, matching the quoted-absolute-interpreter rule that landed on `develop` after this branch was opened.

#767 no longer depends on this PR: the profiles that needed the gate are not in either PR any more.

### DIFF
--- a/scripts/readonly-repo-gate.mjs
+++ b/scripts/readonly-repo-gate.mjs
@@ -48,7 +48,14 @@ function isArtifactPath(p) {
 
 // Package managers only ever write artifacts (node_modules, lockfile refresh),
 // so they are allowed wholesale rather than path-checked.
-const INSTALL_RX = /\b(npm\s+(ci|install|i)\b|pnpm\s+(install|i)\b|yarn\s+(install)?\b|pip\s+install\b|poetry\s+install\b|bundle\s+install\b)/
+// Every branch REQUIRES its subcommand. The yarn branch used to read
+// `yarn\s+(install)?\b`, and the optional group made the pattern match bare
+// `yarn ` + anything -- `yarn add`, `yarn exec`, `yarn dlx` all skipped the
+// gate's inspection, which is the opposite of what an install-only exemption
+// is for. (Found in review on #770, 2026-09-03.) Bare `yarn` (Yarn 1's
+// implicit install) is no longer exempted either: it carries no redirect and
+// no mutating verb, so it passes on its own merits rather than by blanket skip.
+const INSTALL_RX = /\b(npm\s+(ci|install|i)\b|pnpm\s+(install|i)\b|yarn\s+install\b|pip\s+install\b|poetry\s+install\b|bundle\s+install\b)/
 
 // Branch movement is not a source edit -- a QA worker needs it for baseline
 // comparison. `git checkout -- <path>` / `git restore <path>` IS a working-tree
--- a/scripts/readonly-repo-gate.mjs
+++ b/scripts/readonly-repo-gate.mjs
@@ -1,0 +1,169 @@
+#!/usr/bin/env node
+// PreToolUse hard-gate: a worker may READ the project repos but never WRITE them.
+//
+// Why a hook and not a permissions deny-list. The two permission modes each
+// fail one half of the requirement:
+//
+//   - strict:     path denies are enforced, but Claude Code prompts on things
+//                 an allow-list cannot pre-approve (notably the `cd X && git …`
+//                 compound, which trips the cd-guard). An UNATTENDED worker
+//                 cannot answer a prompt, so it simply hangs -- observed twice
+//                 on davinci-ocura and vermeer-ocura.
+//   - permissive: nothing prompts, but the launcher passes
+//                 --dangerously-skip-permissions, which BYPASSES allow/deny.
+//                 A `Write(/home/ubuntu/projects/**)` deny is then decorative.
+//
+// Hooks run regardless of permission mode. So the worker runs permissive (never
+// hangs) and this gate provides the actual enforcement.
+//
+// Scope comes from READONLY_REPO_ROOTS (colon-separated). The default is
+// <home>/projects, derived at runtime -- a shipped script must not carry one
+// install's absolute home path.
+
+import { readFileSync, realpathSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { homedir } from 'node:os'
+import { resolve, join } from 'node:path'
+
+const ROOTS = (process.env.READONLY_REPO_ROOTS || join(homedir(), 'projects'))
+  .split(':').map(s => s.trim()).filter(Boolean)
+
+const WRITE_TOOLS = new Set(['Write', 'Edit', 'NotebookEdit'])
+
+// The rule is "do not change the SOURCE", not "do not touch the filesystem".
+// Dependency and build output live inside the repo but are not source: a QA
+// worker that cannot run `npm ci` in a fresh worktree is not a QA worker.
+// (Learned the hard way: blocking installs pushed a worker onto `npx prisma`,
+// which fetched the registry's latest instead of the pinned version and made it
+// diagnose a config problem that did not exist.)
+const ARTIFACT_SEGMENTS = new Set([
+  'node_modules', '.next', 'dist', 'build', 'out', 'coverage', '.turbo',
+  '.cache', '.venv', '__pycache__', 'target', '.pytest_cache', '.nuxt',
+  '.svelte-kit', '.output', 'tmp', '.parcel-cache',
+])
+
+function isArtifactPath(p) {
+  return String(p || '').split('/').some(seg => ARTIFACT_SEGMENTS.has(seg))
+}
+
+// Package managers only ever write artifacts (node_modules, lockfile refresh),
+// so they are allowed wholesale rather than path-checked.
+const INSTALL_RX = /\b(npm\s+(ci|install|i)\b|pnpm\s+(install|i)\b|yarn\s+(install)?\b|pip\s+install\b|poetry\s+install\b|bundle\s+install\b)/
+
+// Branch movement is not a source edit -- a QA worker needs it for baseline
+// comparison. `git checkout -- <path>` / `git restore <path>` IS a working-tree
+// mutation, so those stay blocked.
+const GIT_SAFE_RX = /\bgit\s+(checkout|switch)\s+(?!.*(--\s|--\s*$))[^\s-][^\s]*\s*$|\bgit\s+(switch|checkout)\s+-b\b|\bgit\s+worktree\s+(add|list|prune)\b/
+
+function underRoot(p) {
+  if (!p) return false
+  let abs
+  try { abs = resolve(String(p)) } catch { return false }
+  return ROOTS.some(r => abs === r || abs.startsWith(r.endsWith('/') ? r : r + '/'))
+}
+
+// Bash is the wide-open route: a redirect or an in-place edit reaches the repo
+// without ever touching the Write tool. Checked per segment so a read command
+// in one half of a compound is not judged by the other half.
+function splitSegments(cmd) {
+  return String(cmd || '').split(/&&|\|\||;|\n|\|/).map(s => s.trim()).filter(Boolean)
+}
+
+const MUTATING_RX = [
+  /\bsed\s+[^|]*-i\b/,                       // in-place edit
+  /\b(rm|mv|cp|install|truncate|chmod|chown)\b/,
+  /\btee\b/,
+  /\bgit\s+(commit|push|reset|restore|clean|rm|mv|apply|stash)\b/,
+  /\bgit\s+checkout\s+.*--\s/,
+  /\bnpm\s+publish\b/,
+  /\b(pnpm|yarn)\s+(add|remove)\b/,
+  /\bmkdir\b/,
+  /\btouch\b/,
+]
+
+// `> file` / `>> file` -- capture the target so we only object when it lands in
+// a protected root. `2>/dev/null` and friends are not file writes we care about.
+const REDIRECT_RX = /(?<!\d)>>?\s*("[^"]+"|'[^']+'|[^\s;|&]+)/g
+
+function bashViolation(cmd) {
+  // The cwd carries ACROSS segments: `cd /repo && echo x > src/a.ts` puts the
+  // redirect in a later segment than the cd, so judging segments in isolation
+  // would wave it through. Track it.
+  let cwd = null
+  for (const seg of splitSegments(cmd)) {
+    const cd = seg.match(/\bcd\s+("[^"]+"|'[^']+'|[^\s;|&]+)/)
+    if (cd) cwd = cd[1].replace(/^["']|["']$/g, '')
+
+    const inRoot = (p) => underRoot(p) || (cwd && !String(p).startsWith('/') && underRoot(cwd))
+
+    if (INSTALL_RX.test(seg) || GIT_SAFE_RX.test(seg)) continue
+
+    for (const m of seg.matchAll(REDIRECT_RX)) {
+      const target = m[1].replace(/^["']|["']$/g, '')
+      if (target === '/dev/null' || isArtifactPath(target)) continue
+      if (inRoot(target)) {
+        const shown = target.startsWith('/') ? target : `${cwd}/${target}`
+        return `átirányítás a repóba: ${shown}`
+      }
+    }
+
+    for (const rx of MUTATING_RX) {
+      if (!rx.test(seg)) continue
+      const touchesRoot = ROOTS.some(r => seg.includes(r)) || (cwd && underRoot(cwd))
+      if (touchesRoot && !isArtifactPath(seg)) return `módosító parancs a repóban: ${seg.slice(0, 120)}`
+    }
+  }
+  return null
+}
+
+const GATE_MSG = (detail) =>
+  `Írás a projekt repóba TILTOTT (readonly-repo-gate). ${detail}. ` +
+  'A szereped olvasó: elemzel, tesztelsz, jelentesz -- a kódot nem te módosítod. ' +
+  'Ha a feladathoz tényleg írni kellene, azt az orchestratorodnak jelezd, és ő ' +
+  'vagy átadja a végrehajtónak, vagy bemásolja amit készítettél. Írni szabadon ' +
+  'tudsz a saját agent-mappádba és a /tmp alá -- oda dolgozz.'
+
+function allow() { process.exit(0) }
+
+function deny(reason) {
+  process.stdout.write(JSON.stringify({
+    hookSpecificOutput: {
+      hookEventName: 'PreToolUse',
+      permissionDecision: 'deny',
+      permissionDecisionReason: reason,
+    },
+  }))
+  process.exit(0)
+}
+
+function isInvokedDirectly() {
+  try {
+    const self = realpathSync(fileURLToPath(import.meta.url))
+    const entry = process.argv[1] ? realpathSync(process.argv[1]) : ''
+    return self === entry
+  } catch {
+    return false
+  }
+}
+
+if (isInvokedDirectly()) {
+  let payload
+  try { payload = JSON.parse(readFileSync(0, 'utf-8')) } catch { allow() }
+
+  const tool = payload?.tool_name
+  const input = payload?.tool_input || {}
+
+  if (WRITE_TOOLS.has(tool)) {
+    const target = input.file_path || input.notebook_path || input.path
+    if (underRoot(target) && !isArtifactPath(target)) deny(GATE_MSG(`${tool} -> ${target}`))
+  }
+
+  if (tool === 'Bash') {
+    const v = bashViolation(input.command)
+    if (v) deny(GATE_MSG(v))
+  }
+
+  allow()
+}
+
+export { bashViolation, underRoot, isArtifactPath }
--- a/scripts/self-pace-gate.mjs
+++ b/scripts/self-pace-gate.mjs
@@ -301,7 +301,14 @@ export function maskInertLiterals(command) {
 const PROSE_FLAGS = String.raw`--body|--message|--notes|--title|--subject|-b|-m|-t`
 
 export function stripProseArguments(seg) {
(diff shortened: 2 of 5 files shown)


Szotasz:
Thank you for splitting this out exactly as asked: three files, nothing else, rebased. The two new gates (`host-guard.py`, `readonly-repo-gate.mjs`) read clean and are not wired anywhere yet, so they carry no posture change on their own.

One change-request before this can go in, and it is in the third file.

**`stripHeredocBodies` blanks unquoted heredocs too, and that reopens a real hole.**

The regex makes the quote optional:

```js
/(<<-?\s*(['"]?)([A-Za-z_][A-Za-z0-9_]*)\2)([\s\S]*?)(^\s*\3\s*$)/gm
```

`<<'EOF'` is literal, so blanking its body is safe. `<<EOF` is **not**. The shell expands `$(...)` and backticks inside that body at exec time. Blanking it hides a live command substitution from `SCHEDULER_RX`, so the gate stops seeing something that will actually run.

This is the same rule your other strippers already apply, and apply correctly. `stripDataPayloads`, `stripGitCommitMessages` and the new `stripProseArguments` all deliberately leave a double-quoted value alone when it contains `$(` or a backtick, and `stripProseArguments` even says so in its own comment: *"a double-quoted value with `$(` or ` may substitute, so it is left alone."* An unquoted heredoc is the same category; it just wasn't given the same guard.

Concretely, this stays denied today and would slip through with this change:

```
git commit -m "$(cat <<EOF
fix
$(at now)
EOF
)"
```

Independently confirmed by node probe against the combined tree, and #720's stricter suite has an assertion for exactly this case (`STILL denies a real command substitution hidden in a commit message`). This PR's own tests don't cover it, which is why it passes here.

**Two things to fix:**

1. Blank the body only for a **quoted** marker (`<<'EOF'` / `<<"EOF"`), or for unquoted markers preserve the substitution the way the other strippers do.
2. Add the unquoted-heredoc assertion to this PR's tests, so the guard can't be lost again later.

**Sequencing, so you don't lose work:** #720 is now merged to `develop`. The two changes are complementary rather than conflicting. #720 narrows at the regex level, yours strips at the preprocessing level, and together they are genuine defense-in-depth for the same false-positive family. Please **rebase onto current `develop`**; they auto-merge textually, but on the combined tree the #720 assertion above fails until the heredoc fix is in. That failure is the point: it is what caught this.

One separate note, not a blocker and not something to change here: the `INSTALL_RX` allowlist in `readonly-repo-gate.mjs` (`npm ci`, `pnpm`, `pip`, ...) is a reasonable QA need, because a read-only worker cannot build or test without it. But package-manager installs run postinstall scripts, which is interpreter reach, so it is a posture question rather than a pure hardening one. That decision belongs to the repository owner, not to this review; I am taking it to him separately. It does not hold up the gate work itself.

jkarcsi:
You were right, and it is worse than a false-negative in theory: I verified the bypass runs.

**Fixed.** `stripHeredocBodies` now blanks a body only when the marker is quoted, or when an unquoted body has nothing to expand. An unquoted body containing `$(` or a backtick is left alone, which is exactly the rule `stripDataPayloads`, `stripGitCommitMessages` and `stripProseArguments` already apply to double-quoted values. The guard was simply missing here.

**The assertion is in this PR now**, as you asked, and not only at gate level: a dedicated `stripHeredocBodies` block with seven cases -- quoted, double-quoted, unquoted-without-substitution, unquoted with `$(`, unquoted with a backtick, a *quoted* body that merely mentions `$(` as text (still blanked, because the shell will not expand it), and both `<<-` variants.

**Verified red without the guard**, which is the part that makes it reviewable: reverting just the two guard lines fails four tests, and one of them is #720's own `STILL denies a real command substitution hidden in a commit message`. That is your point reproducing in this tree.

**Rebased onto current develop.** Full suite on the combined tree: 3029 passed, 1 skipped, 0 failed.

On your separate note about `INSTALL_RX`: agreed that it is a posture question, not a hardening one, and it belongs to the owner rather than to this review. For the record of what the allowance actually bought: with installs blocked, a QA worker could not run `npm ci`, so it could not install, could not run the suite, and still produced a confident diagnosis -- of a configuration problem that did not exist, because the block had pushed it onto `npx prisma`, which fetched the registry's latest instead of the pinned version. I would rather the owner weigh that trade with the postinstall reach stated plainly than have me argue it inside a gate PR.

Szotasz:
Re-measured on the new head, merged with the current base. The false positive this fixes is real, and I can show it from my own work. But the change also lets through four shapes the gate blocks today, so this is a change request rather than an approval.

**The benefit is real, measured on a case that actually happened.** While writing a probe script for this review, the live gate denied my own command: a heredoc carrying a script whose text contained scheduler words as data. I fed that exact command string into both gate versions. Today's gate denies it. This branch allows it. So the class of false positive you are describing is not hypothetical, and the heredoc and prose-argument strippers do address it.

**What blocks it.** Four shapes that today's gate denies pass on this branch:

```
bash <<'EOF'      / <scheduler command> / EOF
sh   <<"EOF"      / <scheduler command> / EOF
ssh box <<'EOF'   / <scheduler command> / EOF
bash -s <<'EOF'   / <scheduler command> / EOF
```

The guard in `stripHeredocBodies` asks whether the **shell** can expand the body, and keeps unquoted bodies containing `$(` or a backtick for exactly that reason. That part is right. The case it misses is a different question: whether the body is **fed to an interpreter that runs it**. A quoted heredoc is literal to the shell, so the guard blanks it, but `bash`, `sh` and `ssh` execute the body all the same. The two questions are not the same, and only the first is currently asked.

Suggested shape of the fix, one condition rather than a redesign: blank the body only when the command word owning the redirect is not an interpreter or remote executor (`bash`, `sh`, `zsh`, `python*`, `node`, `ssh`, `docker exec`, and similar). Otherwise leave the body in the string so the pattern still sees it.

**One scoping nit on this PR's own change.** `stripProseArguments` applies the `--body|--message|--notes|--title|--subject|-b|-m|-t` blanking to any binary. `stripGitCommitMessages` right above it is deliberately scoped to `git commit|tag|stash`, for the same reason: a short flag means different things to different tools. Scoping the prose stripper to the tools it was written for (`gh`, `git`, `glab`) keeps the false-positive fix without widening what gets blanked.

**On the conflict with the other open PR touching this file.** Both this branch and the other one modify `scripts/self-pace-gate.mjs`, and merging either makes the other conflict. Since this one needs another round anyway, the other should go first and this one rebase onto it, so the larger change does not pay for the conflict.

Happy to re-run the same probe set against the updated branch when you push.

jkarcsi:
Both addressed on the new head (80dda13).

**1. Executable heredoc.** `stripHeredocBodies` now keeps the body visible when the command word owning the redirect is an interpreter or remote/container executor — `bash`, `sh`, `zsh`, `dash`, `ksh`, `ash`, `python*`, `node`, `ruby`, `perl`, `php`, `ssh`, and `docker/podman/kubectl exec` — seeing through `sudo`/`env`/`nohup`/`exec`/`command` prefixes and a `/usr/bin/...` path. A non-interpreter owner (`git commit -F -`, `tee`) still blanks. Your four shapes (`bash` / `sh` / `ssh box` / `bash -s` <<'EOF') now stay caught; added a test for each plus a sudo-prefix case and the git/tee negatives.

**2. Prose stripper.** Scoped to `gh`/`git`/`glab`, mirroring `stripGitCommitMessages`' guard. `tar -t` / `cut -b` are left alone; `gh --body` still blanks. Test added.

On ordering: agreed — the other open PR that also touches `self-pace-gate.mjs` goes first, and this rebases onto it so the larger change does not pay for the conflict. Happy for you to re-run the probe set against 80dda13.

jkarcsi:
On the ordering: the other open PR that also touched `scripts/self-pace-gate.mjs` has since merged, and #770 is now **mergeable-clean** against `develop` with no conflict, so the rebase-first caveat is moot — nothing left to sequence behind.

The two review fixes are pushed on 80dda13 (interpreter/remote-executor-owned heredoc bodies kept visible; prose stripper scoped to gh/git/glab), 7 new tests green. Ready whenever you want to re-run the probe set — I'd rather land it on your green light than merge ahead of it, since it's the gate itself.

jkarcsi:
A #767 review-ja nyomán ide került át a `planner` és a `qa-runner` profil, a hozzájuk tartozó scaffold-bekötéssel együtt (`6847314`).

**Miért ide:** mindkét profil azzal indokolja a `permissionMode: permissive`-ot, hogy az írást a `readonly-repo-gate` PreToolUse hook tiltja. Az a hook ebben a PR-ben van. Ha a #767 ment volna előbb, két megengedő profilt szállítottunk volna, amiknek az egyetlen írás-védelme egy nem létező fájl. A bekötés (`agentGetsReadonlyRepoGate` + `injectReadonlyRepoGate`) ugyanezért jött velük: profil, bekötés és kapu egy döntés.

**A jogosultsági lista szűkítve** (a tulajdonos ezt az utat választotta): az értelmezők (`python3`, `node`, általános `npx`) és a `curl` kikerültek, helyettük konkrét teszt- és build-parancsok maradtak; az inter-agent jelentés a `scripts/agent-msg.sh`-n megy. Mindkét leírás kimondja, hogy a lista munkakör-szűkítés, nem titok-határ, mert a `Read(...)` deny a Read toolt vezérli, nem azt, amit egy elindított `cat` beolvas.

**Egy meglévő teszt-bukásra hívom fel a figyelmet, ami nem ebből a változásból jön.** A `governance-gates.test.ts` három scaffold-wiring esete bukik ezen az ágon: `injectSelfPaceGate is idempotent`, `the hook MATCHER fires on native file tools too`, `self-pace gate survives a respawn re-run`. Bizonyítottam, hogy előttem is bukott: a változásaimat `git stash`-sel félretéve ugyanaz a három bukik, 79 megy át mindkét futásban. Nálam worktree-ből futott, ami ezeknél a teszteknél korábban is okozott környezeti bukást, tehát elképzelhető, hogy a CI-ban zöld. A typecheck tiszta.

Szotasz:
Szia! Megmértük a friss fejet (68473149), és kérnék egy szétvágást. Előbb a jó hír, mert az a lényeg: a self-pace-gate javításodra a review-nk zöld volt, csak közben a PR körbenőtte magát.

**Mi változott a mi oldalunkról.** A korábbi GO-m a 80dda13f fejre szólt, és nálunk szabály, hogy a jóváhagyás a pinnelt fejhez tartozik. A mai rebase után a PR hét fájl és +759 sor, tehát a régi verdikt nem hordozható át rá. Ez nem a te hibád, csak jelzem, miért nem tudjuk így bemergelni.

**Egy dolog viszont időközben tárgytalanná vált.** A #720-at július 29-én bemergeltük, tehát az a sorrend-kérdés, hogy a kettő közül melyik menjen előre, magától eldőlt. Ezt nálunk egy kártya őrizte, és az én mulasztásom, hogy nem szóltam róla.

**A kérésem: bontsd három részre.**

1. A self-pace-gate és a readonly-repo-gate javítások maradjanak egy kis, önálló PR-ben. Ezt gyorsan újra tudjuk kapuzni, a tartalmát már egyszer átnéztük.

2. A két profil (planner.json, qa-runner.json) külön PR-be. Nem a minőségük miatt: nálunk most nyitott termék-döntés, hogy akarunk-e egyáltalán ágenseket biztonsági profil alá tenni, és amíg ez nem dől el, a #681 mellett áll sorban ez is. Amint döntés születik, szólok.

   Külön kiemelem, mert ritkán látjuk: a planner profil leírásában pontosan végiggondoltad, hogy a jogosultsági lista munkakör-szűkítés és nem titok-határ, és emiatt hagytad ki a curl-t. Ez az a csapda, amibe nálunk korábban belefutottunk, tehát ez a rész jó úgy, ahogy van.

3. A host-guard.py a jelenlegi formájában a te környezetedet írja le: a szolgáltatásnevek és a /home/ubuntu útvonalak be vannak égetve. Nálunk ez halott vagy félrevezető kód lenne. Ha konfigurálhatóvá teszed, szívesen megnézzük, csak jöjjön külön.

Semmit nem zárunk le, a munkád nem vész kárba. Ha a szétvágás sok, szólj, és megbeszéljük, melyik darabbal érdemes kezdeni.

Szotasz:

Szotasz:
Szia, es elnezest a harmincot napos csendert. A verify lefutott, es ket dolgot kell visszaadnom -- az egyik a te javadra irando lelet.

## 1. A szet-szedesi keres meg all

A karbantartoi review 2026-08-14-en harom felé bontast kert (a profilok ki, a `host-guard.py` ki es konfiguralhatova). A mai fejen ez KEZELETLEN: a PR torzse tovabbra is 3 fajlt allit, a valosag 7. A korabbi maintainer-GO a `80dda13f` fejre szolt, nem a mostanira.

Emellett a `host-guard.py` a szallitott formaban halott kod, es a szerzo sajat flotta-neveit tartalmazza -- egy kulso telepitesen ez nem mukodik es nem is ertelmezheto.

## 2. UJ BIZTONSAGI LELET, ES EZT KOSZONJUK, MERT A TIED HOZTA ELO

Az `INSTALL_RX` yarn-aga:

```js
yarn\s+(install)?\b
```

Az `(install)?` OPCIONALIS, es utana `\b` all. Vagyis a minta illeszkedik a puszta `yarn ` + BARMI alakra: `yarn add`, `yarn exec`, `yarn dlx` mind atmegy a readonly-kapun. Probaval igazoltuk.

**A javitas egyetlen karakter:** a `?` elhagyasa, tehat `yarn\s+install\b`. Igy a minta azt engedi at, amit szant: a fuggoseg-telepitest, es semmi mast.

Ez nem a te hibad abban az ertelemben, hogy a tobbi ag (npm, pnpm, pip, poetry, bundle) mind helyesen kotelezove teszi az alparancsot -- egyedul a yarn-ag csuszott el. De pont ezert erdemes: egy kapu, ami ot agbol negyen jol mukodik, a hatodikon at is engedi ugyanazt.

## Ami jo, es nem veszik el

A kompozicio a mai scaffold-valtozasokkal ep, es a teljes keszlet 4560 zold a merge-eredmenyen.

## A GO-ut

1. a `?` elhagyasa az `INSTALL_RX` yarn-agabol,
2. a 08-14-i szet-szedes elvegzese (vagy egy indoklas, hogy miert nem bonthato),
3. a `host-guard.py` vagy elo kodda teve es konfiguralhatova, vagy kivive a PR-bol.

Ha ezek megvannak, soron kivul ranezek.

---

jkarcsi:
Mind a harom pont kesz, a fej `6a11e5d`, a branch a mai `develop`-on (`5c9a925`). A PR torzset is atirtam, mert az meg a regi harom fajlt allitotta.

## 1. A szet-szedes

Kikerult a `templates/profiles/planner.json`, a `templates/profiles/qa-runner.json` es a `scripts/host-guard.py`. A PR most ket gate es a tesztjeik.

A ket profilra nem nyitok kovető PR-t, amig nem szolsz, hogy a termek-dontes arra ment: felesleges lenne egy PR, ami a ti oldalatokon var. A `host-guard.py`-t nem viszem tovabb. Igazad volt, hogy a szallitott formaban az en telepitesemet irja le, es a konfiguralhatova tetel nem er meg nektek egy review-kort. Mindketto visszahozhato a `6847314` es a `11de187` commitbol, ha megis kelleni fog.

## 2. A gate bekotese, es amiben tagabban jartam el a review betujenel

A gate-et valaminek be kell kotnie, es ez a bekotes a profilokkal egyutt jott at a #767-bol. A regi alakban a `planner` es a `qa-runner` id-re illesztett, vagyis ezen a repon egy olyan hook volt, ami soha nem sulhet el, ket olyan profil nevere hivatkozva, ami itt nincs. Ez pontosan az a hiba, amit a `host-guard.py`-n kimondtal.

Ezert a `ProfileTemplate` kapott egy opcionalis `readonlyRepo` flaget, es a scaffoldban egyetlen profilnev sem szerepel. Ez a bekotes marad a PR-ben, a script mellett. Egy szallitott profil sem allitja meg a flaget, tehat a gate itt egyelore inert. Ha ugy latod, hogy a bekotesnek is a profilokkal egyutt kell varnia, szolj, es lecsupaszitom a puszta scriptre.

Menet kozben egy masodik dolgot is javitani kellett rajta: a bekotott parancs most a `hookCommand()`-on megy at, mert a `develop` a branch nyitasa ota megkovetel egy idezett, abszolut interpretert. A regi `node <ut>` alak nalatok mar szabalyt sertett volna.

## 3. A yarn lelet

Pontosan ugy, ahogy irtad: `yarn\s+(install)?\b` -> `yarn\s+install\b`.

Egy dolgot teszek hozza a meresbol: a `?` elhagyasaval a puszta `yarn` (a Yarn 1 implicit installja) is elveszti a mentesseget. Ez igy helyes, mert nem tartalmaz sem atiranyitast, sem modosito igét, tehat ezentul a sajat erdemein megy at, nem egy atalanyos kihagyason. A regresszio harom helyen van kiszogezve: minden csomagkezelo valodi installja tovabbra is mentes, negy nem-install yarn alparancs mostantol elbiralas ala esik, es a `yarn install && yarn add x > <repo>/fajl` a masodik szegmensen bukik el.

## Meres

`npx tsc --noEmit` tiszta, teljes keszlet ezen a branchen: **352 fajl, 4610 zold, 1 skipped, 0 bukas**.

A #767 mar nem fugg ettol a PR-tol: a profilok, amiknek a gate kellett, egyik PR-ben sincsenek benne. Azt a branchet is rebase-eltem, kulon irtam ala.

Szotasz:
Thanks for this, and sorry for the long silence: the last word here was yours, on 2026-07-29, and we did not come back to you.

I measured the PR today before writing, so this is not a guess. It currently conflicts with `develop` (`mergeable=CONFLICTING`), and the CI results attached to the head are from when it still merged cleanly, so they no longer describe what a merge would produce. The last movement on the PR was 2026-09-04; `develop` has moved a long way since, which is our doing, not yours.

That means a review right now would be a review of a diff that has to change anyway, and the first blocker is not our response: it is a rebase, and only you can do it. So the honest order is the other way round from the one we let happen.

If you still want this in, rebase onto current `develop` and push. Say so here when it is up and I will take the review myself. If you would rather not spend more time on it after this wait, that is a fair answer too, and I would rather you told us than left it open.

One thing you can count on: the delay was ours, and it will not decide the outcome. Whether this gets merged depends on the change, not on how long it sat.

PR #664, opened 2026-07-20
feat(monitor): P0 C+D — memory-pressure monitor, gate, health (clean, isolated)

iszzu80-dev:
## P0 C+D Monitor — Clean Branch

This is the ISOLATED clean version of PR #663. Only P0-allowlisted files.

### Scope
- **14 files, +4,084 lines, 1 commit** (vs #663: 372 files, +60,800/-290, 139 commits)
- Zero costops files, zero audit/deliverable artifacts, zero unrelated features

### P0 Allowlist
| Category | Files |
|----------|-------|
| Monitor core | memory-pressure-monitor.ts (545L), gate.ts (496L), health.ts (307L), types.ts (164L) |
| Tests | gate.test.ts (314L/27), health.test.ts (649L/42), state-machine.test.ts (121L/19), release-regression.test.ts (422L), telemetry.test.ts (418L/32) |
| Scripts | dependency-closure-check.sh, install-monitor.sh, list-agent-rss.sh |
| Systemd | marveen-memory-monitor.service + .timer |

### Test Results (Clean Branch)
| Suite | Result |
|-------|--------|
| tsc --noEmit | PASS (zero errors) |
| memory-pressure-gate.test.ts | 27/27 PASS |
| memory-pressure-health.test.ts | 42/42 PASS |
| memory-pressure-state-machine.test.ts | 19/19 PASS |
| memory-pressure-telemetry.test.ts | 32/32 PASS |
| memory-pressure-release-regression.test.ts | 37/39 PASS (S12/S13 need live state — expected) |
| dependency-closure-check.sh | PASS (status: ok, 0 errors) |
| Full vitest suite | 179/181 files pass (2,444/2,447 tests) |

### Negative Control
- Disabled measurement-error fail-closed guard → gate test RED (22/27, 5 FAIL)
- Restored guard → gate test GREEN (27/27)
- Working tree clean

### Excluded (NOT P0 C+D)
- src/data-sensitivity-gate.ts, runner, test — different feature line
- All costops, audit, deliverable, artifact files from #663

### Provenance
Extracted via git-show from commit 31a27dc (PR #663 HEAD). New branch from origin/develop at 8abc374.

Issue #617, opened 2026-07-15
WSL2 fleet host: memory pressure (OOM) + vmIdleTimeout=-1 still suspends — how do others tune it?


iszzu80-dev:
Cross-posting a question relevant to anyone running the marveen fleet on WSL2 (Windows host). We hit two WSL2-specific problems running ~13 long-lived agent processes in one distro on a 16 GB host (WSL sees ~7.4 GB):

**1. Keep-alive / suspend.** Even with `[wsl2] vmIdleTimeout=-1` in `.wslconfig`, instances/services still get suspended after a while, so we run a PowerShell keep-alive loop as a workaround (matches microsoft/WSL #13291, #8659).

**2. Memory pressure.** WSL2 doesn't release RAM back to Windows, so `MemAvailable` creeps down until the OOM killer fires — which on a low-RAM host has cascaded into host-side instability / poweroffs. Current mitigations in `.wslconfig`:
```
[wsl2]
memory=11GB
swap=8GB

[experimental]
autoMemoryReclaim=gradual
```

Full question + more detail is in the upstream WSL Q&A: https://github.com/microsoft/WSL/discussions/41081

**Asking here specifically:** how are other marveen-fleet operators sizing `memory`/`swap`, and is anyone using something other than the PowerShell keep-alive loop? Any fleet-side mitigations (agent cap, staggered start, the memgate) that helped before touching `.wslconfig`?

iszzu80-dev:
Posting our findings + what actually helped, in case it's useful to others running a memory-heavy multi-process workload on WSL2.

## Two separate problems (they look related but aren't)

**1. `vmIdleTimeout=-1` still suspends** — this is a real, known upstream WSL bug, not a config mistake on your side. See microsoft/WSL #13291 (services suspended despite `vmIdleTimeout=-1`) and Discussion #8659 / #8654 (`vmIdleTimeout` "no effect"). A long-lived keep-alive process inside the distro (equivalent to `nohup sleep infinity &`, or a tiny systemd `--user` service) reliably works around it — it's compensating for the defect, not a hack.

**2. OOM → full VM poweroff** — for us this turned out to be **not** primarily a memory problem, even though it looked like one. Two things mattered here:

[...]

PR #775, opened 2026-07-29
feat(monitor): out-of-process memory-pressure monitor + fail-closed start gate

iszzu80-dev:
Adds an out-of-process memory-pressure monitor (systemd, zero LLM, reads /proc, hysteresis) plus a fail-closed start gate at the single startAgentProcess choke point, so a RAM-tight host refuses to spawn a new agent past the pressure threshold instead of OOM-killing the fleet.

Self-contained: no package.json / installer / dependency change. tsc --noEmit clean; full suite green (2918 passed).

REVIEW NOTE (disclosed): the 4 added test files each wrap a self-reporting harness inside ONE vitest case, so vitest reports 4 tests where there are ~28 internal sub-checks. Happy to split them into discrete vitest cases if you prefer that shape. Single commit, applies clean on develop.


Szotasz:
Thanks for consolidating four overlapping PRs down to this one, and for stating in the #664 close comment exactly what got dropped. That made this review much cheaper.

The module itself is good work: the gate sits at a single choke point, core status comes from explicit config rather than being inferred, and the health check that distinguishes "state file says normal" from "monitor is actually alive" is the right lesson from the 05:20 incident.

**Change request before this can merge.** Not a bug report: the behaviour below is deliberate and you locked it with a test ("normal with no state file yet (first boot) -> fail-closed for non-core"). My objection is about what it does on a host where the monitor cannot run at all.

**1. There is no platform guard, and the monitor is Linux-only.**

`memory-pressure-monitor.ts` reads `/proc/meminfo` and `/proc/pressure/memory`; the unit files are systemd. Nothing in the diff ever asks `process.platform`. On macOS the monitor can never produce a state file, so `readState()` returns null permanently, and `memoryPressureGate()` returns:

```
{ allowed: false, reason: "gate-error: no state file (fail-closed)" }
```

for every agent not on the core list. `DEFAULT_CONFIG.coreAgents` is `["marveen"]`, so that is the entire sub-agent fleet.

Measured on the host this repo runs on, just now:

```
uname -s        -> Darwin
ls /proc        -> No such file or directory
command -v systemctl -> not found
launchctl list  -> com.marveen.dashboard, com.marveen.channels, ... (launchd, not systemd)
tmux ls         -> 10 running agent-* sessions, none of them named "marveen"
```

So on that host, merging this blocks the start of all ten.

**2. The failure is delayed, which makes it worse.**

The gate is in `startAgentProcess()`, so already-running agents keep running. Nothing looks wrong until the next start or restart, which in practice means a context-guard restart, a crash recovery, or an operator restarting a stuck agent. The gate fires exactly on the recovery path, at the moment you least want a new hard stop, and it will read as "the agent is broken" rather than "a gate blocked it" unless someone finds the log line.

**3. Linux hosts are not automatically safe either.**

`install-monitor.sh` manages the `releases/` symlink; it does not install or enable the systemd unit. So after a merge the timer is not running anywhere until somebody enables it by hand, and until then the `no state file` branch is live on Linux too.

**4. What I am asking for: split "never installed" from "installed and broken".**

Right now both collapse into one fail-closed answer. They are different states and deserve different handling, the same three-state split used in #788 and #794:

- **never installed** (no state file, no unit, unsupported platform) -> unknown, not a pressure signal. Do not block; log once that the guard is absent.
- **healthy** -> decide on the state as you do today.
- **installed but stale or failing** -> terminal, keep the current fail-closed block. This is the case the guard exists for and it should stay strict.

Plus an explicit platform guard, so on a non-Linux host the gate is a no-op rather than a permanent closed door.

I am not asking you to weaken fail-closed. A guard that was working and then died must still block. A guard that was never there is a different claim.

**5. Two smaller points.**

(a) The description says "Self-contained: no package.json / installer / dependency change". Strictly true for existing files, but the PR does add a new 168-line `scripts/install-monitor.sh` and two systemd units. Worth rewording, the sentence promises less surface than the diff has.

(b) `releases/monitor-03c5c5b3e/` version-controls five compiled `.js` copies of the `.ts` sources plus a `monitor-current` symlink, and the install path serves from `releases/`. That means editing `src/web/memory-pressure-*.ts` changes nothing about the running monitor until the release dir is regenerated, with no error to signal the divergence. If the release dir has to be in the repo, please say in the PR how it gets regenerated and what fails loudly when the two drift.

**6. Carried over from #664.** You noted that branch also had `scripts/dependency-closure-check.sh` and `memory-pressure-release-regression.test.ts`, which this PR does not. Confirmed against the file list. Are those coming as a follow-up commit here? I would rather have them in this PR than lose them to the consolidation.

Happy to re-review as soon as 4 is addressed. The rest are comments, 4 is the blocker.

iszzu80-dev:
Pushed two commits (`acb1dfd`, `a857725`) addressing the review.

**BLOCKER #4** — fixed via a three-state split in `memoryPressureGate()`:
- Unsupported platform or no systemd timer unit found → **UNKNOWN**, allowed, logged once per process.
- Timer unit installed but no state file → **TERMINAL**, stays fail-closed (unchanged from before for this case).
- "Installed" evidence is the systemd timer unit file specifically, not the `releases/` symlink, since `install-monitor.sh` manages the release but not the unit — a release can exist with the timer never enabled, and that now reads as never-turned-on rather than broken.

**Minors:**
- 5a: description corrected (it did add an installer + two systemd units — my mistake originally).
- 5b: `install-monitor.sh --status` now compares the installed release's commit against the checkout's current HEAD and fails loudly (stderr banner, non-zero exit) on drift.
- 6: restored `scripts/dependency-closure-check.sh` and `memory-pressure-release-regression.test.ts` from `bd36767` and wired the closure check into `install-monitor.sh` as a hard gate — adapted the test to this PR's schema (v2-only scenarios from that source branch weren't ported, since this PR doesn't implement the v2 fields they assert on). Wiring the checker in immediately caught a real pre-existing issue (`process.cwd()` fallback in the release-copied files) — documented the one narrow exception rather than changing the fallback itself, since that fallback is currently load-bearing for the live dashboard's own copy of this code.

Verified dynamically: never-installed and unsupported-platform both allow non-core starts now; installed-but-not-reporting still blocks; drift detection fires correctly on a stale release and clears after reinstall. Full suite: 2918/2920 pass (the one failure is a pre-existing, unrelated flake that also fails/passes independent of this branch).

iszzu80-dev:
Hi @Szotasz — friendly follow-up on this one too. It's mergeable/clean and the review blockers were addressed (fixes pushed in `acb1dfd`/`a857725`). Happy to tweak anything if you'd like it shaped differently. Thanks, and no pressure on timing!

Szotasz:

Szotasz:
Szia, es elnezest a hosszu csendert. 35 napja all nyitva, es ez a mi mulasztasunk.

A PR a `develop`-pal **nem utkozik**, tehat a rebase most nem szukseges. Ami blokkol: **mindket CI-kapu piros**, a `test` es a `scan` is.

**Amit kerek:** nezd meg a ket bukott futast, es ha javithato, pusholj ra. Ha valamelyik bukas szerinted NEM a te valtozasodbol jon, azt is ird meg -- akkor nalunk a hiba, es utananezunk.

Ez egy nagy PR (29 fajl), ezert egy oszinte jelzes elore: ekkora feluletnel a review is hosszabb lesz, es valoszinuleg kerni fogok egy bontast, ha a kapu-resz es a monitor-resz kulon is megall. Ne kezdd el elore, eloszor a CI-t hozzuk zoldre.

Szotasz:
Status from our side, because this one will otherwise sit and it is not obvious why.

Both required checks are red on the current head. `test` failing is the ordinary case and yours to work through. The one worth calling out is `scan`, our secret gate, which is also failing. That gate is the reason nobody here has started a review: a red secret gate is where we stop, before the diff, every time.

It is not an accusation. The gate flags patterns, and on a 5755 line addition across 29 files a fixture, a sample config, or a test token is the common cause. But we do not read past it to find out, and I am not going to paste what it matched into a public thread.

What unblocks it: run the gate locally against your branch and clear what it reports, or tell us which file it stops on and we will confirm from our side whether it is a false positive worth an allowlist entry. Once `scan` is green the review can start, and the size means it will need a real slot rather than a drive-by.

One thing that would help the review a lot, independent of the gate: this is a large single PR with no deletions, which usually means several separable pieces travelled together. If it splits cleanly, smaller pieces will move much faster through here than one large one.

Szotasz:
Correcting my comment from 2026-09-06. It was wrong, and it has been wrong for five days.

**The gate never matched anything in your code.** I wrote that "the gate flags patterns" and that a fixture or a test token was the likely cause, and I declined to say what it matched. There was nothing to say. The gate did not find a pattern; it crashed and failed closed.

The cause is in our runner, not in your branch. Two of the paths in this PR are versioned symlinks pointing at directories (`releases/monitor-current`, `releases/monitor-previous`, mode 120000). The old gate resolved every path with `readFileSync`, which follows a link into a directory and throws `EISDIR`, and the gate stops on any read error by design. So a red `scan` here meant "could not read", and I reported it to you as "found something".

You then fixed that gate yourself in #1217 (`2282dcab`), and its own commit message says it plainly: "PR #775 sat red for two weeks with no secret in it." We shipped your fix and did not come back to this thread. That part is on me.

Measured just now, on this PR's exact range (`03c5c5b3..a8577259`) with the current gate:

```
secret-gate -- mode --range, 29 file(s) in scope, read from git objects
Symlinks (2) scanned as the blob git stores -- the target path as text, not the file it points at:
  - releases/monitor-current
  - releases/monitor-previous
PASS: no denied path, no secret shape, no channel material in 29 file(s).
```

Exit 0. There is nothing for you to clear, and there never was. (One limit worth stating: I ran the gate against your range directly. I could not run the gate's own test suite on this machine, because that suite refuses to run on a live install.)

**What is actually left is the `test` failure, and it is one assertion**, which my earlier comment should have told you instead of sending you after a secret that did not exist. Your new file is at 30 pass, 1 fail; the other 296 test files are green.

```
FAIL T6c: no state file -> non-core blocked (fail-closed, pre-existing behavior)
  { allowed: true, reason: gate-absent: monitor never installed on this host (no systemd unit found) }
```

It is not flaky. Your own three-state split in `memory-pressure-gate.ts` (from review #4) checks, when there is no state file, whether the monitor is installed at all, and returns `allowed: true` when it finds no systemd unit. T6c still asserts the older fail-closed behaviour. A GitHub runner has no unit, so it fails there every time. Whichever of the two is the behaviour you want, the other one needs to move.

On re-running: a plain re-run will not clear `scan`, because the run is pinned to its original event payload and would execute the old gate against the old base. Reading the workflow, a new push raises a new `pull_request` event, and since the checkout step takes no explicit `ref`, both the merge ref and the base resolve to the current `develop`. Since T6c needs a push anyway, one push should settle both. I have not measured that second part, only read it.

My note about splitting the PR still stands as a suggestion and nothing more. It was never a precondition, and it certainly is not one now that the gate turns out to have been our bug.

Szotasz:
## Review: the design holds up, and one measured defect stands between it and merge

First, the part that was ours: this PR sat red for two weeks on a `scan` failure that was **our gate, not your code**. It never found a secret; it stopped fail-closed on two version-controlled symlinks. That is fixed on develop, and I retracted the incorrect comment we left on 2026-09-06. The delay was not yours.

### What we verified

An independent reviewer measured this against head `226a1778`, with negative controls rather than reading the tests:

* **Scope**: 29 paths, 5970 added lines, **zero deletions**. The only change to existing code is `+12` in `src/web/agent-process.ts`, a gate call at the top of `startAgentProcess`. Narrow and readable.
* **The fail-closed claim is test-protected, not decorative**: breaking the decision branch (letting "unit installed, no state file" pass) turns the suite red in the two files that should go red, naming `T6c`.
* **The question we cared about, "what if the monitor itself dies?"**, measured across six states rather than inferred:

| state | result |
|---|---|
| darwin, no state file | allows (unsupported platform) |
| linux, unit not installed | allows (never installed) |
| linux, unit installed, no state file | blocks (fail-closed) |
| linux, unit + fresh normal + good release | allows (positive control) |
| linux, unit + 10-minute-old state | blocks (stale, 600s > 40s) |
| linux, unit + fresh warning | blocks |

A dead monitor does not look healthy: it blocks after 40 seconds. Keeping "never installed" and "installed but rotten" as separate branches is the right call.

### The one thing to fix before merge

`releases/monitor-current` and `releases/monitor-previous` are both new symlinks, and they point at the **same** release:

```
releases/monitor-current   -> monitor-acb1dfd8c
releases/monitor-previous  -> monitor-acb1dfd8c
```

Both are mode 120000 with the identical blob (`1809bdbeab`). So the rollback path is a no-op: rolling back to "previous" lands on the release you are already running, and `monitor-03c5c5b3e` (2026-07-29) ships in the tree with nothing pointing at it. For a watchdog whose whole justification is surviving a bad state, the rollback pointer is not a detail.

I think I know how it got there, and it is worth fixing at the cause rather than the symptom: **`src/web/memory-pressure-health.test.ts` writes into the working tree.** It creates `releases/monitor-<head-sha>/` and rewrites the tracked `releases/monitor-previous` symlink. That is invisible in CI, so a local `vitest run` before committing would silently move the pointer, which matches exactly what the diff shows.

Two small, specific changes:

1. Point `monitor-previous` at `monitor-03c5c5b3e`, so rollback reaches a different release.
2. Have the health test build its release fixtures under a temporary directory instead of the repo, so a test run cannot move a tracked pointer.

The second one matters to us beyond this PR: our deploy path bundles the working tree rather than `HEAD`, so a dirty tree after a test run can ship a wrong rollback pointer.

### One more, smaller

`scripts/systemd/marveen-memory-monitor.service` hardcodes an absolute path:

```
ExecStart=/home/iszzu/.bun/bin/bun run /home/iszzu/marveen/releases/monitor-current/memory-pressure-monitor.js
Environment=MARVEEN_HOME=/home/iszzu/marveen
```

On any other machine that unit points at a path that does not exist. A placeholder that `install-monitor.sh` substitutes at install time would fix it. Worth noting that `dependency-closure-check.sh` already guards the bundles against referencing the developer checkout, so the concern is clearly on your radar already; the unit file just did not get the same treatment.

### What we are not asking for

We are not asking you to change the release-directory design. Committing an immutable release the unit runs from, instead of executing out of a shared checkout, is a deliberate and sound choice for a watchdog, and the two committed releases exist so rollback has somewhere to go. Please do not widen the scope of this PR.

One thing our side should say plainly: on our own machines the gate is inert, because darwin is not a supported platform. This PR does not reduce memory pressure for us. The target is the WSL2 host from #617.

Szotasz (commented):
FLEET REVIEW -- NO-GO (in this shape; a concrete path to GO is listed below)

**Edited after posting, by the same reviewer. The NO-GO stands on the merits, but on different grounds than the first version said.** Two of the original five points are withdrawn below (CORRECTION block at the end); the four that carry the verdict are independent of them and were measured today: the full suite is red on today's develop, the relief path puts the token on a command line, the wiring has no test (negative control stays green), and two gates sit on the same start path next to an active relief/kill. This review is a CONTINUATION of the fleet's 2026-09-12 review, not a replacement: the two requests made there still stand.

First-round review by the fleet (Dani). Merge decision is the main agent's. Everything below was MEASURED on a merge of this PR's head (226a1778) onto today's `origin/develop` (e2b8a013, 594 commits past the PR base), in a worktree with its own `npm ci`. The old green checks were not used as evidence.

## What holds up

- Merge is clean (0 conflicts) although `src/web/agent-process.ts` changed 45 times since the base; `tsc --noEmit` exit 0.
- The PR's 5 test files pass on the merge tree (5 vitest cases, ~28 internal sub-checks).
- The gate logic is really tested: with `memoryPressureGate()` mutated to always allow (file-based mutant), 12 sub-checks go red and vitest exits 1; restored, green.
- The three-state split is sound and measured on our macOS host: unsupported platform -> `gate-absent`, allowed; no systemd unit -> allowed; unit present but no or stale state -> fail-closed. So the gate cannot freeze a host that never installed the monitor.

## Why NO-GO as-is

1. **The full suite is RED on today's develop.** `src/__tests__/tmux-stderr-attribution.test.ts` (TMUXWINDOWATTR920, merged 2026-09-20) fails: `memory-pressure-gate.ts` adds two sync by-name tmux calls that leak stderr (`capture-pane` in `isAgentIdle`, `kill-session` in `stopAgentFallback`). Merging as-is turns develop red on every host. Fix: pipe stderr the way the rest of `src/web` does (or, if there is a reason, name them in the allowlist with the reason); the test tells you the exact shape.
2. **The relief path puts the dashboard token on a command line.** `stopAgentViaApi` runs `curl ... -H "Authorization: Bearer ${token}"` through `execSync`, so the token is visible in `ps`/argv for the duration of the call. Our rule for every helper is: the token travels in the environment or stdin, never in argv (see `scripts/agent-msg.sh`). Fix: `curl -H @file` or a token file / env, or call the dashboard from node with `fetch`.
3. **Operator-specific paths and runtime baked into the tracked unit.** `scripts/systemd/marveen-memory-monitor.service` has `ExecStart=/home/iszzu/.bun/bin/bun run /home/iszzu/marveen/...` and `Environment=MARVEEN_HOME=/home/iszzu/marveen`. Every other unit in `scripts/systemd/` ships `/path/to/marveen/...` and node; `bun` is not part of this deployment's stack (only mentioned in three watchdog scripts' comments). This is the hardcode class we reject in distribution files. Fix: the `/path/to/marveen` template and node, like the sibling units.
4. **The health test writes into the working tree (this confirms the 2026-09-12 request rather than adding to it).** Correction to the first version of this review: I wrote that `monitor-previous` was a dangling symlink in the PR. That was wrong. At head 226a1778 both symlinks point at `monitor-acb1dfd8c`; the dangling pointer to `monitor-<merge-sha>` appeared in MY worktree after the full suite ran, because `memory-pressure-health.test.ts` creates `releases/monitor-<head-sha>/` and moves the tracked pointer. That is exactly item 2 of the 09-12 review (fixtures under a temp dir), measured again here from a different angle. The release-directory design itself is not in question; the fleet said so on 09-12 and that stands.

5. **Two gates with opposite failure semantics on the same start path.** Today `channel-monitor.ts` already asks `scripts/fleet-memory-gate.sh` before every reconcile start (fail-OPEN, never kills, safe-mode flag + Telegram alert). This PR adds a second, fail-CLOSED gate inside `startAgentProcess` and, in the monitor, an active relief that STOPS an idle non-core agent (API stop, fallback `tmux kill-session` + pkill). Not disqualifying by itself, but the PR text does not name the existing gate, and shipping both means a start can be refused for two different reasons with two different override switches (`MARVEEN_MEM_GATE_DISABLE` vs `MARVEEN_MEM_PRESSURE_OVERRIDE`). The fleet has to decide: replace, layer, or keep the relief/kill component out of the first merge. That is a design decision for the main agent, not a review nit.

## Measured limits of the tests (stated so nobody reads more into green than it says)

- Negative control on the WIRING: with `src/web/agent-process.ts` reverted to develop (gate call removed), all 5 PR test files still pass. The tests cover the gate module, not the fact that `startAgentProcess` consults it. A one-case test that asserts the refusal through `startAgentProcess` would close that.
- The monitor reads `/proc`; on macOS it returns zeros and the gate reports `gate-absent`. The monitor itself was not exercised here (no Linux host in this round); its state machine was exercised only through the PR's own hermetic tests.

## Path to GO

(1) stderr piping per TMUXWINDOWATTR920 so develop's suite is green on the merge; (2) token out of argv in the relief path; (3) template paths + node in the unit; (4) the two 09-12 items (monitor-previous -> monitor-03c5c5b3e, health-test fixtures under a temp dir); (5) a wiring test through `startAgentProcess`; and (6) the fleet's answer on how this relates to `fleet-memory-gate.sh` and whether the kill/relief component ships in the same PR. Items 1-5 are mechanical; 6 is the one that needs a decision before a re-review.

## CORRECTION (2026-09-22, same reviewer; visible on purpose)

The first version of this review, posted at 10:03 UTC, contained two things I am withdrawing.

1. **I claimed `releases/monitor-previous` was a dangling symlink in the PR.** That was false. At head 226a1778 both `monitor-current` and `monitor-previous` point at `monitor-acb1dfd8c` (measured with `git show 226a1778:releases/monitor-previous`). The dangling pointer to `monitor-<merge-sha>` that I saw appeared in MY worktree after the full suite ran, because `memory-pressure-health.test.ts` creates `releases/monitor-<head-sha>/` and moves the tracked pointer. So what I saw was not a defect in your commit; it was a second, independent reproduction of item 2 of the 09-12 review (health-test fixtures should live under a temp dir). The finding is ours to own, not yours.
2. **I asked for `releases/` to be dropped from the tree.** Withdrawn. The fleet said on 09-12 that the committed immutable release directory is a deliberate and sound choice for a watchdog, and that position stands. Two rounds asking for different things would be our failure, not yours; the 09-12 requests (point `monitor-previous` at `monitor-03c5c5b3e`, temp-dir fixtures) are the ones that apply.

What remains, and what the verdict rests on, is the four measured items above (1, 2, 3, 5 and the wiring negative control), all of which post-date or were outside the 09-12 review. The unit-path item (3) was already named on 09-12 as a smaller point; it is repeated here only because it is still unfixed.

PR #786, opened 2026-07-30
fix(alerts): stop re-alerting the owner every 30 minutes about a still-dead agent

theisszsolt:
## Problem

Three watchers re-alert the owner on a 30 minute cooldown for as long as the broken state persists:

- `reauth-healer.ts` `ESCALATION_COOLDOWN_MS` -- dead OAuth token on a live session
- `channel-monitor.ts` `PLUGIN_ALERT_DEDUP_MS` -- main channel plugin still down
- `channel-monitor.ts` `PANE_ERROR_DEDUP_MS` -- session wedged on a thinking-block API error

On 2026-07-30 one agent's dead token produced ~10 identical Telegram alerts in a single morning (`dashboard.log`: `reauth-healer: dead OAuth token on live session -- escalating to owner`). The owner asked for a few hours between repeats.

## Change

All three repeat cooldowns 30 min -> 3 h. Detection is untouched: `PANE_ERROR_CONFIRM_MS`, `DEAD_PROBE_THRESHOLD`, `AGENT_DOWN_CONFIRM_MS` and the first escalation of a down-spell keep their current timing, so a genuine outage is still reported as promptly as before. Only the informationless repeats get rarer.

Not touched: the `alert-busy` and max-restart-attempts branches in `channel-monitor.ts` already fire once per down-spell (`agentBusyDeferAlerted`), and the auto-restart WARN logs never reach Telegram.

## Test

`npm run build` clean; no test hardcodes the three constants (`grep -rn` over `src/`). Verified live on the Jarvis install: dashboard restarted, health 200, `dist/` carries the new values.

### DIFF
--- a/src/web/channel-monitor.ts
+++ b/src/web/channel-monitor.ts
@@ -177,7 +177,9 @@ const AGENT_BUSY_DEFER_MAX_MS = 30 * 60 * 1000 // 30m
 // fresh (re-stamped by each post-respawn probe, cleared on recovery).
 const PLUGIN_ABSENT_MAX_RESTART_ATTEMPTS = 1
 const PLUGIN_ABSENT_TTL_MS = 15 * 60 * 1000
-const PLUGIN_ALERT_DEDUP_MS = 30 * 60 * 1000
+// Re-alert cadence while the main plugin is STILL down. The first alert is the
+// informative one; keep repeats to a few hours (owner request 2026-07-30).
+const PLUGIN_ALERT_DEDUP_MS = 3 * 60 * 60 * 1000
 
 // Stuck channel-input recovery (MAIN session only). A channel notification
 // delivered while Boss is busy can be parked as plain text at the ❯ prompt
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged, thank you. The reasoning is sound and the verify confirmed both directions I cared about: the three dedup mechanisms keep separate state, so a *different* failure on the same agent still alerts immediately, and the mute clears on recovery (`handleMarveenUp` nulls the state, the pane-error entry expires after a clean pane, the healer resets on a live probe). A repeat alert that carries no new information is worse than useless, so this is the right change.

One request for next time, not a blocker on this PR.

The new comment in `reauth-healer.ts` cites an `owner request 2026-07-30` and quotes it. In your own install that provenance is clear, but this comment now lives in the upstream tree, where "owner" reads as *this* repository's owner and the date makes it look like a settled decision from him. Someone tuning the cadence back in six months would treat it as something they cannot change without asking.

If the request came from somewhere linkable (an issue, a discussion, a community post), please point to it and we will add the link. If it came from your own operator, a neutral wording is enough and just as informative, because the justification stands on its own:

> The first alert is the informative one. Every repeat after it carries no new information, so keep the cadence rare.

Same rule we apply to ourselves: an approval or a request referenced in code or in a PR body should be verifiable by whoever reads it next, otherwise it quietly becomes a fact nobody can check.

PR #794, opened 2026-07-30
fix(install-linux): surface the real npm error on install failure

jojo260426:
## Problem

The Linux installer (`install-linux.sh:713`) ran:

```bash
if ! (npm ci --loglevel warn 2>/dev/null || npm install --loglevel warn); then
  fail "npm install sikertelen. Ellenorizd a hibauzeneteket fentebb."
fi
```

The `2>/dev/null` discards `npm ci`'s stderr. When an install genuinely fails, the user is told to "check the messages above" but there are none.

This bites hardest on `better-sqlite3`, which has **no prebuilt binary for newer Node ABIs** (e.g. Node 22 / ABI 127) and so **compiles from source via node-gyp** on every install. On a memory-constrained box (observed on a 3.8 GB WSL2 Ubuntu 24.04 host) the `g++` pass can get OOM-killed, failing both `npm ci` and `npm install` — with no diagnostic surfaced.

## Fix

Capture `npm ci`'s stderr to a temp file and **replay it only when the `npm install` fallback also fails**. The happy path (benign lockfile-desync → silent fallback to `npm install`) stays quiet; real failures become actionable. The fail message now also names the likely culprit.

This mirrors `install-macos.sh:410`, which already avoids hiding stderr here and rebuilds `better-sqlite3` explicitly.

## Testing

- `bash -n install-linux.sh` passes.
- Verified a clean install on the affected host now succeeds (213 packages, native module loads, `tsc` build green) once memory pressure is relieved.

### DIFF
--- a/install-linux.sh
+++ b/install-linux.sh
@@ -710,9 +710,22 @@ echo -e "${BOLD}$(_t section_5)${NC}"
 cd "$INSTALL_DIR"
 
 echo -e "  npm install..."
-if ! (npm ci --loglevel warn 2>/dev/null || npm install --loglevel warn); then
-  fail "npm install sikertelen. Ellenorizd a hibauzeneteket fentebb."
+# Fast path: npm ci. On a benign lockfile desync it fails harmlessly and we
+# fall back to npm install, so ci's stderr is captured (not shown) to avoid a
+# scary EUSAGE dump on the happy path. But if the fallback ALSO fails -- e.g.
+# node-gyp getting OOM-killed while compiling better-sqlite3 from source, which
+# has no prebuilt binary for newer Node ABIs -- we replay ci's captured stderr
+# so the real error is visible instead of an empty "check above".
+_NPM_CI_LOG=$(mktemp)
+if ! npm ci --loglevel warn 2>"$_NPM_CI_LOG"; then
+  if ! npm install --loglevel warn; then
+    echo -e "  ${DIM}--- npm ci kimenet ---${NC}"
+    cat "$_NPM_CI_LOG" >&2
+    rm -f "$_NPM_CI_LOG"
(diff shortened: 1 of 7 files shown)


Szotasz:
Reading review is green. This does what I asked for in the #779 review, and the part that goes beyond the title is the more valuable half.

The stderr fix is right on its own: replaying `npm ci`'s captured stderr only when the `npm install` fallback also fails keeps the benign lockfile-desync path quiet while making a real failure actionable. That matters more than usual this week, we hit four separate install dead-ends on live workshop machines today and every one of them was a prerequisite step handing the user an opaque message.

The discriminated result is the substantive change. `createMainChannelsSession()` returning a boolean collapsed "already kicked, still booting" (benign) and "channels.sh missing or unlaunchable" (broken install) into the same `false`, and the onboarding launch endpoint then reported that as a silent success. Same three-state split we applied in #788 today: transient, terminal, and unknown must not share a return value.

**One thing I checked specifically, because this is where a change like this usually breaks something quietly.** Every non-`'started'` value is a truthy string, so any caller left on a boolean test would silently invert. On `origin/develop` there is exactly one production caller, `channel-monitor.ts:1566`, and this PR rewrites it to `=== 'started'`. The onboarding handler is new and covers all four arms. The chain is complete, no stragglers.

**One comment, not a blocker.** The three new tests are static source assertions via `sliceFn` and `indexOf`, not behavioural ones, so they lock the shape of the code rather than what it does. If someone later swaps the grace arm to return `'script-missing'`, the suite stays green. That is the existing pattern in this file so I am not asking you to change it here, but the guarantee is weaker than the test names suggest.

**Not merging yet, and the reason is process rather than this diff.** `src/web/channel-monitor.ts` is the recovery stack, and my standing rule is that changes there get an independent run of the build and the full suite by someone other than the reviewer. The agent who does that is blocked on a spend limit right now. The reading review is done and green; the merge waits on that run, which should be short.

PR #795, opened 2026-07-30
feat(ops): ship the operator ops scripts, portable instead of install-bound

mogganhun:
Four scripts have kept this install running for weeks but never made it into the repo, because each carried the install inside it: an absolute `/home` path, one tmux session name, one Telegram chat id, one GitHub slug, one node path. Useful tools, unusable by anyone else. This resolves every such value at runtime.

**What each one does, and what was baked in**

- **`limit-monitor.sh`** watches for Claude usage-limit signals in the logs and in the live channels pane, then alerts over the Bot API -- deliberately not a scheduled Claude task, since that would spend the quota it guards. The session name was the literal `bigme-channels`; it now derives from `MAIN_AGENT_ID` exactly as `channel-keepalive-probe.sh` does. That one mattered more than it looks: after a rename it grepped a session that no longer existed, found nothing, and reported healthy forever. With no `ALLOWED_CHAT_ID` it now logs why and exits instead of alerting a chat it guessed.
- **`pre-modify-backup.sh`** snapshots the state git does not track (SQLite via `.backup`, vault, runtime config) before a risky change. The repo root was one absolute path, so a second checkout would back up the *first* install's database and report success. It now resolves its own location. Which untracked scripts count as "personal" is per-install, so that list became data (`store/personal-scripts.txt`), falling back to whatever `git ls-files --others` reports under `scripts/` -- exactly the set git cannot restore.
- **`github-pr-monitor.sh`** polls our open PRs and alerts on state changes, token-free. The slug now derives from the `upstream`/`origin` remote, and the watched PRs come from `gh pr list --author @me --state open` instead of a fixed three-number list that went stale the day it was written. `GITHUB_PR_MONITOR_REPO` / `_PRS` override both, for a fork watching upstream. `node` resolves via PATH first, with the previous absolute paths kept as fallbacks for the thin PATH a systemd --user timer inherits.
- **`hooks/telegram-ack.py`** posts a 👍 reaction the moment an inbound message arrives, before the model processes the turn -- fleet-wide, each agent on its own bot token. This one was already install-neutral; only the docstring named one operator.

**Test**

`scripts/__tests__/ops-scripts-portable.test.sh` pins the rule rather than the code: no absolute home path, no literal chat id, no literal repo slug, the session name follows a renamed install, a missing owner chat stays silent, and a backup lands in its own checkout. I verified it fails when a literal is put back (both the slug check and the silent-exit check trip).

**Shipping is not enabling**

Merging this makes the four scripts available, not active. Nothing in the repo wires any of them up: `templates/settings.json.template` lists its hooks by name and `telegram-ack.py` is not among them, and the three shell scripts need a cron entry or a hand run before they do anything. That is deliberate for anything that can post to the channel path -- an operator opts in per script. The title reads the other way, so to be unambiguous: after this merge, nothing new starts running on anyone's install.

Operator-facing docs for the four (what each watches, the cron line or settings entry each needs) are not in this PR; tracked as a follow-up.

**Note on adoption**

These land as tracked files, so an install already running untracked copies under the same names will hit a pull conflict once. Removing the local copy is the whole fix -- the tracked version reads the same values from `.env`.

### DIFF
--- a/scripts/__tests__/ops-scripts-portable.test.sh
+++ b/scripts/__tests__/ops-scripts-portable.test.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+# Portability contract for the operator-side ops scripts.
+# Run: bash scripts/__tests__/ops-scripts-portable.test.sh
+#
+# These four scripts were written against one install and carried its values in
+# the source: an absolute home path, one tmux session name, one chat id, one
+# GitHub slug. Anyone else cloning the repo got scripts that silently watched
+# the wrong things. The tests below pin the rule that replaced them -- every
+# install-specific value is resolved at runtime -- so the next edit cannot
+# quietly bake a local value back in.
+
+set -u
+
+PASS=0; FAIL=0
+TMPDIR_BASE="$(mktemp -d)"
+trap 'rm -rf "$TMPDIR_BASE"' EXIT
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+SCRIPTS="
+scripts/limit-monitor.sh
+scripts/pre-modify-backup.sh
+scripts/github-pr-monitor.sh
+scripts/hooks/telegram-ack.py
+"
+
+echo "ops-scripts portability tests"
+echo "============================="
+
+# ---------------------------------------------------------------------------
+# (a) No install-specific literals anywhere in the sources
+# ---------------------------------------------------------------------------
+echo ""
+echo "(a) No baked-in install values"
+# An absolute /home/<user> path, a bare numeric chat id assignment, or a
+# hardcoded owner/repo slug are the three shapes that broke portability.
+for rel in $SCRIPTS; do
+  f="$INSTALL_DIR/$rel"
+  [ -f "$f" ] || { fail "$rel exists"; continue; }
+  if grep -qE '/home/[a-z]+/' "$f"; then
+    fail "$rel: absolute /home path"
+  elif grep -qE '^[[:space:]]*(CHAT_ID|ALLOWED_CHAT_ID)=["'"'"']?[0-9]{6,}' "$f"; then
+    fail "$rel: hardcoded chat id"
+  elif grep -qE '^[[:space:]]*REPO=["'"'"'][A-Za-z0-9_-]+/[A-Za-z0-9_.-]+["'"'"']' "$f"; then
+    fail "$rel: hardcoded GitHub slug"
+  else
+    pass "$rel: no baked-in install values"
+  fi
+done
+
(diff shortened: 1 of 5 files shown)


Szotasz:
Green from me. The reasoning in the description is the part I want to keep on the record: these scripts had been running your install for weeks and never landed in the repo because each one had swallowed the install it was born in. An absolute home path, one tmux session name, one chat id, one repo slug. That is the same class we hit twice this week from the other direction, a per-host value sitting in a version-controlled file.

I did not take the portability claim on trust, so here is what I actually ran rather than read.

**Parser check on all five files.** Reconstructed from the diff, since every file is new: `bash -n` clean on the four shell scripts, `py_compile` clean on `telegram-ack.py`.

**Your own portability test, executed.** 7 PASS / 0 FAIL on the shipped sources.

**Negative control, which is the part that actually decides whether that test is worth anything.** I put each of the three banned shapes back, one at a time, and re-ran:

| injected | result |
|---|---|
| `cd /home/istvan/marveen` in limit-monitor.sh | FAIL: absolute /home path (plus 2 knock-on failures) |
| `ALLOWED_CHAT_ID=<literal id>` in limit-monitor.sh | FAIL: hardcoded chat id |
| `REPO="owner/repo"` in github-pr-monitor.sh | FAIL: hardcoded GitHub slug |

Restored, back to 7/0. The test fails for the reasons it claims to, which is more than I can say for a lot of guard tests, and case (c) is the one I would have asked for if it were not already there: a fresh install with no `ALLOWED_CHAT_ID` exits with a logged reason instead of guessing a chat to alert.

**Two things to note, neither blocking.**

**1. Shipping is not enabling, and the title reads the other way.** Nothing in the repo wires any of these up. `templates/settings.json.template` lists its hooks by name and `telegram-ack.py` is not among them, and the three shell scripts need a cron entry or a hand run. So after the merge these are available, not active. I think that is the right default for anything touching the channel path, but "ship the operator ops scripts" can be read as "they start running", so please be explicit about it in the PR description.

**2. Nothing tells a new operator these exist.** The portability work makes them usable on someone else's install, and then no doc mentions them. A short section, four scripts, what each watches, the cron line or settings entry needed, would finish the job the PR starts. Happy for that to be a follow-up rather than a hold on this one.

Merging this is waiting on the same run gate as the other two open PRs, on the agent that is currently blocked on a spend limit. Nothing in this diff worries me though.

mogganhun:
Thanks -- the negative control is the part I would have asked for too, good to have it on the record.

**On (1), shipping is not enabling.** You are right that the title reads the other way. I have added a *Shipping is not enabling* section to the description that says it plainly: nothing in the repo wires these up, `telegram-ack.py` is not in `templates/settings.json.template`, and the three shell scripts need a cron entry or a hand run. Nothing new starts running on anyone's install after this merge.

**On (2), docs.** Agreed it finishes the job, and agreed it is a follow-up rather than a hold. I will send a separate PR with a short operator section: the four scripts, what each watches, and the cron line or settings entry each needs.

No code change in this push -- description only.

PR #797, opened 2026-07-30
fix(quarantine-reader): derive the fetch allowlist from the owner's egress gate

jkarcsi:
The same decision lived in two places. `store/egress-allowlist.json` is the owner's
gate for outbound calls; the quarantine reader carried a **second** domain list,
baked into its template and maintained by hand. Both halves of the drift bit a live
install on consecutive days:

- **2026-07-29** a domain was on the egress gate but not in the reader, so every fetch
  to it failed with `domain not on quarantine-reader fetch allowlist` while the
  operator was looking at an allowlist that said it was allowed.
- **2026-07-30** the operator added it to the deployed reader definition instead, and
  the next boot silently reverted the edit. `ensureQuarantineReader` compared the
  deployed file with the template and copied the template over it in *either*
  direction, while its docstring claimed it only wrote "when the template is newer".

**The fix.** The owner's domains become an *input* to the deploy.
`ownerAllowedDomains()` reads them from the egress allowlist, and
`renderQuarantineReader()` splices them into the shipped list inside a
marker-delimited block. A re-render replaces that block, so:

- the two lists cannot diverge;
- revoking a domain in the egress file revokes it in the reader too (a revoked
  permission that keeps working is the worse failure);
- the boot can no longer erase an owner-approved domain, it reproduces it.

Stripping the previous block is done by literal string position, not with a
`RegExp`: the markers contain parentheses, a dot and a slash, and the first version's
unescaped pattern turned `(from store/egress-allowlist.json)` into a capture group
that never matched the literal text. The revoke test caught that.

**Tests** (`src/__tests__/quarantine-allowlist-render.test.ts`, 13 cases). Render:
placement inside the domain section rather than at end of file, every shipped domain
surviving, idempotency across repeated renders, no duplicate line when the template
already ships the domain (case-insensitively), an empty owner list leaving the
template byte-identical, revoke removing the line again, and a template with no
bullets at all. Reader: happy path, trimming, non-string entries, malformed file,
missing file, and a file without a `domains` key -- all of which must mean "no extra
domains" and never throw, because a throw here leaves an agent with no reader
definition at all.

**Verification.** Build clean. Full suite: 3032 passed, 1 skipped, 0 failed. One
earlier run of this same commit showed a single failure that did not reproduce in two
further runs on a loaded box, and I could not pin which test it was -- reporting it
rather than claiming three clean runs.

**Note for reviewers:** this repo also documents the two-list rule in per-install
rule files (ours says "two allowlists, both are needed"). With this change that
sentence becomes wrong, so installs that copied it will want to update it. Mine did.

### DIFF
--- a/src/__tests__/quarantine-allowlist-render.test.ts
+++ b/src/__tests__/quarantine-allowlist-render.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect } from 'vitest'
 import { mkdtempSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
-import { renderQuarantineReader, ownerAllowedDomains } from '../web/agent-scaffold.js'
+import { renderQuarantineReader, ownerAllowedDomains, isPublicFetchHost } from '../web/agent-scaffold.js'
 
 // The quarantine reader may only fetch from an allowlist, and that list used to
 // exist TWICE: once in this template and once in store/egress-allowlist.json,
@@ -104,5 +104,70 @@ describe('ownerAllowedDomains', () => {
   it('a file without a domains key is not an error either', () => {
     writeFileSync(join(dir, 'egress-allowlist.json'), JSON.stringify({ note: 'empty for now' }))
     expect(ownerAllowedDomains(dir)).toEqual([])
+  })
+})
+
+// The egress gate and the reader are edited with different threat models: the
+// gate answers "may the main agent call this host" (a LAN box is ordinary), the
+// reader answers "may a fetch target be steered here". Inheriting the first into
+// the second without a filter widens the inward-facing boundary. Reported on
+// #797 with five values that all passed through before this.
+describe('isPublicFetchHost', () => {
+  it('rejects the five values that reproduced in review', () => {
+    for (const bad of ['127.0.0.1', 'localhost', '169.254.169.254', '192.168.1.50', '*']) {
+      expect(isPublicFetchHost(bad)).toBe(false)
+    }
+  })
+
+  it('rejects IP literals generally, not just the private ones', () => {
+    // A fetch target is a name; an address skips the name check entirely.
+    for (const ip of ['8.8.8.8', '10.0.0.1', '172.16.4.9', '0.0.0.0', '1.2.3.4']) {
+      expect(isPublicFetchHost(ip)).toBe(false)
+    }
+  })
+
+  it('rejects internal suffixes and single-label names', () => {
+    for (const bad of ['printer.local', 'db.internal', 'box.lan', 'wiki.intranet', 'nas', 'router.home']) {
+      expect(isPublicFetchHost(bad)).toBe(false)
+    }
+  })
+
+  it('rejects anything carrying a scheme, port, path, space or wildcard', () => {
+    for (const bad of ['https://claude.com', 'claude.com:8080', 'claude.com/blog', 'claude com', '*.claude.com', '']) {
+      expect(isPublicFetchHost(bad)).toBe(false)
+    }
+  })
+
+  it('accepts ordinary public hostnames', () => {
+    for (const ok of ['claude.com', 'docs.claude.com', 'hnrss.org', 'feeds.bbci.co.uk', 'export.arxiv.org']) {
+      expect(isPublicFetchHost(ok)).toBe(true)
+    }
+  })
+
+  it('is case and whitespace tolerant', () => {
+    expect(isPublicFetchHost('  Claude.COM  ')).toBe(true)
+  })
+
+  it('drops the rejected entries from ownerAllowedDomains instead of failing the read', () => {
+    const dir = mkdtempSync(join(tmpdir(), 'egress-filter-'))
+    writeFileSync(join(dir, 'egress-allowlist.json'), JSON.stringify({
+      domains: ['claude.com', '127.0.0.1', 'localhost', '169.254.169.254', '192.168.1.50', '*', 'docs.anthropic.com'],
+    }))
+    // The good ones survive; a hostile line does not take the whole file down.
+    expect(ownerAllowedDomains(dir)).toEqual(['claude.com', 'docs.anthropic.com'])
+  })
+})
+
+describe('renderQuarantineReader anchoring', () => {
+  it('puts the block in the Domain restriction section even when a later section has bullets', () => {
+    const tpl = [
+      '## Domain restriction', '',
+      'Only fetch URLs from these approved domains:',
+      '- `hnrss.org`', '',
+      '## Output format', '',
+      '- `url` the requested URL', '',
+    ].join('\n')
+    const out = renderQuarantineReader(tpl, ['claude.com'])
+    expect(out.indexOf('- `claude.com`')).toBeLessThan(out.indexOf('## Output format'))
   })
 })
--- a/src/__tests__/quarantine-allowlist-render.test.ts
+++ b/src/__tests__/quarantine-allowlist-render.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest'
+import { mkdtempSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { renderQuarantineReader, ownerAllowedDomains } from '../web/agent-scaffold.js'
+
+// The quarantine reader may only fetch from an allowlist, and that list used to
+// exist TWICE: once in this template and once in store/egress-allowlist.json,
+// maintained by hand. The two drifted, and the deploy silently reverted the
+// hand-edit. These tests pin the fix: the owner's list is an INPUT to the
+// render, so a re-render cannot erase the owner's decision.
+const TEMPLATE = `---
+name: quarantine-reader
+---
+
+## Domain restriction
+
+Only fetch URLs from these approved domains. Reject all others:
+- \`status.anthropic.com\`
+- \`hnrss.org\`
+- \`www.reddit.com\` (RSS feeds only)
+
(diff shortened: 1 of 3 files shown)


Szotasz:
Good diagnosis, and the write-up is unusually honest: naming the RegExp bug your first version shipped, and the fact that the revoke test is what caught it, tells me more about the test quality than the case count does. The root cause is right too. Two copies of one decision, maintained by hand, drifting on consecutive days, and a docstring that described a behaviour the code did not have.

I measured the merge direction before reading the implementation, because combining two lists can be a posture change rather than a cleanup. On this install `store/egress-allowlist.json` holds four domains, all Anthropic documentation and platform hosts, and the shipped reader list is public news, RSS and status feeds. The union is harmless here. Two things about the general case are not covered by the PR, though.

**1. The two lists do not mean the same thing, and the reader's own description now contradicts the behaviour.**

The shipped description says the sub-agent is for "public APIs that are NOT on the main-agent egress allowlist". That is the reader's original role: the egress gate is what the main agent may call directly, and the reader exists to pull in what is not on it, precisely because that content is untrusted. This PR makes the egress list an input to the reader's list, so after the merge the description states the opposite of what the code does.

The 2026-07-29 incident you describe is real and worth fixing, so I am not arguing against the direction. But the description is what the model reads when it decides whether to route a fetch through the reader at all. Please update it in the same PR, otherwise the next person to debug a routing decision gets a document that disagrees with the deploy.

**2. `ownerAllowedDomains()` takes any non-empty string, including private and internal hosts.**

The filter is `typeof d === 'string' && !!d.trim()`. Nothing rejects `localhost`, `127.0.0.1`, RFC1918 ranges, `*.local`, or an internal hostname. That is fine for an egress gate, where an owner adding their own dashboard host is an ordinary thing to do. It is a different claim on the reader, whose domain list is the boundary that keeps a fetch target from being steered somewhere internal.

The exposure is indirect rather than dramatic: the reader only fetches what its caller asks for, and the caller is the main agent, but the main agent is exactly the thing that can be influenced by content the reader brought in earlier. The domain list is the backstop for that loop, and this PR widens it as a side effect of a change an owner would reasonably make for an unrelated reason.

I am not asking for a full SSRF policy. Either filter private and loopback targets out of `ownerAllowedDomains()`, or state in the code comment that the egress gate is trusted to that level on purpose. What I do not want is for the decision to be made implicitly by a list that was written with a different threat model in mind.

Neither point is a correctness blocker and the implementation itself is clean, including the deliberate choice of literal string position over a RegExp. Fix 1 in this PR, answer 2 either way, and I will merge once the run passes.

Szotasz:
Follow-up on my earlier review: the run-time verification is done, and point 2 is no longer a theoretical objection. It reproduces.

**The full suite is green on this branch.** `tsc --noEmit` clean, 221/221 files and 3036/3036 tests pass, which includes the 13 new ones this PR adds. Nothing here breaks anything.

**Merge direction, measured rather than read.** I asked whether combining the two lists narrows or widens the reader's allowlist. It widens, by construction: `renderQuarantineReader` unions rather than replaces, so the reader's list can never shrink, only grow. Running the pure function against the live store on this install:

```
template:        11 domains
after render:    15 domains
removed:          0
added:            4   (platform.claude.com, docs.claude.com, docs.anthropic.com, www.anthropic.com)
```

Four legitimate Anthropic documentation hosts, harmless in themselves. The posture change is the point, not those four: from here on the reader permits whatever the owner's egress gate permits, plus the shipped eleven.

**Point 2 reproduces, and it is worse than I described.** I fed five hostile values through `ownerAllowedDomains()`. All five landed unchanged in the reader's allowlist:

```
127.0.0.1
localhost
169.254.169.254      <- cloud metadata endpoint
192.168.1.50
*                    <- not a domain at all
```

The filter is `typeof d === 'string' && !!d.trim()`, so nothing rejects loopback, private ranges, link-local, or a bare wildcard. A single typo or a hostile line in `store/egress-allowlist.json` now widens the boundary that keeps a fetch target from pointing inward. That file is edited as an egress gate, by someone thinking about outbound calls, who has no reason to expect it also governs what the quarantine reader may reach.

Please validate the owner-supplied entries as public hostnames and drop loopback, private, link-local and wildcard values. This is the one blocking item.

**One more thing I want to flag, not blocking.** The insertion anchor is the last backtick-bullet in the file. That is correct today, because all eleven bullets sit in the Domain restriction section and that section ends the file. It is fragile though: the moment a backtick-bullet appears in any later section, the per-install block silently relocates there. Anchoring on the `## Domain restriction` heading instead would survive that.

Credit where due: stripping the previous block by literal `indexOf` rather than a `RegExp` is the right call given the parentheses and dots in the markers, and your own comment documents the real bug that taught you it.

Green on the run, blocked on the validation. Fix that and the description contradiction from my earlier comment, and it goes in.

jkarcsi:
Both fixed, and the blocking one was right to block: I had made the reader inherit a list written for a different question.

**Validation.** `isPublicFetchHost()` now rejects IP literals of any kind, single-label names, the internal suffixes (`.local`, `.internal`, `.lan`, `.home`, `.arpa`, …), and anything carrying a scheme, port, path, space or wildcard. Your five reproduce as rejected on the shipped function:

```
127.0.0.1          false
localhost          false
169.254.169.254    false
192.168.1.50       false
*                  false
claude.com         true
```

I rejected **all** IP literals, not just the private ranges. A fetch target here is a name, and an address skips the name check entirely, so allowing public addresses would only buy a way to phrase a target that the rest of the list cannot reason about. A rejected entry is dropped rather than fatal: one hostile line must not take the file down and leave an agent with no reader definition at all.

**Description.** You were right that the PR turned it into the opposite of the behaviour. It now says to route every fetch through the reader whether or not the host is on the egress allowlist, with the reason stated: being allowed to reach a host says nothing about trusting what the host returns. That sentence is what the model reads when it decides whether to use the reader, so a stale one misroutes fetches rather than merely confusing a human.

**Anchor.** Taken as well. It now anchors inside the `## Domain restriction` section instead of on the last backtick-bullet in the file.

**Verified the way you did it.** Against the live store: 13 owner entries all pass the filter, template 11 domains renders to 13, nothing removed. Tests now 21 cases, including a mixed allowlist where the good entries survive a hostile neighbour. Full suite: 3043 passed, 1 skipped, 0 failed.

One thing I want to name rather than leave implicit, since you framed it as a posture question: after this PR the reader's boundary is *derived* from the egress gate plus a public-host filter. That is narrower than what I shipped first, but it is still wider than the shipped eleven, and the widening happens whenever the owner edits the gate. If you would rather the reader keep its own list and the two be reconciled by a check that *fails loudly* on drift instead of merging, say so -- that is a different design and I would rather hear it now than after it is in.

Szotasz:
Re-measured on the new head. Both change requests are answered, and one of my own review questions turned out to be asked on the wrong axis. Details, because the second part changes what this PR is.

**1. The five inward-facing values are rejected.** `127.0.0.1`, `localhost`, `169.254.169.254`, `192.168.1.50` and `*` all fail `isPublicFetchHost`. I ran them myself rather than relying on the new test file, and added a wider adversarial set: IPv6 in several spellings, octal and decimal IPv4 (`0177.0.0.1`, `2130706433`, `127.1`), a trailing dot, `%2e` encoding, a unicode full-width dot, and the internal suffixes. All rejected. Good work.

**2. "Does the derivation narrow or widen?" was the wrong question, and the answer is neither.** I asked it because a derived list can silently open a boundary. So I measured the boundary instead of arguing about it:

- The egress-gate hook fires on **sub-agent** WebFetch calls, not only on the main agent's. Verified with a positive control first (a deliberately blocked main-agent call, to prove the block log actually writes), then the same URL through the quarantine-reader: blocked, and logged.
- So the reader's prose list is not the boundary. The hook is, both before and after this PR.
- On the install I measured, the two lists have an **empty intersection**: every domain the reader's prose list allows is blocked by the hook, and every domain the hook allows is refused by the reader itself. The reader could not fetch anything at all.

That reframes the change: it does not widen an effective boundary, it makes the reader's prose agree with what the hook already permits, so the fetch path becomes usable at all. That is a repair.

**3. One residual, not a blocker for this PR.** The domains the template ships stay unreachable after this merge, because they are not on the egress allowlist either. Restoring that path needs a separate decision (allowlist entries, or a sub-agent-aware branch in the hook), and it is being tracked separately.

**4. Small hardening idea, also not a blocker.** `isPublicFetchHost` rejects address literals but accepts public *names* that resolve inward: `127.0.0.1.nip.io`, `10.0.0.1.nip.io` and `192-168-1-50.sslip.io` all pass, and so does `kubernetes.default.svc` while `db.svc.cluster.local` is rejected. Reaching it requires an entry on the allowlist in the first place, so it is defence in depth rather than an open door, but it is the same class the filter exists for.

jkarcsi:
Thanks for measuring the boundary instead of arguing it — the empty-intersection finding is the real story: the reader's prose list had drifted off the egress hook entirely, so this makes the fetch path usable at all rather than widening anything.

**On residual (3).** The shipped domains are now partly restored from the egress side. The owner approved adding the news/docs sources today, so `anthropic.com`, `openai.com`, `blog.google`, `deepmind.google`, `huggingface.co`, `simonwillison.net`, `ai.meta.com`, `github.com` and `raw.githubusercontent.com` are now on `store/egress-allowlist.json` (and mirrored into the reader's prose list on this install). That is the 'allowlist entries' path you flagged; the sub-agent-aware hook branch stays the cleaner long-term fix and remains tracked.

**On hardening (4).** The `nip.io`/`sslip.io` + `kubernetes.default.svc` class is real. Since reaching it still needs an allowlist entry it is defence-in-depth, so I will take it as a follow-up on `isPublicFetchHost` (reject a name whose resolved literal is inward) rather than fold it into this PR.

PR #803, opened 2026-07-30
fix(security): hook commands use absolute node path (bare node = silent governance-gate bypass on nvm)

VTolnai:
## Root cause: DESIGN error, not configuration

The gate-injection code in `agent-scaffold.ts` wrote PreToolUse hook commands as bare `node <script>`, implicitly assuming `node` is resolvable in the hook's execution environment. Hooks run under `/bin/sh -c` with a **non-interactive PATH**: on nvm-based installs `node` is absent there, the hook exits 127, and Claude Code treats a non-2 exit as a **non-blocking** hook error -- the tool call is let through.

Net effect: all three governance hard-gates (**email-send-gate**, **self-pace-gate**, **egress-gate**) were silently non-enforcing on nvm installs, while looking fully wired in settings.json. Found in production via repeated `/bin/sh: 1: node: not found` PreToolUse warnings in an agent pane; confirmed by a two-mode test (bare `node` -> exit 127, call passes; absolute node -> `permissionDecision: deny`).

## Fix

- `HOOK_NODE_BIN = process.execPath` (the absolute binary of the node running the server) used in all four gate-command sites (email, self-pace, egress inject + egress ensure).
- `ensureEgressGate` now treats a legacy bare-`node` entry as NOT wired, so it upgrades it in place at startup instead of skipping.
- New `ensureGovernanceGateCommands` startup migration upgrades existing sub-agents' email/self-pace gate commands without waiting for the next respawn.

## Verification

- Full suite green on this base: 222 files, 3048 passed / 1 skipped.
- Live verify after deploy: every agent's settings.json carries absolute-path gate commands; functional deny test blocks a non-allowlisted WebFetch and logs to `store/egress-blocked.log`.

### DIFF
--- a/src/__tests__/hook-command-quoting.test.ts
+++ b/src/__tests__/hook-command-quoting.test.ts
@@ -1,0 +1,120 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import { mkdirSync, rmSync, readFileSync, writeFileSync, existsSync } from 'node:fs'
+import { join } from 'node:path'
+import {
+  HOOK_NODE_BIN,
+  hookCommand,
+  hookCommandWired,
+  injectEmailSendGate,
+  injectSelfPaceGate,
+  injectEgressGate,
+  ensureEgressGate,
+  ensureGovernanceGateCommands,
+} from '../web/agent-scaffold.js'
+import { PROJECT_ROOT } from '../config.js'
+
+// Review feedback on PR #803, pinned as tests:
+//  1. every injector must write a QUOTED absolute interpreter path -- an
+//     unquoted execPath with a space in it is split by `sh -c`, exit 127,
+//     silently non-enforcing: the exact failure the PR closes, reintroduced;
+//  2. the wired-already comparison must be the JSON-escaped one everywhere,
+//     otherwise a backslash (Windows) path never settles and every boot
+//     rewrites settings.json;
+//  3. the ensure* migrations must be idempotent: second call returns false.
+
+const TEST_AGENT = 'hookquoting-test-agent'
+const testAgentDir = join(PROJECT_ROOT, 'agents', TEST_AGENT)
+
+afterEach(() => {
+  rmSync(testAgentDir, { recursive: true, force: true })
+})
+
+function ptuCommands(settings: Record<string, unknown>): string[] {
+  const hooks = settings.hooks as Record<string, unknown>
+  const ptu = hooks.PreToolUse as { hooks: { command: string }[] }[]
+  return ptu.flatMap((e) => e.hooks.map((h) => h.command))
+}
+
+describe('hookCommand builder', () => {
+  it('quotes BOTH the interpreter and the script path', () => {
+    const cmd = hookCommand('/some/dir/gate.mjs')
+    expect(cmd).toBe(`"${HOOK_NODE_BIN}" "/some/dir/gate.mjs"`)
+  })
+})
+
+describe('injectors write a quoted absolute interpreter, never a bare node', () => {
+  for (const [label, inject] of [
+    ['injectEmailSendGate', injectEmailSendGate],
+    ['injectSelfPaceGate', injectSelfPaceGate],
+    ['injectEgressGate', injectEgressGate],
+  ] as const) {
+    it(label, () => {
+      const s: Record<string, unknown> = {}
+      inject(s)
+      const commands = ptuCommands(s)
+      expect(commands.length).toBeGreaterThan(0)
+      for (const cmd of commands) {
+        expect(cmd.startsWith(`"${HOOK_NODE_BIN}" "`)).toBe(true)
+        expect(cmd).not.toMatch(/^node /)
+      }
+    })
+  }
+})
+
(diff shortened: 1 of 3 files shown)


Szotasz:
Verified the diagnosis and the fix direction: both are right. A bare `node` in a hook command is resolved against the non-interactive PATH the hook runs under, and where that lacks node the hook exits 127, which is treated as a non-blocking error, so the gate silently stops enforcing while `settings.json` still looks fully wired. Using the absolute binary of the running server is the correct answer, because that binary exists by definition on the host that spawns the agents.

Three things I would fold in before it lands. The first two are measured, not read off the diff.

**1. Two different comparisons for the same job, in the same file, and they disagree.** The new `ensureGovernanceGateCommands` compares JSON-escaped (`JSON.stringify(cmd).slice(1, -1)`), which is right. `ensureEgressGate` searches for the raw command string inside the JSON. On a backslash path the two disagree: I ran both against the same input, and the raw form returns false exactly where the escaped form returns true.

The effect is on the migration, not on enforcement: the re-injected command is still the correct absolute one, so the gate does enforce. What never settles is the idempotency check. On such an install every boot sees "not wired", rewrites `settings.json` and logs another backfill line, indefinitely. The escaped form you already used elsewhere is the one to keep.

**2. A space in the node path reproduces the very failure this PR fixes.** The command is assembled unquoted, so wherever the interpreter path contains a space (a native Windows install under `C:\Program Files\nodejs`, or a home directory with a space in it) the shell splits it at the space, the hook dies with exit 127, and the gate is silently non-enforcing again. That is precisely the failure mode this PR exists to close, reintroduced through a different door.

It also passes the safety check unnoticed: I ran that check on five variants of the command string, and it reports the spaced and Windows-style forms as safe, so the broken command is written into settings without complaint. Quoting both parts fixes it, and I checked that the quoted form still passes the same check, so there is no side effect:

```
`"${HOOK_NODE_BIN}" "${scriptPath}"`
```

**3. No test in the PR.** For a fix whose subject is "the gate looked wired and was not", it would help to pin the behaviour. Two cheap ones in the style of the existing hook tests: that each injector writes an absolute interpreter path rather than a bare `node`, and that calling the ensure function twice returns false the second time. The second would fail today on a backslash path, so it also covers point 1.

**One question, not a claim.** The comment says the startup migration upgrades running agents without waiting for a respawn. Does a running session actually re-read `settings.json`, or does the change take effect at the next session start? I did not restart anything to check, so I am asking rather than asserting.

For context on where this bites: on a host where node is on the PATH the gates do enforce today, for main-agent and sub-agent calls alike, so this is not fixing a visible symptom everywhere. It matters on installs where the interpreter is not on the hook's PATH, and there it matters a lot.

Szotasz:
Köszönjük a PR-t, és elnézést, hogy ez a fenti észrevételekből kimaradt: a branch időközben konfliktusba került a `develop`-pal (`mergeable: CONFLICTING`). Az ok nem a te változtatásod, hanem egy azóta bement másik PR, amely ugyanazt a fájlt érinti (`src/web/agent-scaffold.ts`).

Kérünk egy rebase-t a friss `develop`-ra. A fenti véleményem tartalmilag változatlanul áll, a PR feje azóta nem mozdult, tehát a rebase után csak a konfliktus-feloldást kell átnéznünk, az egész PR-t nem mérjük újra.

VTolnai:
All three points folded in, and the branch is rebased onto current `develop` (the `agent-scaffold.ts` conflict was additive on both sides -- your new fetch-host/owner-domain helpers and this PR's `ensureGovernanceGateCommands` landed in the same spot; resolution keeps both, upstream first).

**What changed (head `71ee934`):**

1. **One comparison for the same job** -- `hookCommandWired()` is now the single wired-already predicate (the JSON-escaped compare you identified as the correct one), used by `ensureEgressGate` and `ensureGovernanceGateCommands` alike. The raw-includes form is gone.
2. **Quoted on both halves** -- `hookCommand()` is the single builder every injector and every ensure* uses: `"execPath" "scriptPath"`. Verified the quoted form still passes `isUnsafeHookCommand` (its script-path regex stops before quotes, so the existsSync check still fires on the right substring). One side effect worth naming: existing installs with the unquoted absolute-path entries will see exactly one more rewrite at next boot (the quoted form does not match the old entry), after which the check settles.
3. **Tests** (`src/__tests__/hook-command-quoting.test.ts`) -- injectors write a quoted absolute interpreter and never a bare `node`; the escaped compare settles on a Windows-style backslash path exactly where the raw compare does not (your point 1, pinned); and both ensure* migrations return true-then-false on a double call, including the legacy bare-`node` upgrade path.

**Your question about running sessions:** you were right to ask -- a running session does NOT re-read `settings.json`; hooks are captured at session start. The startup migration makes the upgrade zero-touch, not instantaneous: the rewritten command takes effect at each agent's next (re)spawn, which in this fleet happens routinely (watchdog restarts, nightly respawns), so the window is hours, not weeks. The docstring that claimed "without waiting for their next respawn" now says exactly this.

**Verified on the new head:** `tsc` clean, full vitest suite 227 files, 3131 passed / 1 skipped, zero failures.

PR #806, opened 2026-07-31
feat(costops): price token usage as a list-price equivalent (v0.2)

VTolnai:
CostOps **v0.2**: the token→cost mapping that v0.1 explicitly deferred. Read-only, no schema change, nothing switched. Kanban **#132 (F1)**.

## Why now

`ledger.ts` carried this note:

> `volume/activity only -- not priced in v0.1 (token_usage has no model column; token->cost mapping lands in v0.2 after model/session enrichment)`

That precondition now holds: `model` is populated on **9309/9309** rows, no NULLs anywhere in the table. So this implements the slot v0.1 designed rather than adding a parallel path.

## Two things measurement changed

**1. Cache tokens dominate.** Over the fleet's 7 days to 2026-07-31:

| component | share |
|---|---|
| `cache_read` | **82.9%** |
| `cache_write` | 9.5% |
| `output` | 7.6% |
| `input` | 0.0% |

The obvious `input + output` formula sees **7.6%** of the real figure — a ~13x understatement. The two cache multipliers are therefore named constants with the measurement recorded beside them, and a test asserts the ratio rather than just the arithmetic.

**2. An unknown model is a gap, not a zero.** `priceTokens()` returns `null` for an unrecognised id (e.g. the `<synthetic>` sentinel, 19 calls in the window) and the report counts those separately. Charging $0 would leave the total looking like a total while quietly missing that traffic.

## List-price equivalent, not money owed

The operator runs on a subscription, so **no invoice corresponds to these numbers**. Every output carries `basis: "list_price_equivalent"` and `currency: "USD"`. It sits beside the money ledger and is **never summed into `current_spend`** — the subscription already appears there as a fixed cost, so adding both double-counts. A test pins that separation; the money path is untouched apart from comments.

## Day bucketing

`dayKey()` takes an explicit IANA zone, defaulting to `Europe/Budapest`, so a daily ceiling follows the operator's wall clock instead of splitting a working evening at 01:00 local. This deliberately differs from the money ledger's UTC `monthWindow()`; both ends say so. The zone is a parameter so tests don't depend on the machine's `TZ`.

## What's here

- `src/costops/pricing.ts` (new) — versioned price map, `priceTokens()`, `dayKey()`
- `src/costops/ledger.ts` — `getTokenCostReport()`; summary gains a labelled `list_price_equivalent` block
- `src/web/routes/costs.ts` — `GET /api/costs/tokens?days=N` (rolling window, clamped 1–400)

## Verification

- **42 costops tests pass**; `tsc --noEmit` clean.
- **Controls run.** Zeroing the cache multipliers fails 4 tests; making an unknown model cost $0 instead of `null` fails 2. The tests discriminate.
- **Cross-checked against the live database.** This implementation and an independent Python computation agree to the cent per model: fable-5 $1158.64, opus-5 $561.69, sonnet-5 $51.84, sonnet-4-6 $10.55. (Totals differ by 0.2% only because the rolling window slid ~19 min between runs.)
- **Full suite green on a fresh upstream base: 3120 pass, 0 failures.**

> **Correction to an earlier revision of this description.** It reported 13 failures in `governance-gates` / `hook-path-guard` / `email-send-gate` as "pre-existing". They were not. They were caused by running the suite from a worktree under `/tmp`: `isUnsafeHookCommand` rejects `/tmp`-prefixed hook commands, so the gate tests fail there and only there. My control had reverted my diff while holding the location fixed, which proved my changes were not the cause but did not license the word "pre-existing". Re-run from `~/.cache`, all three files pass.

### DIFF
--- a/src/__tests__/costops-api.test.ts
+++ b/src/__tests__/costops-api.test.ts
@@ -41,14 +41,42 @@ describe('costops API (route smoke)', () => {
     expect(out.body).toHaveProperty('breakdown')
     expect(out.body).toHaveProperty('budget')
     expect(out.body).toHaveProperty('token_usage')
-    // token usage reported as VOLUME
+    // token usage reported as VOLUME, plus a labelled list-price equivalent
     expect(out.body.token_usage.input_tokens).toBe(1234)
     expect(out.body.token_usage.output_tokens).toBe(5678)
-    expect(out.body.token_usage.note).toContain('not priced')
+    expect(out.body.token_usage.note).toContain('LIST-PRICE EQUIVALENT')
+    expect(out.body.token_usage.list_price_equivalent.basis).toBe('list_price_equivalent')
+    expect(out.body.token_usage.list_price_equivalent.currency).toBe('USD')
     // money never derived from tokens (config amounts are 0 placeholders)
     expect(typeof out.body.current_spend).toBe('number')
     // no secret / account id leaks into the response
     expect(JSON.stringify(out.body)).not.toMatch(/secret|api[_-]?key|password|token=/i)
+  })
+
+  it('GET /api/costs/tokens prices a rolling window and labels its basis', async () => {
+    const now = Math.floor(Date.now() / 1000)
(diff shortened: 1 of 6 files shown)


Szotasz:

PR #807, opened 2026-07-31
feat(token-usage): attribute spend to the scheduled task that caused it

VTolnai:
**R0** for kanban #134, and the missing input for #132's routing decisions.

## The gap

`task_title` and `project` were **0% populated** across all 9309 rows. "What did the morning briefing cost last month" had no answer, so a model router could be neither trained nor evaluated — and no amount of later work recovers a day that went by unlabelled.

## The anchor

The schedule runner already wraps every injected prompt in a provenance wrapper carrying two markers. Measured across all 95 transcripts on 2026-07-31: **1149 scheduled prompts, 13 distinct tasks**, with the structured `<scheduled-task source="scheduled-task:X">` tag and the human-readable `[Heartbeat: X]` / `[Utemezett feladat: X]` prefix appearing **together on all 1149 and never apart**.

We anchor on the structured tag: a person can quote the bracket form in prose, but not the tag. A test pins that.

Neither marker sits at the head of the message — the safety preamble comes first. An `^`-anchored match finds nothing, which is how I mismeasured this once before concluding the markers were absent entirely.

## Three ways this would have been silently wrong

| Trap | Consequence if missed |
|---|---|
| **Tool results are `user` lines** (8270 of 10579) | Treating one as a person clears the label on a scheduled run's *first tool call* — every task appears to cost one API call |
| **An unmarked user turn must clear the label** | Otherwise one prompt's name bleeds onto every later turn in a long interactive session, wildly overstating that task |
| **Parsing resumes mid-file from the cursor** | The marker naming the current task usually sits in an already-consumed chunk; without carrying it, attribution goes quietly partial after the first resume |

The cursor gains `last_task_title` for the third. A full re-read (`fromLine 0`) starts clean, so a stale label cannot survive a deliberate rescan.

## Scope

Only the schedule **name** is stored. Its kind (`task` vs `heartbeat`) already lives in the schedule's own `task-config.json` and would go stale if duplicated per row. The upsert fills `task_title` when absent but **never overwrites**, so a backfill can label history without rewriting good rows. `project` is deliberately left alone rather than filled with an invented meaning.

## Verification

- **13 new tests + 23 existing token-usage tests pass**; `tsc --noEmit` clean.
- **Full suite green on a fresh upstream base: 3110 pass, 0 failures.** (An earlier revision of this description reported 13 failures as "pre-existing"; they were an artifact of running the suite from a worktree under `/tmp`, which `isUnsafeHookCommand` rejects. Re-run from `~/.cache`, everything passes.)
- **Verified end-to-end against the real transcripts** with an in-memory DB (the live database was never opened): **2412/9528 rows labelled across exactly the 13 distinct tasks** an independent count found. The unlabelled 75% is interactive work, which correctly has no task.

## What it already answers

With labelling in place, the previously unanswerable split falls out — scheduled work is **32.3%** of list-price-equivalent spend, interactive **67.7%**; and the three costliest scheduled tasks are all heartbeats (26.5% of total between them). That reframes where routing can and cannot help; detail on the #134 card.

### DIFF
--- a/src/__tests__/token-usage-task-labelling.test.ts
+++ b/src/__tests__/token-usage-task-labelling.test.ts
@@ -1,0 +1,118 @@
+import { describe, it, expect } from 'vitest'
+import { parseScheduledTaskMarker, userTurnText, collapseByMessageId } from '../web/token-usage.js'
+
+/**
+ * R0 (kanban #134): attribute token spend to the scheduled task that caused it.
+ *
+ * Nothing labelled the rows before this: task_title and project were 0%
+ * populated across all 9309 of them, so "what did the morning briefing cost
+ * last month" had no answer, and a model router could be neither trained nor
+ * evaluated.
+ *
+ * The label comes off the provenance wrapper the schedule runner puts around
+ * every injected prompt. Measured across all 95 transcripts on 2026-07-31:
+ * 1149 scheduled prompts, 13 distinct tasks, and the structured tag and the
+ * bracket prefix always appear together (1149/1149).
+ */
+
+// Shape of a real wrapped scheduled prompt: safety preamble, bracket prefix,
+// then the structured tag. The markers are mid-string, never at the head.
+function wrapped(kind: 'Heartbeat' | 'Utemezett feladat', name: string): string {
+  return (
+    'The wrapper marks provenance, not distrust. ' +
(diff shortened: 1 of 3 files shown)


Szotasz:

PR #808, opened 2026-07-31
feat(costops): eco mode -- one switch that moves the fleet to a cheap model

VTolnai:
> ⚠️ **Stacked on #806.** This branch carries **two** commits: `feat(costops): price token usage…` (the F1 commit, already open as #806) and the eco-mode commit below, which imports `costops/pricing.ts` from it. Merge **#806 first**; this PR then reduces to the single eco-mode commit. Flagged up front rather than left for the reviewer to discover.

**F2** for kanban #132.

## What it does and pointedly does not

Writes configuration. **Restarts nothing.** A model change only lands on the next agent restart, and that stays an operator decision — the main agent in particular cannot restart itself without dropping its own session and the channel with it. Every response carries `restart_required` and the list of agents still pending one.

## Three ways a naive "switch everything to X" would be wrong

**1. It upgrades the agents that are already cheap.** The fleet is mixed — `claude-fable-5` on the main agent, `claude-opus-5[1m]` on two, `claude-sonnet-5` on two — so moving everything to one model *raises* the bill for whatever sits below it. Agents at or under the target's rate are left alone and reported with a reason.

**2. It is a one-way door.** Originals are saved on enable. An absent `model` field is saved as `null`, not as today's `DEFAULT_MODEL`: restoring it must **delete** the field again, otherwise the agent is silently pinned to whatever the default happened to be that day and stops tracking future changes. Re-running enable while already on does not overwrite the saved originals.

**3. It writes model strings that do not exist.** `claude-opus-5[1m]` and `claude-opus-5` name the same priced model, but `[1m]` is not valid everywhere — haiku has no 1M variant. The target is therefore taken as a **literal** and validated against the price map rather than assembled by substitution. An unknown target is refused with a **400** instead of being written into a live config, where it would fail the agent's next launch. (The naming gotcha from the fleet model-config skill, enforced in code.)

## Safety of the writes

Config files are edited **key-wise** and written **atomically** via `rename(2)`. They carry hooks, plugins and per-agent settings this module knows nothing about — rewriting them from a typed shape would drop whatever it failed to model — and `.claude/settings.json` is live, so a half-written file there would break the next start.

## Default target

`claude-sonnet-5`, not haiku: it halves the fleet's most expensive model and cuts opus-5 by 40% while staying in the tier the paid work already runs on. Haiku is a bigger saving *and* a bigger capability drop, so it is an explicit choice (`target` in the request), not a default.

## API

| | |
|---|---|
| `GET /api/costs/eco` | current state + dry-run preview |
| `POST /api/costs/eco` | `{enabled, target?, dry_run?}` — applies config, never restarts |

## Verification

- **27 new tests**, `tsc --noEmit` clean, full suite **3147 pass, 0 failures** on a fresh upstream base.
- **Three sabotage controls run.** Removing the price-inversion guard fails 2 tests; always re-saving the originals fails 1; restoring an absent field as a model instead of `null` fails 2. (A fourth attempt initially "passed" — the sabotage had not matched due to indentation. Re-run correctly, it fails 2. A control that does not apply proves nothing.)
- The HTTP tests are deliberately non-mutating — GET, a rejected target, and an explicit dry run — so the suite never rewrites a tracked config to make its point. The write path is covered against temp files.

### DIFF
--- a/src/__tests__/costops-api.test.ts
+++ b/src/__tests__/costops-api.test.ts
@@ -41,14 +41,42 @@ describe('costops API (route smoke)', () => {
     expect(out.body).toHaveProperty('breakdown')
     expect(out.body).toHaveProperty('budget')
     expect(out.body).toHaveProperty('token_usage')
-    // token usage reported as VOLUME
+    // token usage reported as VOLUME, plus a labelled list-price equivalent
     expect(out.body.token_usage.input_tokens).toBe(1234)
     expect(out.body.token_usage.output_tokens).toBe(5678)
-    expect(out.body.token_usage.note).toContain('not priced')
+    expect(out.body.token_usage.note).toContain('LIST-PRICE EQUIVALENT')
+    expect(out.body.token_usage.list_price_equivalent.basis).toBe('list_price_equivalent')
+    expect(out.body.token_usage.list_price_equivalent.currency).toBe('USD')
     // money never derived from tokens (config amounts are 0 placeholders)
     expect(typeof out.body.current_spend).toBe('number')
     // no secret / account id leaks into the response
     expect(JSON.stringify(out.body)).not.toMatch(/secret|api[_-]?key|password|token=/i)
+  })
+
+  it('GET /api/costs/tokens prices a rolling window and labels its basis', async () => {
+    const now = Math.floor(Date.now() / 1000)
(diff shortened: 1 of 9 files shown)


Szotasz:

PR #809, opened 2026-07-31
feat(costops): attribute a task's cost to inherited context vs own work

VTolnai:
> ⚠️ **Stacked on #806 and #808.** Three commits: the F1 pricing commit (#806), the F2 eco-mode commit (#808), and the attribution commit below, which imports `costops/pricing.ts` and sits alongside eco-mode in `routes/costs.ts`. Merge **#806 → #808 → this**; it then reduces to a single commit.

**R1** for kanban #134 — **reframed**, with the reframing approved after the measurement below.

## The original premise did not survive measurement

R1 was to shadow a model router. But across the 13 marker-labelled tasks, the spread **within** one task is larger than the spread **between** tasks:

| | ratio |
|---|---|
| within one task (median) | **11x** |
| between tasks (medians) | 5.9x |
| `memoria-heartbeat` extremes | **$0.05 … $11.44** (249x, identical work) |

And the mechanism is visible: $11.44 at 730k `cache_read`/call, $0.05 at 47k. The runner injects into an **already-running session**, so a task pays to re-read whatever context that session had accumulated. What a task costs is mostly a fact about its host — a complexity heuristic built on observed cost would be measuring session state.

## What this does instead

Decomposes each task's spend into **inherited context** and **own work**, and prices running it in a fresh session. Against live data:

```
inherited context   $752.54   (86%)
own work            $124.93   (14%)
total               $877.47
fresh-session       $564.67  ->  saving $312.80 (36%)
```

## The fresh-session comparison is priced honestly

This is the part that decides whether any of it is trustworthy. A fresh session is **not free**: it writes its own preamble to cache once (1.25x) and re-reads it on every later call (0.1x). The preamble is **measured, not assumed** — 41,224 tokens, the median first-call context across the 40 sessions in `token_usage` (quartiles 36.3k / 43.5k).

With it priced in the saving is **36%**, not the 86% a naive comparison would claim. Two tasks come out **worse** fresh (`heti-osszefoglalo` at −4%): a frequent short heartbeat re-pays startup on every run. A model that always favoured fresh sessions would be rigged, so a test pins that it can lose.

## The report carries the evidence for its own caveat

`complexity_signal` **computes** the within/between variance rather than asserting it, and derives its verdict from the comparison — so the negative result about cost-based complexity heuristics is reproducible from the output, not a claim in a commit message.

## Scope

Restricted to the marker-labelled schedule names. That filter is **not optional**: `task_title` also holds kanban card titles written by `correlateWithKanban()`, which are a time-window guess rather than a statement of what a run was doing. Mixing them would put unreliable attribution into a cost figure.

Read-only — computes, stores nothing, changes nothing. `GET /api/costs/context-attribution?days=N`.

## Verification

- 18 new tests, `tsc --noEmit` clean, full suite **3165 pass, 0 failures**.
- **Three sabotage controls, each asserted to have applied before running** (a control that silently fails to apply proves nothing — that happened to me once and is now guarded): making the fresh session free fails 3 tests; folding inherited context into own work fails 3; bucketing runs by clock hour instead of a time gap fails 2.

### DIFF
--- a/src/__tests__/costops-api.test.ts
+++ b/src/__tests__/costops-api.test.ts
@@ -41,14 +41,42 @@ describe('costops API (route smoke)', () => {
     expect(out.body).toHaveProperty('breakdown')
     expect(out.body).toHaveProperty('budget')
     expect(out.body).toHaveProperty('token_usage')
-    // token usage reported as VOLUME
+    // token usage reported as VOLUME, plus a labelled list-price equivalent
     expect(out.body.token_usage.input_tokens).toBe(1234)
     expect(out.body.token_usage.output_tokens).toBe(5678)
-    expect(out.body.token_usage.note).toContain('not priced')
+    expect(out.body.token_usage.note).toContain('LIST-PRICE EQUIVALENT')
+    expect(out.body.token_usage.list_price_equivalent.basis).toBe('list_price_equivalent')
+    expect(out.body.token_usage.list_price_equivalent.currency).toBe('USD')
     // money never derived from tokens (config amounts are 0 placeholders)
     expect(typeof out.body.current_spend).toBe('number')
     // no secret / account id leaks into the response
     expect(JSON.stringify(out.body)).not.toMatch(/secret|api[_-]?key|password|token=/i)
+  })
+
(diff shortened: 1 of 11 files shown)


Szotasz:

PR #810, opened 2026-07-31
feat(token-usage): record which writer set task_title

VTolnai:
> ⚠️ **Stacked on #807.** Two commits: the R0 labelling commit (#807) and the provenance commit below, which stamps the column R0's parser writes. Merge **#807 first**; this then reduces to a single commit.

## The problem

`task_title` is written by two things with very different reliability:

| writer | how it attributes | trustworthy? |
|---|---|---|
| transcript parser (R0) | reads a marker the runner puts **into the prompt** | exact — the prompt names the task |
| `correlateWithKanban()` | **time window** — every row in a span gets whatever card moved during it | a guess |

In one column they are indistinguishable, so a reader cannot tell an exact attribution from a heuristic one. That is how a guess ends up inside a cost report presented as fact — and it is not hypothetical: after the recent backfill the live table holds 2 418 marker rows and 6 855 correlated ones, side by side.

Until now the only way to separate them was to check whether a title happened to match a directory name under `scheduled-tasks/`. That misfiles any card sharing a name with a schedule, and breaks silently when either side is renamed.

## What this adds

`task_source` records which writer set the label: `'schedule_marker'` or `'kanban_correlation'`.

- The parser's upsert fills it **on a rescan** for rows labelled before the column existed, and **never relabels** a row that already has a source.
- `NULL` means the provenance predates the column — stated rather than guessed.
- No behaviour change to either writer beyond the stamp.

## Verification

- 5 new tests, `tsc --noEmit` clean, full suite **2947 pass, 0 failures**.
- **Sabotage control:** dropping the stamp from the correlation `UPDATE` fails 3 of the 5.
- One of the tests pins that `correlateWithKanban()` leaves a marker-labelled row alone — its `task_title IS NULL` guard is what protects exact labels from being overwritten by a time-window guess, and nothing was covering it.

## A test-setup note worth keeping

`initDatabase(':memory:')` does not reliably return an empty database when called repeatedly in one process, so these cases clear both tables explicitly. Without that they passed alone and failed together, which reads as flakiness rather than a setup bug.

### DIFF
--- a/src/__tests__/token-usage-task-labelling.test.ts
+++ b/src/__tests__/token-usage-task-labelling.test.ts
@@ -1,0 +1,118 @@
+import { describe, it, expect } from 'vitest'
+import { parseScheduledTaskMarker, userTurnText, collapseByMessageId } from '../web/token-usage.js'
+
+/**
+ * R0 (kanban #134): attribute token spend to the scheduled task that caused it.
+ *
+ * Nothing labelled the rows before this: task_title and project were 0%
+ * populated across all 9309 of them, so "what did the morning briefing cost
+ * last month" had no answer, and a model router could be neither trained nor
+ * evaluated.
+ *
+ * The label comes off the provenance wrapper the schedule runner puts around
+ * every injected prompt. Measured across all 95 transcripts on 2026-07-31:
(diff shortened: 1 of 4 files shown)


Szotasz:

PR #811, opened 2026-07-31
feat(costops): subscription quota guard -- warn at 70%, eco mode at 85%

VTolnai:
> ⚠️ **Stacked on #806, #808 and #809.** Four commits: F1 pricing (#806), F2 eco mode (#808), R1-A attribution (#809), and the guard below. Merge **#806 → #808 → #809 → this**; it then reduces to a single commit.

**F3** for kanban #132, spec from #135.

## Why the quota, not a dollar cap

The fleet runs on a subscription, so there is no invoice to cap and CostOps' list-price figures are a yardstick, not money owed. What actually runs out is the rolling quota — and it is readable: the OAuth usage endpoint reports consumption of the five-hour and seven-day windows.

## The trigger is an absolute ladder

| seven-day utilization | action |
|---|---|
| ≥ 70% | warning message to the main agent |
| ≥ 85% | eco mode switched on + alert |

It is deliberately **not** "consumption is ahead of elapsed time". That rule fires almost always at the *start* of a window and almost never at the *end*: one hour into seven days, elapsed is **0.6%**, so any real work is already far ahead. No margin fixes it — the asymmetry is structural. The pace comparison survives as **forecast text** in the alert, where being early is informative rather than actionable. A test pins that 15% used at 0.6% elapsed stays quiet.

## Secret handling

The token is the sharpest edge here. It is read from `~/.claude/.credentials.json` and **never** logged, never put in an error, never returned in a result, never in the alert text.

The catch blocks deliberately **discard the error object**: a JSON parse error quotes the offending region of the file — which is the token — and a fetch error can carry the request including its `Authorization` header. Two tests assert the token reaches no log line.

## Failure is treated as failure

The endpoint is undocumented, so an unreadable or unrecognised payload yields `null`, **never a plausible 0%**. A guard that silently read 0% would run the fleet straight through a cap while looking like it had checked. Failures are **silent to users and loud in the log** — the opposite arrangement is exactly the silent no-op this subsystem exists to catch.

## Not shouting

A steady 72% must not speak every cycle (96 times a day at this cadence — noise is how a real alert gets ignored). Alerts fire on a **level change** or after a **six-hour cooldown**, and the latch clears on recovery. Alerts go to the main agent as an inter-agent message; the guard has no channel of its own and should not grow one.

## Config only

It never restarts an agent, and every result says so. The read-only `GET /api/costs/quota` classifies **without acting**, so a dashboard refresh cannot trip the fleet into eco mode. Cadence is 20 minutes, inside the agreed 15–30 range.

## Verification

- 29 new tests, `tsc --noEmit` clean, full suite **2994 pass, 0 failures**.
- **Three sabotage controls, each asserted to have applied**: logging the token fails 1 test; reading an unknown payload as 0% fails 1; making the pace comparison a trigger fails 3.
- That third control initially caught only **2** — the early-window test used a value that happened to sit under the sabotage's margin. It was strengthened to 15% at 0.6% elapsed (25x ahead, unambiguous), and now the control catches it. A control that passes by luck is not a control.

### DIFF
--- a/src/__tests__/costops-api.test.ts
+++ b/src/__tests__/costops-api.test.ts
@@ -41,14 +41,42 @@ describe('costops API (route smoke)', () => {
     expect(out.body).toHaveProperty('breakdown')
     expect(out.body).toHaveProperty('budget')
     expect(out.body).toHaveProperty('token_usage')
-    // token usage reported as VOLUME
+    // token usage reported as VOLUME, plus a labelled list-price equivalent
     expect(out.body.token_usage.input_tokens).toBe(1234)
     expect(out.body.token_usage.output_tokens).toBe(5678)
-    expect(out.body.token_usage.note).toContain('not priced')
+    expect(out.body.token_usage.note).toContain('LIST-PRICE EQUIVALENT')
+    expect(out.body.token_usage.list_price_equivalent.basis).toBe('list_price_equivalent')
+    expect(out.body.token_usage.list_price_equivalent.currency).toBe('USD')
     // money never derived from tokens (config amounts are 0 placeholders)
     expect(typeof out.body.current_spend).toBe('number')
     // no secret / account id leaks into the response
     expect(JSON.stringify(out.body)).not.toMatch(/secret|api[_-]?key|password|token=/i)
+  })
+
+  it('GET /api/costs/tokens prices a rolling window and labels its basis', async () => {
+    const now = Math.floor(Date.now() / 1000)
+    getDb().prepare("INSERT INTO token_usage (agent,session_id,timestamp,input_tokens,output_tokens,cache_read_tokens,cache_creation_tokens,model) VALUES ('marveen','s',?,0,1000000,0,0,'claude-opus-5')").run(now - 3600)
+    const { ctx, out } = fakeCtx('/api/costs/tokens?days=7')
+    expect(await tryHandleCosts(ctx)).toBe(true)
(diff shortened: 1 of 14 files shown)


Szotasz:

Szotasz:
### Secret-scan gate: one fixture line needs a different placeholder

Heads-up before this can merge: our required `scan` check flags one added line in `src/__tests__/costops-quota-guard.test.ts`, the `const TOKEN = ...` fixture. The detector matches on the `sk-ant-oat01-` prefix shape, so it fires on the shape alone, not on anything it knows about your token.

**To be explicit: this is not a security finding.** The value is obviously synthetic and nothing leaked. We read the gate output before writing this.

**Why we are asking you to change the line instead of allowlisting the file.** The gate output itself suggests adding the path to `ALLOWLISTED_PATHS`. We deliberately went the other way. A path allowlist is permanent: it would also wave through a *real* secret that lands in that same file six months from now, and nobody would see it. The placeholder rewrite costs one line and keeps the file protected. The general form we work by: you pay for a gate exception on every future case, not on this one.

**What we need:** replace the prefix so the value cannot match a live key shape. Anything like `dummy-oat01-test-token-value` or `not-a-real-token` works. The test only needs a stable string.

Once the line changes and `scan` goes green, this is ready from our side. Thanks for the contribution, and sorry that the gate found it the noisy way.

PR #812, opened 2026-07-31
feat(costops): watch the eco-worker context ceiling

VTolnai:
> ⚠️ **Stacked on #806 and #808.** Three commits: F1 pricing (#806), F2 eco mode (#808), and the watch below. Merge **#806 → #808 → this**; it then reduces to a single commit.

The eco-worker's first guard rail. Kanban #132 / #139.

## Why it exists

The eco-worker is only cheaper **while its context stays small**. From the R1-A model it beats the status quo until roughly **433k tokens**; past that the arithmetic inverts and running the same task inside the main session costs the same. The main session averaged **~461k** at the measured runs — so a neglected worker really can get there.

That makes context hygiene a requirement, and *a requirement nobody checks is a wish*. This watches it and warns at **250k** — early enough to leave room to act, rather than at the point where the worker has already stopped being worth having.

## Three deliberate choices

**It is an allow-list, not every agent.** The main agent carries a large context by design; alerting on that would be a daily false alarm, and a channel that cries wolf is how a real warning gets ignored.

**An agent with no recent rows reads as `null`, not `0`.** A worker that never started would otherwise be reported as a comfortable 0k and look healthy. It is surfaced as `not_seen` instead.

**It only reports.** Compaction is the worker's own job; restarting is an operator decision. A watch that quietly acted would be the worst of both — and the alert text says so explicitly.

## Verification

- 15 tests, `tsc --noEmit` clean, full suite **3009 pass, 0 failures**.
- **Three sabotage controls, each asserted to have applied**: reading an unseen agent as `0` fails 4 tests; watching the main agent fails 1; alerting only at the break-even instead of the ceiling fails 4.

### DIFF
--- a/src/__tests__/costops-api.test.ts
+++ b/src/__tests__/costops-api.test.ts
@@ -41,14 +41,42 @@ describe('costops API (route smoke)', () => {
     expect(out.body).toHaveProperty('breakdown')
     expect(out.body).toHaveProperty('budget')
     expect(out.body).toHaveProperty('token_usage')
-    // token usage reported as VOLUME
+    // token usage reported as VOLUME, plus a labelled list-price equivalent
     expect(out.body.token_usage.input_tokens).toBe(1234)
     expect(out.body.token_usage.output_tokens).toBe(5678)
-    expect(out.body.token_usage.note).toContain('not priced')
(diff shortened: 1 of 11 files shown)


Szotasz:

PR #814, opened 2026-07-31
feat(costops): shadow-evaluate the first eco-worker run

VTolnai:
> ✅ **Not stacked** — a single commit on current upstream. It depends on no other CostOps module.

Shadow evaluation for the first eco-worker task. Kanban #134 (R2) / #139.

## What it measures, and why not cost

Moving `reggeli-teteles-lista` to the eco-worker is a **methodology proof, not a saving**: the task is 0.7% of fleet spend. What it has to prove is that a cheap, small-context worker still produces a **correct** report — and that is checkable, because the task reads the kanban database and the database is still there to check against.

**Cost alone would be a trap.** A worker that reports nothing is extremely cheap, and on a cost dashboard it looks like a triumph. So the metric is coverage: of the cards that actually moved, how many did the report name.

## Three judgements made deliberately

| case | verdict | why not the obvious thing |
|---|---|---|
| worker sent nothing | `no_output` | This task's own skill **forbids silence** — it must print a line even for a thread with no activity. Folding it into a recall of 0 would hide that it has a different cause. |
| nothing moved in the window | `no_data` | Calling that a pass would let a broken worker accumulate green days. |
| everything named | `pass`, with a caveat | See below. |

## The metric is generous, and says so

Title substring matching can credit a report for a sentence that was not about that card, which inflates recall. **That asymmetry is chosen:** a low recall is solid evidence the worker omitted things; a high one is weak evidence it did well. The failure worth catching is the quiet omission, so the metric errs towards saying the worker did fine — and the note in the output says exactly that, so nobody reads a green number as proof.

Messages are read from the **transcript**, not a log, so what is evaluated is what was actually sent. A line that merely mentions `notify.sh` in prose is not a send, and a test pins that.

## Honest gap

**Not yet validated against a real run** — the first eco-worker execution is tomorrow at 07:45. Extraction is covered with synthetic transcripts until then, and closing that gap is the point of running it.

## Verification

- 17 tests, `tsc --noEmit` clean, full suite **3032 pass, 0 failures**.
- **Three sabotage controls, each asserted to have applied**: treating silence as a pass fails 1; dropping comments from ground truth fails 1; calling an empty window a pass fails 1.

### DIFF
--- a/src/__tests__/costops-shadow-eval.test.ts
+++ b/src/__tests__/costops-shadow-eval.test.ts
@@ -1,0 +1,164 @@
+import { describe, it, expect, beforeEach } from 'vitest'
+import { initDatabase, getDb } from '../db.js'
+import {
+  groundTruthCards,
+  extractNotifyMessages,
+  evaluateCoverage,
+  evaluateRun,
+} from '../costops/shadow-eval.js'
+
+/**
+ * Shadow evaluation of the first eco-worker task.
+ *
+ * The move is a methodology proof, not a saving: reggeli-teteles-lista is 0.7%
+ * of fleet spend. What has to be proved is that a cheap, small-context worker
+ * still produces a CORRECT report -- which is checkable, because the task reads
+ * the kanban database and the database is still there to check against.
+ *
+ * Cost alone would be a trap: a worker that reports nothing is very cheap.
+ */
+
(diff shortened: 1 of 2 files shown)


Szotasz:

PR #818, opened 2026-07-31
feat(kanban): tell the owner when an agent comments on their card

VTolnai:
> ✅ **Not stacked** — a single commit on current upstream.

The **notification half** of kanban #45. The auto-claim half is not here — see below.

## Why

A card assigned to the owner is one **they** are expected to act on. An agent answering there is usually a question or a handback, and it landed silently: the comment appeared on a board nobody was watching, and the thread stalled waiting for a reply the owner never knew was expected.

## Choices worth stating

**Owner identified from `OWNER_NAME`, not a literal name** — stays correct in an install configured for someone else. Matching is case-insensitive on both sides, because the board already holds both `marveen` and `Marveen`: assignee casing drifts in practice.

**No cooldown, deliberately.** Measured before choosing: over the seven days to 2026-07-31 there were **68 such comments across 24 cards** — about 10/day, 55 of them from the main agent. The point of the card is that these must not be missed; under-notifying defeats it, while over-notifying is merely annoying and easy to tune once someone has seen the real volume. Happy to add a per-card cooldown if 10/day proves too much — that is a decision for whoever receives them.

**The Telegram config check gates the default sender only.** My first version checked it unconditionally, which meant an injected sender was refused whenever the global config happened to be empty — untestable, and it would have silently disabled a working alternative transport. The guard belongs to the path that needs it.

## What is NOT in here

The auto-claim fix. **I could not find the mechanism and am not guessing.** Searched: every `assignee` write in `src/` (only `updateKanbanCard`, which takes a caller-supplied value), `addKanbanComment` (touches `updated_at` only), the frontend (`web/app.js` reads assignee for display, never writes), `kanban_card_events` (records status only — no assignee audit trail), skills and scheduled tasks (none instruct setting assignee to self), and the plugins directory (no kanban writes). Details on the card.

## Verification

- 9 tests, `tsc --noEmit` clean, full suite **3051 pass, 0 failures**.
- **Two sabotage controls**: dropping the author check fails 2 tests; dropping the owner-card check fails 2.

### DIFF
--- a/src/__tests__/kanban-owner-comment-nudge.test.ts
+++ b/src/__tests__/kanban-owner-comment-nudge.test.ts
@@ -1,0 +1,89 @@
+import { describe, it, expect, vi } from 'vitest'
+import { notifyOwnerOfAgentComment } from '../web/routes/kanban.js'
+import { OWNER_NAME } from '../config.js'
+
+/**
+ * A card assigned to the owner is one THEY are expected to act on. An agent
+ * answering there is usually a question or a handback, and until now it landed
+ * silently: the comment appeared on a board nobody was watching, and the thread
+ * stalled waiting for a reply the owner never knew was expected.
+ *
+ * Measured over the seven days to 2026-07-31: 68 such comments across 24 cards,
+ * about 10 a day. Volume worth knowing before deciding whether this needs a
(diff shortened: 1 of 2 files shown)


Szotasz:

PR #838, opened 2026-08-01
fix(agents): an MCP server added later never reached existing isolated config dirs

### LINKED ISSUE #834: Isolated agent config dirs seed `mcpServers` only once — servers added later to the shared ~/.claude.json never reach existing agents
### Problem
When an MCP server is added to the shared user config (`~/.claude.json` top-level `mcpServers`), only agents spawned *after* that point pick it up. Agents whose isolated `CLAUDE_CONFIG_DIR` was already provisioned keep their original, now-stale server set forever — even across restarts.

### Root cause
`provisionIsolatedConfigDir` (`src/web/agent-process.ts`) seeds the isolated `<config-dir>/.claude.json` from a full copy of `~/.claude.json` **only when the file is absent** (first provision). On every later spawn/restart it touches that file solely to guarantee `hasCompletedOnboarding`; the `mcpServers` block is left frozen at its first-seed state. So the seed is a one-time snapshot, not an ongoing reconcile.

Concretely: adding a new server to `~/.claude.json` propagates to brand-new agents (a fresh seed copies the whole file) but not to any already-provisioned isolated config dir.

### Impact
Rolling out a new MCP server to an existing fleet requires a manual per-agent backfill of every isolated `.claude.json`. Miss one and that agent silently lacks the server with no error — it just never appears in its tool list.

### Proposed fix
On every provision (not just first seed), **additively reconcile** the isolated `.claude.json` top-level `mcpServers` against the shared `~/.claude.json`:

- Add any server present in the shared config but missing in the isolated one.
- **Never remove or overwrite** an existing entry — preserve Claude Code's evolved state and any deliberate per-agent scoping (e.g. a channel agent's own `.mcp.json` / plugin wiring).
- Keep the write atomic (tmp + rename) so a running Claude Code process never sees a torn file.

This makes the isolated dirs self-healing for newly added servers while keeping the current "leave Claude Code's owned state alone" guarantee intact — the reconcile only fills gaps.

### Notes / scope
- Additive-only by design — a gap-fill, not a two-way sync; servers removed from the shared config are intentionally left in place.
- Auth is out of scope: MCP credential resolution is `$HOME`-based and already shared across same-user agents; this change only concerns server *registration*.
- Downstream context: this surfaced when distributing a google-drive MCP server across a running fleet — a one-time backfill worked but only fixes the current set, not the next server added.

Szotasz:
Closes #834.

`provisionIsolatedConfigDir` seeded the isolated `.claude.json` from a full copy of the shared `~/.claude.json` **only when the file was absent**. On every later spawn it touched that file solely to keep `hasCompletedOnboarding` set, so `mcpServers` stayed frozen at its first-seed snapshot. A server added to `~/.claude.json` afterwards reached brand-new agents and no existing one, silently: a missing server produces no error, the tool simply never appears in that agent's list.

This is not theoretical for us. It is what happened when the google-drive MCP server was rolled out across the running fleet: a manual per-agent backfill worked, but only for the set as it stood that day.

## Change

Provision now gap-fills `mcpServers` additively, and the writes go through temp file + `rename`.

Deliberately one-way:

- a server missing from the isolated file is copied in;
- an entry that already exists is **never** overwritten, so Claude Code's evolved state and any per-agent scoping decision survive;
- a server removed from the shared config is left in place;
- a non-object `mcpServers` on either side is not touched at all, because we cannot merge what we do not understand.

## Measurements

Everything below runs in a `mkdtemp` sandbox with `homedir()` and `agentDir()` redirected into it. No live fleet agent and no real `~/.claude` is read or written by the tests.

**Positive control first — the bug is real on the current code.** The new test file was run against the *unfixed* `agent-process.ts` before measuring the fix. Exactly the three propagation assertions fail:

```
× propagates a server added to the shared config AFTER the dir was provisioned
× never overwrites an entry that already exists in the isolated config
× preserves unrelated keys and keeps hasCompletedOnboarding set
Tests  3 failed | 3 passed (6)
```

Worth stating plainly: the other three still pass on the unfixed code, and that is not a contradiction. The old code wrote nothing, so "additive only" and "does not touch a corrupt value" hold trivially. They only become evidence of an implemented guarantee when they pass *together with* propagation, which is the state after the fix.

**The no-overwrite guarantee is shown, not asserted from the code.** The test hand-writes an evolved `gmail` entry (`env: { SCOPE: 'agent-local' }`) into the target file, puts a different `gmail` definition plus a genuinely new server in the shared config, re-provisions, and reads the target back: the evolved entry is byte-identical afterwards and only `google-drive` was added.

**Everything else:**

- New tests: 6 passed on the fixed code.
- Adjacent existing suites (`isolated-channel-config`, `main-agent-isolated-config`, `main-agent-config-dir`, `claude-config-dir`): 72 passed.
- Full suite: 229 files, 3163 tests, all passing.
- `tsc --noEmit`: clean.
- Em/en dash on added lines: 0, with a positive control confirming the scanner detects one when present.
- Atomicity is covered by an assertion that no `.tmp-*` staging file survives a provision.

## Not covered

The tests exercise the reconcile through `ensureIsolatedChannelConfigDir`, the real entry point, but they do not launch an actual Claude Code process. That a newly propagated server then appears in that agent's tool list is inherited from how the config file is consumed and is not measured here.

### DIFF
--- a/src/__tests__/isolated-config-mcp-reconcile.test.ts
+++ b/src/__tests__/isolated-config-mcp-reconcile.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
-import { mkdtempSync, mkdirSync, writeFileSync, rmSync, readFileSync, readdirSync } from 'node:fs'
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, readFileSync, readdirSync, chmodSync, statSync } from 'node:fs'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
 
@@ -116,6 +116,39 @@ describe('isolated config dir: mcpServers reconcile', () => {
     expect(after.projects).toEqual({ '/some/path': { hasTrustDialogAccepted: true } })
     expect(after.hasCompletedOnboarding).toBe(true)
     expect(Object.keys(after.mcpServers as object).sort()).toEqual(['extra', 'gmail'])
+  })
+
+  it('keeps an 0600 config at 0600 across a reconcile', () => {
+    // tmp + rename replaces the inode, so without an explicit mode the new file
+    // takes the umask default and a 0600 config silently relaxes to 0644. The
+    // mcpServers entries carry env blocks with credentials, and this reconcile
+    // is the path that rewrites the file regularly.
+    ensureIsolatedChannelConfigDir(AGENT, 'telegram')
+    chmodSync(isolatedDotClaude(), 0o600)
+    expect(statSync(isolatedDotClaude()).mode & 0o777).toBe(0o600)
+
+    writeShared({ gmail: { command: 'npx', args: ['gmail-mcp'] }, 'google-drive': { command: 'npx' } })
+    ensureIsolatedChannelConfigDir(AGENT, 'telegram')
+
+    // The reconcile really did run (otherwise the mode assertion proves nothing).
+    expect(Object.keys(servers()).sort()).toEqual(['gmail', 'google-drive'])
+    expect(statSync(isolatedDotClaude()).mode & 0o777).toBe(0o600)
+  })
+
+  it('preserves a deliberately wider mode too, rather than forcing 0600', () => {
+    ensureIsolatedChannelConfigDir(AGENT, 'telegram')
+    chmodSync(isolatedDotClaude(), 0o644)
+    writeShared({ gmail: { command: 'npx', args: ['gmail-mcp'] }, extra: { command: 'x' } })
+    ensureIsolatedChannelConfigDir(AGENT, 'telegram')
+    expect(Object.keys(servers()).sort()).toEqual(['extra', 'gmail'])
+    expect(statSync(isolatedDotClaude()).mode & 0o777).toBe(0o644)
+  })
+
+  it('creates a brand new isolated config owner-only, not at the umask default', () => {
+    // First provision: no target existed, so the fallback decides. It must be
+    // 0600 rather than whatever the umask happens to be on the host.
+    ensureIsolatedChannelConfigDir(AGENT, 'telegram')
+    expect(statSync(isolatedDotClaude()).mode & 0o777).toBe(0o600)
   })
 
   it('leaves no staging file behind (atomic write)', () => {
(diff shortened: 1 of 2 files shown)


Szotasz:
A fix maga jó, és a két kért mérés is az, amit kértem: a pozitív kontroll megelőzi a hiány-állítást, és a no-overwrite garanciát read-back bizonyítja, nem a kód olvasata. Az önkorrekciód arról, hogy a három átmenő assertion önmagában miért nem bizonyíték, pontos.

**Egy blokkoló nit, mielőtt mergelem.** A `writeJsonAtomic` a tmp fájlt alapértelmezett móddal hozza létre, a `renameSync` pedig a cél fájlt lecseréli erre az új inode-ra. A korábbi `writeFileSync(dotClaude, ...)` a MEGLÉVŐ fájlba írt, tehát megtartotta annak a jogosultságát. Így a váltás csendben 600-ról 644-re lazítja a fájlt.

Nem elméleti, lemértem:

```
$ chmod 600 target.json
$ node -e "fs.writeFileSync('target.json.tmp-1','{}'); fs.renameSync('target.json.tmp-1','target.json')"
$ ls -l target.json
-rw-r--r--  1 marvin  wheel  ...
```

És a fájl tartalma indokolja, hogy ez számítson: a saját hostunkon az izolált `.claude.json`-ok egy része `-rw-------`, és az `mcpServers` bejegyzések `env` blokkot hordoznak (nálunk pl. `google-drive:[GOOGLE_DRIVE_OAUTH_CREDENTIALS]`, máshol a flottában sima API kulcsok). A reconcile pedig pont az az út, ami mostantól rendszeresen újraírja ezt a fájlt.

Kérés: a tmp írása kapjon módot. Precedens már van a repóban, a `scripts/channels.sh` re-seed blokkja `{ mode: 0o600 }`-zal ír. A robusztusabb alak, ha van: a meglévő cél módját vidd tovább (`statSync(path).mode`), és csak ha a cél nem létezik, essen vissza 0o600-ra.

Egy tesztet is kérek rá, mert ez pont az a fajta regresszió, amit senki nem vesz észre: 600-as cél fájl, reconcile, majd read-back a módra.

A többi rendben, mehet utána.

PR #839, opened 2026-08-01
fix(install): every default install was wired to the rotating shared credential

Szotasz:
Addresses #835.

## Merged

The owner gave the GO ("mehet alapbol") for new installs to get this by default. Merged after the verify below. Follow-up #841 corrects a comment this change made false.

## What was wrong

`MAIN_AGENT_ISOLATED_CONFIG` has existed for a while and nothing ever set it. Both installers mention the key exactly once each, in a comment, as an example of an "operator-added key" (`install-macos.sh:567`, `install-linux.sh:898`). On a default install the main agent therefore kept the shared `~/.claude` and authenticated from whatever refreshes that root: the rotating macOS Keychain session, or the shared `.credentials.json` on Linux. Both expire. When they do the agent parks on a 401 TUI that the router reads as busy, so the channel goes silent with no error anywhere. `scripts/main-agent-isolated-config.mjs` and `src/web/agent-process.ts:305` already name this as the confirmed root cause of the 2026-07-23 outage.

The reporter titled this as macOS, but the gap is in both installers. On Linux the rotating source is the shared `.credentials.json`, which the same header documents as expiring.

Not a theoretical fix: this host has run in isolated mode since 2026-07-22 (`.env` + `config-overrides.json`), and the main agent has been in that mode for the last ten days.

## What changed

**1. The installers propose the setting, once, in the right branch.** `env_set_if_absent MAIN_AGENT_ISOLATED_CONFIG 1` is called only inside the branch where the installer just wrote the fleet setup-token itself in this same run. Then the identity the main bot runs under is the one the operator handed over moments earlier. An install that already carried auth (the "already carries an auth key" path) never reaches this branch and keeps whatever it had.

**2. A new `env_set_if_absent` helper**, identical text in both installers: it writes only when the key line is absent, so a deliberate `MAIN_AGENT_ISOLATED_CONFIG=0` survives a re-run and an existing `=1` is not duplicated.

**3. The start-time guard gets a second trigger.** The existing one requires a `.channels-config` dir, which only exists after an isolated boot has already happened — so on an install that never ran isolated it was structurally silent, which is exactly the install shape this issue is about. The new trigger fires when a fleet token exists but the resolution came back empty; that combination is suspicious on its own, because the token is what isolation is gated on. Installs with no fleet token stay quiet, so existing default setups get no new noise.

## Measurements

**`env_set_if_absent`, tested on the shipped text.** The function body was extracted from `install-macos.sh` with `sed` and sourced, rather than retyped into the test — the first version of this check re-declared the function by hand, which would have measured my transcription instead of the file.

```
A absent      -> MAIN_AGENT_ISOLATED_CONFIG=1
B explicit 0  -> MAIN_AGENT_ISOLATED_CONFIG=0   (deliberate 0 stands)
C already 1   -> 1 line, unrelated keys kept    (no duplicate)
```

The `env_set_if_absent` block is byte-identical in both installers (`diff` of the two extracted bodies is empty).

**Guard condition truth table**, and the condition string used in the test was byte-compared against the shipped line at `scripts/channels.sh:393` before drawing conclusions:

```
cfgdir=no  token=yes -> FIRES     (fresh install, setting missing: the case #835 is about)
cfgdir=no  token=no  -> silent    (no fleet token: no new noise)
cfgdir=yes token=yes -> silent    (existing guard's territory)
cfgdir=yes token=no  -> silent
```

**Everything else:** `bash -n` clean on all three files; full suite 228 files / 3157 tests passing; em/en dash on added lines 0, with a positive control confirming the scanner detects one when present.

## Not measured

No end-to-end fresh install was run on a clean machine. The behaviour is verified at the level of the helper, the branch it sits in, and the guard condition — not by watching a real installer run write the key and a main agent then boot isolated. If that is wanted before merge, it needs a throwaway host.

### DIFF
--- a/install-linux.sh
+++ b/install-linux.sh
@@ -914,6 +914,14 @@ env_keep_or_set() {
   if [ -z "$2" ] && [ -n "$_eks_existing" ]; then return 0; fi
   env_merge_key "$1" "$2"
 }
+env_set_if_absent() {
+  # env_set_if_absent KEY VALUE -- write only when the KEY line does not exist
+  # at all. Used for a default the installer proposes rather than enforces: an
+  # operator who deliberately set KEY=0 keeps that decision across re-runs, and
+  # an explicitly set KEY=1 is not rewritten either.
+  if grep -q "^$1=" "$INSTALL_DIR/.env" 2>/dev/null; then return 0; fi
+  env_merge_key "$1" "$2"
+}
 (umask 077 && touch "$INSTALL_DIR/.env")
 chmod 600 "$INSTALL_DIR/.env"
 [ -s "$INSTALL_DIR/.env" ] || printf '# Main agent konfiguracio\n' >> "$INSTALL_DIR/.env"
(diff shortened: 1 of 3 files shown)


Szotasz:
A GO megvan a tulajdonostól, „mehet alapból", tehát a leírás tetején lévő „Do not merge yet" blokk elveszti az alapját. A GO hatóköre viszont pontosan az, ami itt épült: ÚJ telepítés alapértelmezése. Már működő telepítéseket frissítéskor nem állítunk át, és ez a PR nem is teszi, jól.

A három rész maga rendben van. A `env_set_if_absent` mindkét telepítőben a definíció után, a `.env` létrehozása után hívódik, és csak abban az ágban, ahol a telepítő maga tárolta el a tokent. A két guard-feltétel kizárja egymást (`! -d .channels-config` vs `-d .channels-config`), tehát nincs dupla értesítés.

**Egy nit, ami nem stílus.** A meglévő komment utolsó mondata a beszúrás után hamis lett:

> Installs that never ran isolated have no .channels-config dir and stay quiet, so default setups see no new noise.

Az új trigger után ez pont azokra a telepítésekre NEM igaz, amelyek flotta-tokent hordoznak: ők mostantól kapnak jelzést. A mondat így egy nem létező garanciát ígér a következő olvasónak. Írd át arra, ami igaz: a token nélküli telepítések maradnak némák.

Ezt javítsd, aztán mergelem. Van egy másik change-request a #838-on is.

Ami a mergen túl van, és rögzítem, hogy ne vesszen el: valódi friss telepítést tiszta gépen te sem mértél, én sem. A develop-merge-hez ezt nem követelem meg, mert az még nem kiadás. A KEDDI promote elé viszont igen: a telepítő-szkripteket a soak nem futtatja, és pont ez volt a mai reggeli kiadás kapu-gapje. Eldobható hosztot kérek rá, előtte szólok.

PR #849, opened 2026-08-02
fix(costops): let the quota guard prove it ran, not just complain

VTolnai:
Kanban #225. The guard logged **only failures**, which made "running and fine" indistinguishable from "not running at all" -- both were silence, and only one of them is good news. On 2026-08-02 it failed three checks in a row (03:09, 03:29, 03:49) and nobody was told, because nothing had crossed a threshold.

This is the false-green shape we hunt in other people's code, sitting in ours.

## What changed

**A successful read now leaves a trace** -- `last_success_at`, the utilization it read, and an INFO line -- on **every** cycle, not only when it alerts. Written *before* the alert branch, so a failing notify cannot lose the evidence that the check happened.

**Repeated failures become their own alert** after three in a row (one miss is ordinary: a token refresh, a blip at the endpoint). The message says how long we have been blind and, explicitly, that the quota is **not being watched** until it clears -- a guard that cannot look must not be mistaken for a guard reporting all-clear. Latched, so an outage alerts once rather than every twenty minutes.

**`isMeasurementStale()`** answers "has this thing run recently" without the caller re-deriving the arithmetic: no successful reading, or one older than two intervals, means presumed dead.

## One existing test changed meaning, and I did not quietly re-point it

The recovery path asserted `toHaveBeenCalledWith(EMPTY_QUOTA_STATE)` -- the whole state blanked. That is exactly the bug now: the alert latch is about alerting, the success trace is about liveness, and clearing one must not erase the other, or the guard looks dead on the cycle right after it recovered. The test now asserts both halves separately, with the reason in a comment rather than in a commit nobody re-reads.

## Verification

- 12 new tests; full suite **3150 pass** in a worktree on develop head; `tsc --noEmit` clean.
- **Control**: with the success write, the state-preserving alert write and the failure alert each disabled, exactly the 4 tests that cover them fail.

Refs: kanban #225


VTolnai:
Not merging as-opened: the head branch sits on the box-local develop (645d5ae), which has diverged from this repo's develop (35 ahead / 74 behind), so the PR would carry the entire un-curated local stack upstream, against the curated-PR policy. The actual fix (7561585, quota-guard.ts + tests only) was reviewed, fast-forwarded onto the box develop, full suite 3150 pass in a clean worktree, and deployed to the live dashboard. Upstreaming the costops stack remains a separate, owner-gated decision.

PR #875, opened 2026-08-04
fix(channels): the respawn path must read the model from the same place the launch path does

Xerath-gnx:
## What

`readConfiguredMainModel()` (channel-monitor.ts) now mirrors the precedence that `scripts/channels.sh resolve_main_model()` already uses: `.env MAIN_AGENT_MODEL` wins over the tracked `.claude/settings.json`.

## Why

The two paths disagreed, and the disagreement was silent.

`.env` is gitignored precisely so an install can choose its own main-agent model without dirtying a tracked file (a dirty tracked file blocks the update preflight's clean-tree check and gets reverted by the next update). The launch path honours that. The respawn path did not: it read `settings.json` only, so the two agreed exactly as long as somebody kept both files in sync by hand.

That is worse than a plain inconsistency, because the tracked `settings.json` ships a model of its own and an update reverts local edits to tracked files. The next soft resume can then hand the main session a **different model than the one it launched with** -- no dialog, no error, no log line -- while the launch path keeps reporting the intended value. An operator who requires a minimum model tier is moved below it by a routine update, and nothing in the system says so.

Observed on a live install on 2026-08-04: `.env` named one model, the running session had been started from it, and the tracked `settings.json` still carried the repo default from an earlier release. Nothing had broken yet only because no soft resume had happened since the edit.

## How

- new `readEnvValue(projectRoot, key)` helper, parsing the same way channels.sh does (`grep '^KEY=' | head -1 | cut -d= -f2-`)
- `readConfiguredMainModel()` checks `.env MAIN_AGENT_MODEL` first, then falls back to `settings.json` as before
- `readExtraChannelPluginIds()` reuses the helper (same file, same rule, less duplication)
- an empty `MAIN_AGENT_MODEL=` deliberately does **not** win: a half-edited `.env` would otherwise drop the `--model` flag entirely and hand the session to the CLI's built-in default, which is the drift this function exists to prevent

## Tests

`src/__tests__/main-model-env-precedence.test.ts`, 10 cases. They lock **both** sides -- the TS precedence and the shell function it mirrors -- so flipping the order in `channels.sh` alone re-opens the gap loudly instead of silently.

Verified as a real regression guard, not just green: with the `.env` branch removed the suite fails 2/10 in exactly the intended places, and passes 10/10 with it.

`npx tsc --noEmit` clean.

Note on the wider suite: on this branch `npx vitest run` reports 19 failures in 4 files (`email-send-gate`, `governance-gates`, `hook-command-quoting`, `hook-path-guard`). I measured a pristine `origin/develop` checkout at the same commit (8df36ff) and it reports the **same 19 failures in the same 4 files**, so they are pre-existing and untouched by this change.

### DIFF
--- a/src/__tests__/main-model-env-precedence.test.ts
+++ b/src/__tests__/main-model-env-precedence.test.ts
@@ -1,0 +1,124 @@
+// Regression tests for the split-brain between the LAUNCH and the RESPAWN path
+// when resolving the main agent's model.
+//
+// Background (2026-08-04): scripts/channels.sh resolves the main model from
+// `.env MAIN_AGENT_MODEL` first and only then from `.claude/settings.json` --
+// deliberately, because .env is gitignored while settings.json is tracked, so an
+// install that wants its own model can set it without dirtying a tracked file
+// (which would otherwise block the update preflight and be reverted by the next
+// update). `readConfiguredMainModel()` in channel-monitor.ts read ONLY
+// settings.json, so the two paths agreed just as long as someone kept both files
+// in sync by hand.
+//
+// Why that is worse than a plain inconsistency: the tracked settings.json ships a
+// model of its own. An update reverts local edits to tracked files, so the very
+// next respawn can hand the main session a DIFFERENT model than the one it
+// launched with -- silently, with no dialog, no error and no log line, while the
+// launch path keeps reporting the intended value.
+//
+// The asserts below lock BOTH sides: the TS behaviour, and the shell function it
+// is supposed to mirror. A future edit that flips the precedence in channels.sh
+// alone would otherwise re-open exactly the same gap from the other direction.
+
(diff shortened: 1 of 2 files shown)


Szotasz:
Koszonjuk. A **te implementaciod nyert**, es kiment a v1.30.0-ban.

Ket PR erkezett ugyanarra a hibara (ez es a #884). A tied lett a kanonikus, ket okbol: a helyes agra (develop) ment, es tisztabb megoldas volt a kozos helper miatt. A #884 tesztje viszont erosebb volt (valodi shell-futtatas, tobb hataresettel), ezert azt atemeltuk a te agad tetejere. A PR #886 hordozza a kettot egyben, ezert zarom ezt.

A szerzoseged megorzodott.

PR #878, opened 2026-08-04
feat(desktop-lock): make the "I am driving the screen" lock something the scheduler can see

Xerath-gnx:
## What

Turns the fleet's desktop-lock convention into state a machine can gate on: `POST/DELETE/GET /api/desktop-lock`, a `requiresDesktop` flag on scheduled tasks, a recorded skip, and a TTL sweeper.

## Why

The lock already existed as a convention -- the agent about to drive the GUI broadcast `[DESKTOP-LOCK]` and its peers kept their hands off. That works for agents reading their inbox, and not at all for the scheduler, which fires a GUI round into a session already driving the same screen.

## Four decisions worth reviewing

**One call, not two.** `POST` sets the state **and** sends the broadcast. The obvious alternative -- let the holder call the API next to its existing broadcast -- is two mechanisms with no shared state, which is not two protections but one extra failure path: broadcast without the call leaves the gate open while the fleet believes the screen is taken; the reverse leaves it closed while everyone believes it is free. Neither divergence produces an error message.

**The gate is narrow.** Only tasks marked `requiresDesktop` in their own `task-config.json` wait. Email, kanban, memory and the watchdogs keep running -- a blanket suspension produces blanket silence, and silence is what nobody questions. The flag lives on the task, not in a list of task names in the runner, so a GUI round added later is covered without anyone remembering this file.

**A skipped round is never silent.** Each gated tick writes a `skipped-desktop-lock` task_run plus a record naming the round, the time, the holder and how long the screen is expected to stay taken. An untraced dropped tick is indistinguishable from "there was nothing to do" -- which is how a missed customer message disappears. Observed the day this was written: two consecutive WhatsApp rounds were lost behind two GUI rounds, and only a human noticing the gap brought them back.

**Expiry is a finding, not routine.** Estimate + 30 min grace (30 min default if no estimate was given, and the gate says it fell back -- a missing estimate means someone skipped a step the protocol asks for). A second `POST` by the same holder **extends** rather than conflicts, so a round that grows mid-run does not force an up-front over-estimate. On expiry the sweeper opens the gate, logs it, tells the fleet, and tells the **holder** -- the one party that may still be at the screen and would otherwise learn about it from a collision.

## Tests

`src/__tests__/desktop-lock-gate.test.ts`, 13 cases: every gate branch (non-desktop task runs, holder not gated against itself, grace period, default-estimate wording, expiry), state IO (a corrupt or half-written lock file reads as unlocked rather than wedging the fleet), and the skip trail (two consecutive skips both survive, the file is bounded, an unwritable path never throws -- losing the trail beats losing the run loop).

Verified as a real guard, not just green: with the gate short-circuited 4 of the 13 fail.

`npx tsc --noEmit` clean. The wider suite shows the same 19 pre-existing failures in 4 files (`email-send-gate`, `governance-gates`, `hook-command-quoting`, `hook-path-guard`) that a pristine `origin/develop` at 8df36ff shows; untouched here.

## Follow-up (not in this PR)

The holders' own runbooks need the one-line call added where they broadcast today, and the three screen-driving tasks need `requiresDesktop: true`. Both are install-side config, not code.

### DIFF
--- a/src/__tests__/desktop-lock-gate.test.ts
+++ b/src/__tests__/desktop-lock-gate.test.ts
@@ -25,6 +25,7 @@ import {
   decideDesktopGate, lockExpiresAt, readDesktopLock, writeDesktopLock, clearDesktopLock,
   recordDesktopSkip, TTL_GRACE_MS, DEFAULT_ESTIMATE_MS, type DesktopLock, type DesktopSkipRecord,
 } from '../web/desktop-lock.js'
+import { parseLockBody } from '../web/routes/desktop-lock.js'
 
 const NOW = 1_785_855_000_000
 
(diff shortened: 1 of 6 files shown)


Szotasz:

PR #882, opened 2026-08-04
feat(agents): show what each agent is doing on its own card, and retire the Activity page

greksazoo:
## What

The Activity page and the agents page answered two halves of the same question -- *who* the agent is, and *what it is doing right now* -- and you had to navigate between them to put the halves together. The live status is now a card mode on the agents page, and the separate page is gone.

## Agents page

- A second, **icon-only** toggle picks the card content: **Adatlap** (the original card) or **Aktivitás** (the live pane status). Avatar, name, main-agent badge and description stay in both, so the card never stops saying *which* agent it is.
- Activity mode reuses the Activity page's markup and rules: state badge, permission-mode chip, last lines of the pane, state colour on the left border, and click-to-open-terminal **only for a running agent** (a stopped one has no session to look at).
- The 3s `/api/agents/activity` poll that already ran on this page feeds the card bodies too -- no second consumer, no new endpoint. Bodies are written every tick regardless of mode, so switching modes shows the current state instead of a blank card.
- Thin vertical rules separate the header button clusters. The card-mode toggle and its trailing rule hide together in org-chart view, which would otherwise leave two rules with nothing between them.

## Org chart

- The run-state line gets the same pulsing dot the agent cards use. It used to be a `●` baked into the translation, which read as "stopped" next to a card whose dot was breathing.
- A node's border turns green while that agent is working, from the same signal, toggled **in place** rather than re-rendering -- so a drag-and-drop reporting edit survives the tick.
- The hint below the tree is centred and now also says the tree can be rearranged by dragging. **It never localised:** the `#teamPage > p` selector stopped matching anything when the Team page was merged into this one, so it was stuck on its hard-coded Hungarian.

## Accessibility

Icon-only buttons carry a native `title` tooltip, and the label element stays in the DOM visually hidden, so screen readers still announce which button is which.

## Testing

No automated test covers `web/`, so this was measured in the browser rather than asserted:

- All five activity states verified against the card styling (working green, idle grey, error red, unknown yellow, stopped grey + dimmed + not clickable). The stopped branch had to be forced, since every agent was running at the time -- stated as forced, not sold as live.
- Measuring a forced state needs the 3s poll stopped and the CSS transition allowed to settle, otherwise the read lands mid-animation on the previous value.
- Org chart: 8 running agents, 3 with the active border -- i.e. the border discriminates rather than painting everything.
- Mode and view round-trips checked both directions, including that the card-mode toggle and its divider hide together.
- `npm run syntax-check` passes.


Szotasz:
Heads-up on why this PR looked stuck, and what is blocking it now.

**It was never actually reviewed as red.** Two status checks are required on `develop` (`scan` and `test`), but `test.yml` only landed on 2026-08-22, and the head of this PR is from 2026-08-19. The required check therefore never ran on this commit and never would have on its own, so the PR sat in a permanent BLOCKED state that looked like "checks pending". I reopened it today with no commit and no force-push (the head is byte-identical, `5e310f94`), which re-triggered the suite.

**Now there is a real, small failure.** `src/__tests__/refactor-config-widget-index.test.ts` fails on both lang files:

```
found orphaned keys in en.js: activity.no_output
found orphaned keys in hu.js: activity.no_output
```

That test asserts every key in `web/lang/en.js` and `web/lang/hu.js` has a consumer in the codebase. Since this PR retires the Activity page, the last reference to `activity.no_output` went away, but the key itself stayed behind in both files. Removing that one key from `en.js` and `hu.js` should turn the suite green.

Nothing else is blocking: `scan` is green, the branch has no conflicts, and the diff is scoped to the five web files it says it touches.

Thanks for the patch, and sorry it sat this long for a reason that had nothing to do with your code.

greksazoo:
Thanks for digging into why it stalled -- the `test.yml` timing explanation matches what we see.

One correction on the suggested fix, because we measured it and it would ship a visible regression.

**`activity.no_output` is not orphaned in this PR -- it is reused.**

`web/app.js:748` (this PR's head, `5e310f94`) still calls it, in the *new* per-agent card, not in the retired Activity page:

```js
: '<p class="activity-tail-empty">' + (entry.running ? t('activity.no_output') : t('activity.not_running')) + '</p>')
```

And `t()` (`web/app.js:26-34`) falls back to returning the **raw key**:

```js
const str = window._i18n?.[lang]?.[key] ?? window._i18n?.['en']?.[key] ?? key
```

So deleting the key from `en.js` / `hu.js` would turn the suite green while making the UI print the literal string `activity.no_output` on every running agent that currently has no output.

**What is actually stale is the `DEAD_KEYS` list**, not the lang files. `src/__tests__/refactor-config-widget-index.test.ts:128` lists the key on the assumption that it dies with the Activity page. This PR keeps the message and moves it onto the agent card, so the right change is to drop that one entry from `DEAD_KEYS` and leave both lang files untouched.

Worth noting for the check itself: **the test is green either way.** Removing the key from the lang files also satisfies it, so the assertion cannot distinguish the correct fix from the harmful one. Whatever we do here, that is a gap in the guard rather than in this PR.

**Second, smaller thing:** the branch had no conflicts when you wrote your comment, but `60e4bcc` (`feat(web): thinking-orb indicator on the working activity badge`, #1020) landed about a minute later and touches the same region. There is now exactly one conflict, a single block in `web/app.js`; `index.html`, `style.css` and both lang files still merge cleanly.

We are preparing both changes on the branch now. If the conflict resolution turns out to need a product decision -- i.e. the thinking-orb behaviour and this PR's card rendering are two different answers to the same thing rather than two separate edits -- we will stop and ask here rather than pick one.

greksazoo:
Pushed `fee1e2d`: `develop` merged in, and one line removed from `DEAD_KEYS`. Both checks are green and the branch is mergeable again.

**If you would rather not have your guard test touched, there is an alternative that leaves it alone.** We can rename the key on our side instead:

1. add a new key (`agents.card.no_output`, say -- the `activity.` prefix is what went stale) to `en.js` and `hu.js` with the existing text,
2. point `agentActivityBodyHtml` at it,
3. then delete `activity.no_output` from both lang files -- at which point it really is orphaned, and your `DEAD_KEYS` entry stays as written.

Same green, your test untouched, and the message survives on the card. We checked that nothing else in `refactor-config-widget-index.test.ts` objects to a *new* key -- every assertion there tests for absence, not presence. Say the word and we will swap it; otherwise the current one-line version stands. Your call, it is your guard.

---

**Two heads-ups, both measured, neither of them a request.**

**1. #1020's thinking-orb loses its only call site through this PR.** `act-orb` is referenced exactly once in `web/app.js` on `develop` (line 778), inside `renderActivity()` -- which this PR removes along with the rest of the Activity page. The new `agentActivityBodyHtml` renders the same `activity-badge` without it, so after the merge `web/app.js` has zero `act-orb` references while the 4 CSS rules stay. Net effect: the `working` state silently loses the indicator you shipped, and the CSS is dead.

Carrying it over is a one-liner in the new badge, but it is your design call whether the per-agent card should have it at all, so we did not decide it. Happy to add it in this PR, or leave it and let you do it separately.

**2. Unrelated to this PR: three keys `app.js` calls do not exist in either lang file on `develop`.**

```
autonomy.level.1               web/app.js:13672
connectors.tooltip.docs        web/app.js:7805
ideas.toast.score_saved_error  web/app.js:15141, 15151
```

`t()` falls back to the raw key, so those three render as literal `autonomy.level.1` etc. in the UI today. `autonomy.level.1` is also in `DEAD_KEYS`, so the guard is actively asserting that a key the code still calls stays absent -- the same shape as the case above, just already shipped.

That suggests the missing assertion is the other direction: every key `app.js` passes to `t()` should exist in both lang files. We measured it -- 908 keys called, 8 not found, of which 5 are runtime-built prefixes (`approvals.status.`, `federation.peer_state.`, `federation.routing.mode.`, `kanban.gantt.legend.`, `settings.desc.`) and 3 are the real misses above. Happy to open a separate issue or PR for it if useful, but we did not want to bundle it into this one.

greksazoo:
Correction to something I wrote above, and the current state.

**I said "the `activity.` prefix is what went stale". That was wrong.** The namespace is alive: after this PR, six `activity.*` keys remain (`state.working`, `state.idle`, `state.unknown`, `state.error`, `state.stopped`, `tooltip.mode`) and every one of them still has a caller. What actually went stale is the Activity *page's* keys -- `badge.main`, `empty`, `error_load`, `loading`, `page_title`, `page_subtitle`, `updated` -- and this PR drops those.

So the rename was not a namespace migration and should not be read as the start of one. It is scoped to the one line we had to rewrite anyway.

**Current head is `e6ee385`, mergeable, both checks green, and the diff is back to exactly the five `web/` files.** Your `refactor-config-widget-index.test.ts` now has a zero-line diff against `develop` -- we reverted the `DEAD_KEYS` edit and took the rename route instead, so your guard stands exactly as you wrote it.

Two cleanups went in alongside it, both consequences of our own change:

- `activity.not_running` sat on the same line as the renamed key and would have left that line half-migrated, so it became `agents.card.not_running` too. The commit message states the limit explicitly, so nobody later reads it as an abandoned migration.
- `activity.tooltip.terminal` was left in both lang files with no remaining caller after `renderActivity()` went away. That is an orphan *this PR created*, so we removed it.

On the second one, worth reporting because it bears on the guard: deleting that orphaned key moved **zero** tests. Full suite before and after is byte-identical. Nothing in the repo asserts that a key in the lang files still has a consumer, so that class has to be caught by hand.

We also swept the other direction while we were there: of the 1616 keys in `en.js`, exactly one was orphaned *by this PR* (the one above). Separately, ~209 appear unused on `develop` as well -- mostly `common.*` and `approvals.*`. We did not touch those; they are not ours to clean, and the number is a literal-match upper bound, so treat it as a lead rather than a count.

PR #884, opened 2026-08-05
fix(channel-monitor): respawn must honour MAIN_AGENT_MODEL, like the launch path

HiperiorNr1:
> AI-generated by Marveen dev agent (forge, Claude Opus 5).

Refs #697. Completes option (a): the non-versioned override exists and now
takes precedence in EVERY consumer, not just the launch path. Does not
implement option (b) -- `.claude/settings.json` stays tracked. The override
lives in `.env MAIN_AGENT_MODEL` rather than the suggested
`store/settings.local.json`: different file, same principle.

## Problem

The main agent's model is resolved by two independent implementations:

| | | |
|---|---|---|
| launch | `scripts/channels.sh` | `resolve_main_model()` |
| respawn | `src/web/channel-monitor.ts` | `readConfiguredMainModel()` |

Since 2026-07-29 the shell honours `MAIN_AGENT_MODEL` from `.env` and falls back
to `.claude/settings.json`. The TypeScript side read **only** `settings.json`,
while its own comment claimed to be "mirroring scripts/channels.sh". It was not.

So an install that uses the supported `.env` route boots on the operator's model
and comes back on the **repository's** model after every nightly restart,
recovery resume and hard respawn (the three call sites this function feeds).
The failure is silent in the worst way: nothing throws, no probe goes red, the
agent simply answers as another model.

This installation is a live reproduction of #697. Before the fix, `git status`
showed exactly the permanent local diff the issue describes:

```
 M .claude/settings.json
-  "model": "claude-opus-4-8[1m]"
+  "model": "claude-opus-5"
```

That diff could not be dropped, because dropping it is what activated the bug.

## Fix

`readConfiguredMainModel()` now applies the shell's precedence exactly:

- `MAIN_AGENT_MODEL` from `.env` wins over `.claude/settings.json`
- an **empty** `MAIN_AGENT_MODEL` does not shadow `settings.json` -- the `[ -n ... ]` semantics
- an anchored `MAIN_AGENT_MODEL=` prefix, so a similarly named key cannot leak in
- an unreadable `.env` falls through to `settings.json` instead of breaking the respawn
- exported with a `projectRoot` test seam, matching `readExtraChannelPluginIds()` below it

`main-model-resolution-parity.test.ts` drives 8 fixtures through **both**
implementations -- the TS function and `channels.sh --resolve-main-model` -- and
asserts each is right *and* that they agree. A TS-only test would have stayed
green through the entire defect, because the defect was a divergence.

## Verification

Before/after on the reporting installation, restoring the clean tree that #697
asks for:

```
.claude/settings.json  claude-opus-4-8[1m]   (repository default, tracked, tree CLEAN)
.env MAIN_AGENT_MODEL  claude-opus-5         (per-install override)

launch  path (channels.sh)              -> claude-opus-5
respawn path (readConfiguredMainModel)  -> claude-opus-5
```

The permanent local diff is gone and both paths agree. Before this change the
respawn path returned `claude-opus-4-8[1m]` from the same state.

- Negative control: with the previous `settings.json`-only body exported unchanged, exactly the 4 `.env`-dependent fixtures fail and the other 6 pass -- the suite discriminates on logic, not on the new export.
- Tests: `npm run test` -> 3224 passed, 1 skipped.
- `npm run typecheck` -> exit 0.

One unrelated test fails on this branch: `installer-start-and-fallback.test.ts`
expects `TRAP:6` from an ERR trap's `$LINENO`, which is bash-version dependent
(bash 5.2.21 reports `TRAP:5`, the failing assignment itself). It reproduces on
a clean `upstream/main` checkout at 6add7da with none of these changes applied,
so it is pre-existing and untouched here. Happy to send it as a separate PR.

### DIFF
--- a/src/__tests__/main-model-resolution-parity.test.ts
+++ b/src/__tests__/main-model-resolution-parity.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import { execFileSync } from 'node:child_process'
+import { mkdtempSync, mkdirSync, writeFileSync, copyFileSync, rmSync } from 'node:fs'
+import { join, dirname } from 'node:path'
+import { tmpdir } from 'node:os'
+import { fileURLToPath } from 'node:url'
+
+import { readConfiguredMainModel } from '../web/channel-monitor.js'
+
+// The main agent's model is resolved by TWO independent implementations:
+//
+//   LAUNCH   scripts/channels.sh   resolve_main_model()      (shell)
+//   RESPAWN  src/web/channel-monitor.ts readConfiguredMainModel()  (TS)
+//
+// They must agree. When they disagree the failure is SILENT in the worst way:
+// the agent boots on the model the operator chose and comes back on a DIFFERENT
+// one after the nightly restart / a recovery resume / a hard respawn. Nothing
+// throws, no probe goes red -- the agent simply answers as another model.
+//
+// That is not hypothetical. Until 2026-08-03 readConfiguredMainModel() read
+// ONLY .claude/settings.json, while its own comment claimed to mirror
+// channels.sh. channels.sh had honoured MAIN_AGENT_MODEL from .env since
+// 2026-07-29 (that route exists because settings.json is TRACKED: a per-install
+// model written there is a permanent local diff that blocks the update
+// preflight's clean-tree check and is reverted by the next update). So any
+// install using the supported .env route drifted on every respawn.
+//
+// The defect was MASKED on the install where it was found: settings.json
+// happened to be dirty with the same value the .env carried, so both paths
+// agreed by accident. Restoring a clean tree would have ACTIVATED it. A test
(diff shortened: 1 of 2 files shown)


Szotasz:
Koszonjuk, es elnezest, hogy nem ez a PR ment be egy az egyben.

Ugyanarra a hibara ket PR erkezett (ez es a #875). Az implementaciot a #875-bol vettuk, mert az a develop agra ment, ez pedig a mainre -- a mi folyamatunkban a main a kiadott ag, tehat oda kozvetlenul nem mergelunk.

**A te tesztedet viszont atvettuk**, mert erosebb volt: valodi shell-futtatassal ellenorzi a ket ut egyezeset, es olyan hataresetekre is kiter, amiket a masik nem fedett (ures ertek, szogletes zarojeles utotag, hasonlo nevu kulcs). A PR #886 ezt hordozza, co-authorkent.

A javitas kiment a v1.30.0-ban.

PR #887, opened 2026-08-05
fix(self-pace-gate): stop quoted prose from faking a command position

Xerath-gnx:
## What happened

Five hard-gate denials in one morning (2026-08-05), three mine and two taric's. None of them scheduled anything. Each was an inter-agent message or a kanban write whose **text** quoted the same grep pattern:

```
Minta: stop.sh | launchctl | com.janna.dashboard
```

`splitSegments` is not quote-aware, so those bars split the prose. The middle piece trimmed down to the bare word `launchctl`, and `SCHEDULER_RX`'s end-of-segment branch reads a bare `launchctl` as a real interactive invocation -- which is correct for a real command line.

The expensive part is not the denial, it is its shape: from outside, a hard-gate denial is indistinguishable from an agent that simply stayed silent. Two peers spent a round each hunting the wrong hypothesis (heredoc? `subprocess.run`? a cron-like token? reading the dashboard token?) before the cause was measured.

## Why it is a design bug, not a bad pattern

The same text **passes** as `curl -d '<json>'` and is **denied** from a python heredoc. `stripDataPayloads` blanks the first; the second has no `-d` argument to blank. Choosing how to send a message had quietly become a security decision, and no one designed it that way.

`stripDataPayloads`' own comment names this false-positive class as its target. It is implemented for exactly one route -- so this is unfinished work, and every future send route inherits the same gap.

## The fix

Mask inert literal text to spaces before the **anchored** scheduler check: single-quoted, ANSI-C `$'...'`, double-quoted without a substitution, and quoted-tag heredoc bodies. The text is gone, and with it the separators it was faking. `splitSegments` and every regex are untouched and run unchanged on what remains.

### Scope, which is a measurement and not caution

The masked text feeds **only** the anchored scheduler check. The unanchored patterns (`tmux`+send-keys, `nohup`+claude, `claude`+`/loop`) keep scanning the raw segments, quoted regions included.

My first design handed quote-aware output to every check, on the assumption that the gate does not catch a real vector inside a heredoc anyway. Three of four probes agreed:

| probe | today |
|---|---|
| `subprocess.run(['crontab','-e'])` in a heredoc | ALLOW |
| `os.system("crontab -e")` in a heredoc | ALLOW |
| `python3 -c "... os.system('crontab -e')"` | ALLOW |
| `subprocess.run(['tmux','send-keys',...])` in a heredoc | **DENY** |

The fourth is the one that matters: the tmux pattern is unanchored and matches anywhere, so the tidy version of this change would have removed the detection of this gate's **founding incident vector** while calling itself a structural fix.

### Fail-closed in three places

"Cannot parse" must mean *scan more*, never *scan less*:

- unterminated quote or heredoc -> `null`, caller falls back to the naive split
- a double-quoted region containing `$(...)` or a backtick is not inert -- the shell runs what is inside, so a separator in there is a real one
- a heredoc with an **unquoted** tag whose body substitutes is not inert either

### One corpus case drove the shape

With only a quote-aware *splitter*, `echo 'foo | crontab | bar'` stayed denied: `SCHEDULER_RX` carries its own boundary anchor and re-finds a command position **inside** a segment. Leaving the quoted text in the segment at all was the mistake. The `launchctl` cases had passed by luck -- `LAUNCHCTL_SUBCOMMAND`'s lookahead happened to reject the following bar.

## Verification

- 30-case corpus: every vector the gate exists to stop, the legitimate reads, the three fail-closed paths, and the five measured false positives. All as expected.
- Committed suite: **65 -> 73 passing**. The 3 failures in `governance gate scaffold wiring` are present on untouched `develop` too -- measured by re-running with the change stashed, not assumed.
- `tsc --noEmit` clean.
- taric's two exact blocked commands are preserved as reproduction material and both now pass.

## Separate decision, deliberately NOT in this PR

`os.system("crontab -e")` and `python3 -c` scheduler calls pass the Bash layer **today** (see the table above). Closing that is a **strictness increase**, not a false-positive fix. It does not belong in a PR whose subject is over-blocking, and it needs its own decision -- I am not smuggling a behaviour change into a bug fix.

The Bash hook is documented as the second layer (the runtime tool-deny is primary) and the file already lists accepted limitations, so this may well be a known trade-off. It was not written down, so now it is.

### DIFF
--- a/scripts/self-pace-gate.mjs
+++ b/scripts/self-pace-gate.mjs
@@ -153,6 +153,120 @@ export function splitSegments(command) {
     .map((s) => s.trim())
 }
 
+// Split like splitSegments, but ONLY on separators the shell would actually
+// treat as separators -- never on one that sits inside a quoted string or a
+// heredoc body. Returns null when the quoting cannot be resolved with
+// confidence, and every caller must then fall back to the naive splitter.
+//
+// WHY THIS EXISTS (measured 2026-08-05, five denials in one morning -- three
+// mine, two taric's): splitSegments is not quote-aware, so PROSE can manufacture
+// a command position that never existed. All five denials had the same cause: a
+// grep pattern quoted inside an inter-agent message,
+//   Minta: stop.sh <bar> launchctl <bar> com.janna.dashboard
+// The `<bar>` split it, the middle piece trimmed down to the bare word
+// `launchctl`, and SCHEDULER_RX's end-of-segment branch reads a bare `launchctl`
+// as a real (interactive) invocation -- correctly, for a real command line.
+// Nothing was scheduled; five messages simply never went out. From outside, a
+// hard-gate denial is indistinguishable from an agent that stayed silent.
+//
+// The route decided it: the SAME text passes as `curl -d '<json>'` (the payload
+// is blanked by stripDataPayloads) and is denied when sent from a python
+// heredoc, which has no -d argument to blank. Choosing how to send a message
+// had quietly become a security decision. stripDataPayloads' own comment names
+// this false-positive class as its target -- it is implemented for exactly one
+// route, so the gap is unfinished work, not an oversight.
+//
+// SCOPE, and this is the part that matters: the result feeds ONLY the anchored
+// scheduler check. The unanchored patterns (tmux+send-keys, nohup+claude,
+// claude+/loop) keep scanning naive segments, quoted regions included, because
+// they do NOT depend on a command position that prose can fake -- and because
+// measurement showed the naive scan is what catches a real
+// `subprocess.run(['tmux','send-keys',...])` hidden in a heredoc body. Handing
+// them quote-aware segments would have removed the detection of the very
+// incident vector this gate was built for, under the banner of a structural fix.
+//
+// FAIL-CLOSED in three places, because "could not parse" must mean "scan more",
+// never "scan less":
+//   - unterminated quote or heredoc -> null (caller uses the naive split)
+//   - a double-quoted region containing $(...) or a backtick -> null; the shell
+//     runs what is inside, so a `;` in there IS a real separator
+//   - a heredoc with an UNQUOTED tag whose body contains $(...) or a backtick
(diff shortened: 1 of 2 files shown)


Szotasz:
Mergelve, köszönjük. Két dolgot emelek ki, mert a PR-en mindkettő látszik, és ritka.

**A scope-döntés volt a lényeg, és jól választottad.** A maszkolt szegmensek kizárólag a horgonyzott scheduler-ellenőrzéshez jutnak el; a nem-horgonyzott minták és a store/API írás-ellenőrzés változatlanul a naiv szegmenseken futnak. Ezt a diffben magunk is visszamértük: a `maskInertLiterals` pontosan egy helyen fut be. A kommentben leírt indoklás sem utólagos magyarázat, hanem a döntés oka: quote-aware szegmenseket adni a nem-horgonyzott mintáknak épp azt a felismerést vette volna el, amiért a kapu készült.

**A fail-closed irány végig helyes.** A feloldhatatlan idézés `null`, a hívó pedig visszaesik a naiv splitre, tehát a bizonytalanság mindig "többet nézünk", soha nem "kevesebbet". A `$(...)` és a backtick kezelése a dupla idézőjeles régióban és a nem idézett heredoc-tagnél ugyanezt a szabályt követi.

Amit a merge előtt ellenőriztünk, mert a kapu élesben minket is fog:

- Élő probe-mátrix a develop és a PR ága mellett: a valódi service-restart parancs (kickstart és stop) mindkét ágon DENY marad, a heredocba rejtett valódi pane-injektálás szintén. Egyetlen valódi parancs sem csúszott át; a változás pontosan a két idézett-próza esetet fordítja meg.
- Mutációs kontroll: a javítás kivételével a governance-suite bukása 3-ról 7-re nő, tehát a nyolc új tesztből négy ténylegesen ehhez a fixhez kötött. A tesztek nem fogatlanok.

A "hogyan küldöm el az üzenetet" mint biztonsági döntés jó megfogalmazás volt, és a rögzített mérés (öt megtagadás egy délelőtt) többet ér, mint maga a patch.

PR #891, opened 2026-08-05
feat(heartbeat-gates): token-free gates for garmin-run-watcher + memoria-heartbeat

maslentrek-dev:
## Summary
- Adds `scripts/garmin_run_gate.py` and `scripts/memoria_heartbeat_gate.py`: deterministic, token-free gates that run in front of the garmin-run-watcher and memoria-heartbeat scheduled tasks and only wake the agent when there's something to actually judge, instead of a model call on every tick.
- Finished and tested since 2026-07-28, sitting untracked on disk through two unexpected reboots before being committed here.
- Follow-up commit resolves 5 hardcoded `/Users/gruzmanarnold/...` absolute paths found in pre-merge review (both repos are public, this string was never in history before): `MARVEEN_DIR` now derives from `Path(__file__).resolve().parents[1]`, `GARMIN_DIR` reads `GARMIN_DATA_DIR` env var with a `~/garmin-data` fallback.

## Test plan
- [x] `python3 scripts/test_garmin_run_gate.py` -- all scenarios passed
- [x] `python3 scripts/test_memoria_heartbeat_gate.py` -- all scenarios passed
- [x] Verified no `/Users/gruzmanarnold` strings remain in either script or its tests
- [x] Verified working-tree files are byte-identical to what's committed

### DIFF
--- a/scripts/garmin_run_gate.py
+++ b/scripts/garmin_run_gate.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python3
+"""Token-free gate in front of the garmin-run-watcher heartbeat.
+
+The watcher used to be a `type: heartbeat` task: every tick woke Seven with a
+prompt whose first step was to run running_analysis.py and fall silent when it
+printed "Mar elemezve". Measured over 7 days that was 100% of the ticks -- a
+model call per tick to learn that nothing happened. This script makes the
+no-op path cost nothing and only wakes Seven when there is something to judge.
+
+Exit-code contract of running_analysis.py (read from the source, not guessed --
+running_analysis.py:157-196):
+
+    main() -> 0  ... no runs at all, or latest run already analysed
+    main() -> 1  ... NEW run; analysis written to pending_run_analysis.txt AND
+                     last_analyzed_run.json advanced
+    sys.exit(0 if result == 0 else 2)
+
+so the process exits 0 for "nothing to do" and 2 for "new run". An uncaught
+exception (Garmin login failure, network) exits 1. The codes already
+discriminate, so we do NOT string-match the Hungarian output: those strings are
+accented and inconsistent ("Mar elemezve" vs "Uj futas talalva") and a locale
+or wording change would silently turn every tick into "nothing new".
+
+WHY A CRASH IS NOT exit 0. The task spec asked for "already analysed OR error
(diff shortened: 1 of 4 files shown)


Szotasz:
Reviewed, and the engineering here is good. One change request, and it is about product scope rather than code quality.

What we verified independently (not read from the description):

- The merge into today's develop is clean, the scope is exactly four new files, nothing existing is modified.
- The token-free claim holds for the gate decision, measured with both a positive and a negative control against the real notifier rather than the PR's mock: on the no-op path with no token present the gate still returns success, so it never touches a token; on the rare new-run path it does read the local dashboard token and fails loudly when it is absent. So the per-tick model wake is eliminated on the common path, not moved elsewhere.
- No absolute user paths and no personal identifiers remain in the added lines. Your follow-up cleanup held.

On `memoria_heartbeat_gate.py` we are happy to take it as is. The design decision we liked is the one you wrote down explicitly: the gate answers a fact (has anything happened since the agent last caught up) and refuses to judge importance, because a heuristic would fail exactly on the interesting cases. The failure direction matches: if the turn dies before the watermark write, the next tick looks at the same window again. Looking at an hour twice is cheap, skipping it silently is not. That is the property we care about most in anything that can suppress work.

The change we are asking for: please drop `garmin_run_gate.py` and its test from this PR.

The reason is not quality. It is that the task it gates does not exist in this product. We measured: there is no garmin-run-watcher among the scheduled tasks here, and the gate depends on exit codes from `running_analysis.py`, which lives outside this repository entirely. On any install other than yours the script cannot run, so shipping it means we carry code that no user can execute and that we cannot test. Your own reviewer raised the related point: if those exit codes ever change, the gate breaks silently, and we would have no way to notice.

If you would like the garmin gate upstream anyway, the form that works is a documented example rather than a live script, with the exit-code contract stated in the file so a future reader knows what it depends on. Happy to take it that way in a separate PR.

One optional nit while you are in there: the title says token-free, which is exact about the gate decision but can read as "this never touches any token". A sentence in the module docstring noting that the notify path reads the local dashboard auth token, which is not a model call, would prevent that misreading.

Drop the two garmin files and we will merge the memory gate straight away.

Szotasz:

PR #904, opened 2026-08-06
fix(effort): set the fleet effort level via env, not settings.json (EFFORT806)

Xerath-gnx:
## The bug

Claude Code validates `effortLevel` in `settings.json` as
`enum(["low","medium","high","xhigh"]).catch(void 0)`. `"max"` is **not** in that list, and the
`.catch` drops an unknown value **silently** -- no error, no log -- so the default `"high"` wins.

A settings file that says `max` therefore runs at `high`, and nothing anywhere reports the gap.

## The measurement

Measured 2026-08-06 on a live install. Seven settings files said `"max"`; every agent actually ran
at `high`. Two independent readable signals agreed:

- `CLAUDE_EFFORT=high` in every launched subprocess (the CLI writes this itself for children)
- `effort: "high"` on every transcript message, across all agents, since 07-28 -- ~90 sessions, no
  other value ever recorded

Control: a probe session launched with `CLAUDE_CODE_EFFORT_LEVEL=max` recorded `effort: "max"` on
`claude-opus-5`. So the env path works, the signal does react, and the model does not downgrade it.

## The fix

`"max"` **is** a valid level -- but only through the env var / `--effort` flag, which parse against
the CLI's wider list (that one does include `max`) and outrank settings. So the level moves to an
env export at every launch site:

| file | path |
|---|---|
| `src/model-id.ts` | `FLEET_EFFORT_LEVEL` + `EFFORT_LEVELS`, one source of truth |
| `src/web/agent-process.ts` | local sub-agent launch (Claude models only) |
| `src/web/agent-worker.ts` | worker sessions |
| `src/web/ssh-tmux.ts` | remote agent launch |
| `scripts/channels.sh` | main session + the tmux global env every session inherits |

Models without the `max_effort` capability downgrade themselves to `high`, so the export is safe to
set unconditionally for Claude launches. Non-Claude models (Ollama/DeepSeek/OpenRouter) never see it.

## Guards

Two tests, both aimed at the silent-failure class:

- the remote launch command must carry the export -- if a later edit moves this back into
  `settings.json` or drops it, behaviour changes with no visible symptom
- `FLEET_EFFORT_LEVEL` must be a level the CLI recognises -- a typo (`maximum`, `MAX`) would not
  fail loudly, it would quietly cost the whole fleet its effort level

## Trade-off (deliberate)

The env var pins the level for the whole session, so an in-session `/effort` reports
`Not applied: CLAUDE_CODE_EFFORT_LEVEL overrides effort this session`. That is the intent -- an
operator-set fleet floor -- not a side effect.

## Test status

`npx tsc --noEmit` clean. New and neighbouring suites pass (30 tests).

The full suite has 19 failures in 4 files (`governance-gates`, `hook-path-guard`,
`hook-command-quoting`, `email-send-gate`). Those are **pre-existing**: verified by stashing this
change and re-running the same four files on clean `develop` -- identical 19 failures. Not touched
by this PR.

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -281,6 +281,9 @@ fi
 # source removes the phantom entirely. Inherited by every sub-agent via the tmux
 # global env set below.
 export CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false
+# Effort level for THIS shell (the main channels claude inherits it). See the
+# set-environment -g block below for why this is an env var and not settings.json.
+export CLAUDE_CODE_EFFORT_LEVEL=max
 
 CLAUDE="$(command -v claude)"
 TMUX="$(command -v tmux)"
@@ -533,6 +536,16 @@ if [ -n "${ANTHROPIC_API_KEY:-}" ]; then
 fi
 # Propagate the prompt-suggestion disable to every sub-agent tmux session.
 $TMUX set-environment -g CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION false 2>/dev/null || true
+# Effort level for the whole fleet, propagated the same way. It MUST be an env
+# var: Claude Code's settings.json schema validates effortLevel as
+# enum(["low","medium","high","xhigh"]).catch(void 0), so "max" written into a
+# settings file is dropped SILENTLY and the default "high" wins -- measured
+# 2026-08-06, seven config files claimed max while every agent ran at high
+# (card cb42ad6d). The env var parses against the wider list (which includes
+# "max") and outranks every other source. Keep in sync with FLEET_EFFORT_LEVEL
+# in src/model-id.ts -- that constant governs the TS launch paths, this line the
+# tmux-inherited ones.
+$TMUX set-environment -g CLAUDE_CODE_EFFORT_LEVEL max 2>/dev/null || true
 
 # Hybrid channel-coordinator model: the native plugin stays the PRIMARY inbound
 # path (it always polls getUpdates here -- never outbound-only). The standalone
--- a/src/__tests__/remote-launch-command.test.ts
+++ b/src/__tests__/remote-launch-command.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import { buildRemoteLaunchCommand, classifyRunState, classifyRunStateFromExit, buildContinueProbeCommand } from '../web/ssh-tmux.js'
+import { FLEET_EFFORT_LEVEL, EFFORT_LEVELS } from '../model-id.js'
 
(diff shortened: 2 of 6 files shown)


Szotasz:
Szia, es elnezest a hosszu csendert. 27 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

Szotasz:
We rebased this branch onto current `develop` ourselves and pushed the result to your branch. The conflict was ours. One thing in your diff changed in the process, and this comment exists so you hear it from us rather than find it later.

**The base moved under the launch path.** `develop`'s local sub-agent launch grew a `umask` prefix, a `buildLaunchCmd(launchCwd)` closure used by both the normal start and the EPERM `/tmp` fallback, a `DISABLE_AUTOUPDATER` export, a combined `providerEnv` replacing the separate `ollamaEnv` / `deepseekEnv` / `openrouterEnv`, and a second prompt-suggestion variable for the feedback survey. We kept all of that and grafted your `effortEnv` into it, rather than restoring the older command shape your diff was written against. Same for `channels.sh` and `agent-worker.ts`, where `develop` had added lines at exactly your insertion points.

**The one deliberate change to your code, in a separate commit (`620faad5`): the default is now `high`, not `max`, at all three sites** (`FLEET_EFFORT_LEVEL` and both `channels.sh` sites).

The reason is not a disagreement with your measurement. Your finding stands: `effortLevel` in `settings.json` is validated as `enum(["low","medium","high","xhigh"]).catch(void 0)`, so `"max"` is dropped silently and the env var is the only lever that works. What changed is the surrounding fact: the level this fleet actually runs at today is `high` (`templates/settings.json.template:5`). Merging the branch as written would have raised the level of every install that sets nothing, as a side effect of moving the lever.

Our rule for this: a contribution that makes something configurable must not change the current effective value. Whoever sets nothing should not be able to notice the change at all. Raising the level stays available -- that is exactly what the new env var is for -- and becomes its own decision instead of arriving as a rebase artefact.

Your second regression guard still covers the constant: mutating it to `'ultra'` turns "uses an effort level the CLI actually recognises" red, as intended.

Measured on the rebased tree: `tsc` clean; `remote-launch-command` 16/16; the full suite 498 files, 6216 assertions green, 1 skipped.

**One review finding, not blocking, not fixed by us.** The guard you added covers the REMOTE launch command: mutating `CLAUDE_CODE_EFFORT_LEVEL` in `ssh-tmux.ts` turns your test red. The LOCAL sub-agent launch -- the path most agents actually start through -- is not pinned: removing `${effortEnv}` from `buildLaunchCmd` in `agent-process.ts` leaves the full suite green at 6216. The same assertion you wrote for the remote command, applied to the local one, would close it.

PR #909, opened 2026-08-06
fix(model-credit): escalate instead of guessing when no option names the configured model

Betmanforever:
### What this changes

The FABLEFALL1 consent-dialog handler answers the one dialog shape captured in
the field by pressing option 1 -- safe only because that shape's option 1 *is*
"Continue with &lt;configured model&gt;". Nothing constrains the CLI to that one
shape: variants where the first row is a purchase action (e.g. "Buy usage
credits") or where the options list only models the agent is not configured
for are possible renderings of the same dialog family. Only the shape whose
option 1 is "Continue with &lt;configured model&gt;" has been field-captured; both
variants above are anticipated, not observed, as the diff's own comments
state. On those, a blind option 1 spends money or silently switches the
fleet's model -- and the modal then vanishes while the pane, the dashboard
and agent-config.json all keep naming the configured model, so nothing
surfaces the change until someone reads the transcript. Escape is not an out
either: the CLI records it as `cancelled` and continues on the FALLBACK model.

This PR chains a stricter detector *before* FABLEFALL1:

- `detectsModelCreditDialog` + `findModelCreditDialogOption` (pane-state.ts,
  pure string functions): find the option that unambiguously names the
  CONFIGURED model.
- If found: navigate to it -- as one recover-mode critical section on the
  session send lane (`withSessionSendLock`), with a re-probe inside the lane
  (a digit selects by position; a stale capture could confirm a row that
  moved). Fail-closed: a busy lane sends nothing, the dialog is re-detected on
  a later sweep.
- If NOT found (unknown model id, none offered, or two rows would match): NO
  keystrokes at all -- log + `sendAlert`, dialog left parked for a human.

FABLEFALL1 stays as the second net, unchanged, for the shape it measured.

### Why

A model switch that happens as the side effect of a dialog auto-dismissal is
the worst kind of configuration change: nobody chose it, nothing logged it,
and every status surface keeps reporting the old model.

### Evidence / verification

- 11 decision-level tests (`channel-monitor-model-credit-dialog.test.ts`) --
  including "the no-option shape escalates instead of pressing 1", against a
  fixture verified to satisfy the upstream detector -- plus detector unit
  tests in `pane-state.test.ts`. 465 tests green across the 17 files that
  read channel-monitor as source; `tsc` clean.
- The lane integration mirrors the existing recover-mode users
  (channel-monitor reauth/recovery writers), not a new locking scheme.

### Notes for maintainers

- The pre-existing writers in this same recovery loop (the consent dismissal,
  the generic Escape, the first-run gate answers) still write to the pane
  without acquiring the send lane -- PANEWRITERS805 did not enumerate them.
  Deliberately NOT retrofitted here (separate concern, real blast radius);
  flagging it.
- The escalate path re-alerts on the existing menu-recovery debounce (5 min),
  matching the login-picker branch's cadence.
- The accept path can move the session onto a usage-credit-consuming model
  WITHOUT a new human confirmation -- that is inherent to auto-answering a
  billing-adjacent dialog, and worth saying plainly. The mitigations are
  structural: it only ever selects the operator's pre-configured model (never
  a different one, never a purchase row), anything ambiguous escalates with
  zero keystrokes, and a busy send lane fails closed.
- On our fleet, the per-event owner alert on a successful navigation proved
  too noisy and was later folded into a digest; the alert cadence here is a
  reasonable default, not a proven one.
- The escalation is single-phase: the no-option case goes straight to the
  owner alert (`sendAlert`), because that is the only operator channel this
  file has. A staged escalation (a coordinator/supervisor process gets a
  chance to resolve the dialog before the owner is paged) would need an
  inter-agent messaging path that channel-monitor does not currently use;
  flagging the reduction explicitly rather than inventing that path here.

### Changed files

- `src/pane-state.ts` (+139)
- `src/web/channel-monitor.ts` (+78/-2)
- `src/__tests__/channel-monitor-model-credit-dialog.test.ts` (new, 172)
- `src/__tests__/pane-state.test.ts` (+178)

### DIFF
--- a/src/__tests__/channel-monitor-model-credit-dialog.test.ts
+++ b/src/__tests__/channel-monitor-model-credit-dialog.test.ts
@@ -1,0 +1,172 @@
+// The model usage-credit dialog branch in the blocking-menu recovery pass.
+//
+// FABLEFALL1 established that the credit consent dialog must never get the
+// generic recovery Escape: the CLI records Escape as `cancelled` and continues
+// on the FALLBACK model, silently. Its answer is to press option 1, which is
+// correct for the one dialog shape it detects -- there, option 1 IS "Continue
+// with <configured model>".
+//
+// The gap this branch closes is the shape where that assumption does not hold:
+// a dialog whose options do not include the configured model. Pressing 1 there
+// either buys credits or moves the agent onto a model nobody chose, and because
+// the modal then disappears, the pane, the dashboard and agent-config.json all
+// keep naming the configured model. Such a dialog must ESCALATE, not guess.
+//
+// The pure decision is tested behaviourally below. A real tmux interaction
+// cannot be driven from a unit test, so the wiring asserts read the source and
+// lock in the structural invariants -- ordering, the absence of an Escape, and
+// the send-lane discipline. This file deliberately does NOT import
+// channel-monitor: loading it would pull in the live notification sinks.
+
+import { describe, it, expect } from 'vitest'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, join } from 'node:path'
+import {
+  detectsModelConsentDialog,
+  detectsModelCreditDialog,
+  findModelCreditDialogOption,
+} from '../pane-state.js'
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const MONITOR_PATH = join(__dirname, '..', 'web', 'channel-monitor.ts')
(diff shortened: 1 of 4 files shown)


Szotasz:
Szia, es elnezest a hosszu csendert. 27 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

PR #920, opened 2026-08-06
marveen fork convergence: kanban + main-agent detail + dashboard/i18n (re-scoped per split request)

VTolnai:
**Re-scoped 2026-08-09 per the maintainer's split request** ([comment](https://github.com/Szotasz/marveen/pull/920#issuecomment-review)) — this PR is now only group 5, the independently reviewable remainder. The branch was force-pushed; the old 42-commit bundle is gone.

**Contents (17 commits, cherry-picked fresh onto v1.31.0 main):**
- kanban (8): owner comment notify, drive-by assignee protection, card blockers/wake-on-release, info-emoji nudges, seq-shaped refs on comment path, subtask lane fix, all-digit card id, assignee canonicalisation
- main-agent detail / voice (4): case-insensitive detail route + voice-config, truthful main-agent detail, dedicated opener voice/profile
- dashboard / i18n (5): diagnostics banner (+ opt-in debug flag + self-clearing), dark-mode token-chart contrast, approvals 'decided' label

**Deviations from the requested 19** (found while splitting, details in the comment below): the two Hungarian fx-label fixes depend on i18n keys that only exist on #833's branch, so they were moved there instead of riding here.

**Split siblings:** #939 (egress-gate), #940 (delivery-path, offered as the third candidate for your decide-between-three), #941 (support-mail From split). The sixteen parked costops/token-usage commits are dropped as requested — they live on as #762, #806–#833.

**Verification on this branch:** `tsc --noEmit` clean; `node --check web/app.js web/sw.js` clean; full `vitest run`: 3534 passed, 1 skipped, 1 failed — the failure is `installer-start-and-fallback` ERR-trap line attribution (`TRAP:5` vs `TRAP:6`), which fails identically on pristine upstream/main v1.31.0 in the same environment (bash 5.2.21), so it is environmental, not introduced here.


Szotasz:
Thanks for the curation work, and for being explicit about what you left out and how you resolved the conflicts. That part is genuinely useful and it is why this comment can be specific instead of vague.

I am not asking for a split because 42 commits is a big number. I am asking because I measured what is inside, and the bundle crosses a line that is not a review question on our side.

**What I measured.** I listed the commits on `4fe1d84e` and matched their headlines against our open PR queue. Sixteen of them are already open here as individual PRs, all still `OPEN`:

| commit | open as |
|---|---|
| `3c31ddcc` approvals: card decision requests | #762 |
| `3d0e40f5` costops: list-price equivalent | #806 |
| `3259c35d` token-usage: attribute to scheduled task | #807 |
| `e53cc604` costops: eco mode | #808 |
| `7e31739a` costops: context vs own work | #809 |
| `5dc31e28` token-usage: task_title writer | #810 |
| `d19d6d74` costops: quota guard 70/85 | #811 |
| `e6b78cf5` costops: eco-worker context ceiling | #812 |
| `27079997` costops: actually run the ceiling watch | #813 |
| `540be4ad` costops: shadow-evaluate the first run | #814 |
| `852d796b` scheduler: runner alerts to the main agent | #817 |
| `36d3f2cc` costops: per-project cost | #823 |
| `68d334d5` costops: shadow evaluator measures what the fleet writes | #824 |
| `fbe3d656` costops: one command for a shadow review | #827 |
| `c935b39e` costops: fetch FX rates | #831 |
| `a9d0cfbf` costops: USD or EUR | #833 |

**Why that matters, and it is our problem not yours.** That costops cluster is not waiting on review. It is waiting on a product-boundary decision by the repo owner about whether cost operations belong in the product at all, open on our side since 2026-08-02 and still unanswered. Merging #920 would answer that question silently, by merge, which is exactly the thing we do not let a merge do. So the blocker is ours, and no amount of rebasing on your side moves it.

**What I would like instead.** Please separate along that fault line rather than into six neat topics:

1. **Drop the sixteen commits above from this PR.** They already exist as #762 and #806-#833 and they are parked behind the owner decision. Keeping them here does not speed them up, it only makes the rest unmergeable.
2. **Split the egress-gate work into its own PR** (`7eab924f` payload-field recording, `2b4403bb` quarantine tier). This touches our outbound security boundary, so it gets its own verify pass and never rides along with feature work.
3. **Split the delivery-path commits into their own PR** (`63a8b51a` router session-stuck escalation, `39a86a8d` stop alerting on working sessions, `4fe1d84e` scheduler catch-up summary alert). Heads-up so this is not a surprise: we currently have a standing hold on new delivery-path changes, and there are already two other in-flight designs for the same visibility gap, one external PR (#899) and one of our own cards. Whichever form we keep, we will keep exactly one. This PR would be the third, so it needs to be decided against the other two rather than merged next to them.
4. **Split `cde64a10` (support-mail login vs From split) out on its own.** Changing which identity outbound mail authenticates as and which it sends as is a configuration-surface decision for every install, not a refactor.
5. **The rest can stay together**: kanban (`267917a5`, `49e2a498`, `402e2176`, `60df1002`, `aea45ab7`, `c5f5805d`, `927f7f2a`, `31e3f1ab`), the main-agent detail and voice-config fixes (`3ea4091d`, `6bc3a8ad`, `5c6030f5`, `78265cd5`), and the dashboard/i18n items (`8e72991c`, `562bcf5c`, `5716618a`, `88529c00`, `3040e1ed`, `5a6ad1bd`, `e9cb2831`). That is a reviewable PR and the one most likely to land soonest.

**If you would rather not do the rebase work**, say so and we will carry the split ourselves from your branch. You have already done the hard part, which was resolving the conflicts. What I cannot do is review the merged shape as one decision.

One thing I have not measured and am not claiming: I have not verified the test-count delta or the individual conflict resolutions you describe. That verify happens per split PR, not on the bundle.

VTolnai:
Done, and thank you for measuring instead of gesturing — the table made this mechanical. Split exactly along your fault lines:

1. **The sixteen parked commits are dropped.** This branch was force-pushed and now contains only group 5 (17 commits, fresh cherry-picks onto v1.31.0).
2. **Egress-gate** → #939.
3. **Delivery-path** → #940, explicitly offered as the third candidate in your decide-between-three (vs #899 and your card). If another design wins, close it without ceremony.
4. **Support-mail From split** → #941.

Four things the split surfaced that your table couldn't see from the headlines:

- **`4fe1d84` cannot stand in the delivery-path PR.** It is a two-line follow-up whose function (`recordSchedulerAlert`) and test file both originate in `852d796` (#817). It also needs current main's `ownerChat` rename, so it doesn't apply to #817's older base either — it only exists meaningfully on "#817 rebased onto v1.31". I left it out of #940 and am noting it here: when #817 is unparked and rebased, that two-liner needs to come back (catch-up summary alert must record like the other two sends).
- **The 42nd commit** (your table listed 41): `fe33477`, a fix making the quota guard prove it ran. It belongs to the parked cluster — appended to #811's branch, where it is now included.
- **The two Hungarian fx-label fixes** (`5a6ad1b`, `e9cb283`) depend on i18n keys introduced by the parked FX work — upstream main has no `costs.fx_converted_at` at all. Moved onto #833's branch (clean cherry-picks, pushed), so the label is already correct if/when that PR lands.
- **`402e217` (card blockers) contexted against the card-decisions code** from `3c31ddc` (#762) in `web/app.js`; resolved by keeping only its own additions. Worth a glance at that file in the #920 diff since it was a manual resolution.

**Verification, per branch, not inherited from the bundle:** `tsc --noEmit` clean on all four; full `vitest run` green on all four except one pre-existing failure (`installer-start-and-fallback` ERR-trap `TRAP:5` vs `TRAP:6`) which fails identically on pristine upstream/main v1.31.0 in the same environment (bash 5.2.21) — control run done before claiming that.

What I do not claim: I have not re-verified the parked PRs beyond the two appends compiling/passing on their own bases (#811 full suite green; #833 tsc + syntax clean).

Szotasz:
Hi, and sorry for the silence that came before this.

You closed all nineteen of your PRs on August 26. I only noticed today, while cleaning up cards on our side that were still waiting on you, and I went back to measure what actually happened rather than guess.

Here is what I found, including the part that is on us.

**Where we did the job.** On #920 we asked you to split the bundle, with a measured argument. You did it two days later, dropped the sixteen parked commits yourself, and opened #939 and #940. We merged #940 the same day and #939 two days later. That exchange worked, and your "thank you for measuring instead of gesturing" is the kind of feedback that tells us the argument landed.

**Where we did not.** Of the nineteen PRs, eight never received a single comment from us. Not once, in twenty five days. Nine more got exactly one, usually a note that the branch was conflicting and the rebase stayed with you. That is not a review, that is a holding pattern, and you had no way to tell the difference between "we are thinking about it" and "nobody is looking".

I am not going to guess why you closed them. But if the reason was that half your work sat untouched for most of a month, that is a fair reason, and the fault is ours.

**Two things follow from this.**

The process gap is real and we are fixing it rather than promising to try harder. Our PR watcher has been disabled since mid August because it kept hanging, and the backlog you ran into is downstream of that. The owner asked for it to be rebuilt properly, with someone actually accountable for it, and that work is starting now.

And if there is anything in that set you still want landed, say the word and I will treat it as a first class item, not backlog. You already showed you will do the work when the ask is specific, so the least we owe you is a specific answer within days rather than weeks.

Either way, thank you for the contributions, and for the way you handled the split.

PR #922, opened 2026-08-06
fix(install): report zstd version in the dependency check summary

arnoldvakaria:
zstd is installed as part of the missing-package loop but was never echoed back with the rest of the dependency versions.

- [x ] Bug fix (hibajavítás / bug fix)
- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

## Rövid leírás / Summary

     HU: 

Hiányzó zstd és unzip csomagok pótlása a Linux telepítőben

Friss Ubuntu/WSL image-eken sem az unzip, sem a zstd nincs alapból telepítve. Emiatt a telepítés két ponton is elszállt:
- a Bun hivatalos telepítője unzip-pel csomagolja ki a binárist → error: unzip is required to install bun
- az Ollama telepítője zstd-vel csomagolja ki a binárist → ERROR: This version requires zstd for extraction

A javítás mindkét csomagot felveszi a kezdeti függőség-ellenőrző listába, és emellett közvetlenül a Bun-, illetve Ollama-telepítés előtt is biztosítja a jelenlétüket (apt-get install, hibatűrő módon) — így akkor is működik, ha valaki a csomaglistát megkerülve, önállóan futtatja azt a szakaszt. A függőség-összegzőbe is bekerült a zstd verziószáma, konzisztensen a többi csomaggal.

     EN: 

Add missing zstd and unzip packages to the Linux installer

Neither unzip nor zstd ship by default on a fresh Ubuntu/WSL image, which broke the install at two separate points:
- Bun's official installer unpacks its binary with unzip → error: unzip is required to install bun
- Ollama's installer unpacks its binary with zstd → ERROR: This version requires zstd for extraction

The fix adds both packages to the upfront dependency check, and also ensures their presence right before the Bun and Ollama install steps respectively (apt-get install, fault-tolerant) — so it still works even if that section runs independently of the package-list check. zstd's version is now also echoed in the dependency summary, consistent with the other packages.

Closes #

## Ellenőrzőlista / Checklist

- [ x] Kipróbáltam a módosítást a lokális környezetben (Telegram/Slack + dashboard), az ágensek hiba nélkül kommunikálnak. / Tested locally (Telegram/Slack + dashboard); the agents communicate without errors.
- [x ] A kód nem tartalmaz beleégetett szenzitív adatot (API kulcs, token, személyes adat). / No hardcoded secrets (API keys, tokens, personal data).
- [x ] Saját, beszédes nevű branch-ről nyitom (nem közvetlenül `develop`-ra). / Opened from an own, descriptively named branch (not directly on `develop`).

## Tesztelés

- hianyzo zstd miatt kiakad WSL friss Ubuntu alatt

linux_install_missing_zstd_in_WSL

- javitott verzio

linux_install_fixed_zstd_in_WSL_01
linux_install_fixed_zstd_in_WSL_02

### DIFF
--- a/install-linux.sh
+++ b/install-linux.sh
@@ -357,7 +357,7 @@ if command -v free &>/dev/null; then
 fi
 
 MISSING_PKGS=""
-for pkg in ffmpeg git tmux lsof curl python3 pipx unzip; do
+for pkg in ffmpeg git tmux lsof curl python3 pipx unzip zstd; do
   if ! command -v "$pkg" &>/dev/null; then
     MISSING_PKGS="$MISSING_PKGS $pkg"
   fi
@@ -427,7 +427,7 @@ if [ -n "$MISSING_PKGS" ]; then
     # dnf/yum (Fedora/Nobara/RHEL). A disztro nodejs csomagja v20+ az aktualis
     # kiadasokon, es az npm-et is tartalmazza -- nincs szukseg kulso repora.
     # Csomagnevek megegyeznek a Debian-belivel (ffmpeg/git/tmux/lsof/curl/
-    # python3/pipx/unzip/nodejs). Az ffmpeg-hez Fedoran az RPM Fusion repo
+    # python3/pipx/unzip/zstd/nodejs). Az ffmpeg-hez Fedoran az RPM Fusion repo
     # kellhet; ha mar engedelyezve van, a csomag elerheto.
     # shellcheck disable=SC2086
     pkg_install_noninteractive $MISSING_PKGS
@@ -457,6 +457,7 @@ ok "pipx" $(pipx --version)
 ok "python3 $(python3 --version | awk '{print $2}')"
 ok "tmux $(tmux -V | awk '{print $2}')"
 ok "unzip" $(unzip -v | awk 'NR==1 {print $2}')
+ok "zstd $(zstd --version | awk '{gsub(",","",$5); print $5}')"
 
 # ─────────────────────────────────────────────
 # Repo bootstrap
@@ -602,6 +603,11 @@ export PATH="$BUN_INSTALL/bin:$PATH"
 if command -v bun &>/dev/null; then
   ok "bun mar telepitve: $(bun --version)"
 else
+  # unzip -- a Bun hivatalos telepitoje ezzel csomagolja ki a binarist; nincs alapertelmezetten telepitve friss Ubuntu
+  # WSL kepen, nelkule "error: unzip is required to install bun"-nal elhasal.
+  if ! command -v unzip &>/dev/null; then
+    sudo apt-get install -y unzip 2>/dev/null || true
+  fi
   echo -e "  Bun telepitese (Telegram plugin fuggoseg)..."
   curl -fsSL https://bun.sh/install | bash 2>/dev/null
   if ! command -v bun &>/dev/null; then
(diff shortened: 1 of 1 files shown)


Szotasz:
Thanks -- this is a real bug with a real reproduction, and the two screenshots showing before and after on a fresh WSL image are exactly the evidence that makes a review short. The diagnosis is right: neither `unzip` nor `zstd` ships on a fresh Ubuntu image, and both Bun's and Ollama's installers need them to unpack.

Two items, one of them substantive.

**1. The two new fallbacks bypass this file's own package-manager abstraction, and they are Debian-only.**

`install-linux.sh` supports dnf/yum as well as apt-get (the detection is at the top, and it fails hard on anything else). More to the point, the file states an invariant about itself just above `pkg_install_noninteractive()`:

> MINDEN apt-get hivas ezen a fuggvenyen megy at. Egy hely, ahol a fenti harom...

Both new blocks call `sudo apt-get install -y` directly:

```sh
if ! command -v unzip &>/dev/null; then
  sudo apt-get install -y unzip 2>/dev/null || true
fi
```

On Fedora/Nobara/RHEL there is no `apt-get`, so the command fails, `2>/dev/null || true` swallows it, and the install then dies at the original error with no explanation. That is not a regression -- it fails the same way today -- but it is precisely the case this belt-and-braces check exists for, and your PR description says it should work "even if that section runs independently of the package-list check". On dnf systems it does not.

Please route both through the existing helper:

```sh
command -v unzip &>/dev/null || pkg_install_noninteractive unzip
command -v zstd  &>/dev/null || pkg_install_noninteractive zstd
```

That keeps the file's one-place invariant intact and makes the fallback true on every distro the installer claims to support. Worth a glance at the helper's own error handling before you swap it in, so the `|| true` tolerance you intended is preserved.

**2. The version parse works, but it is tied to one output format.** I checked rather than guessed -- on zstd 1.5.7 your awk gives the right answer:

```
$ zstd --version
*** Zstandard CLI (64-bit) v1.5.7, by Yann Collet ***
$ zstd --version | awk '{gsub(",","",$5); print $5}'
v1.5.7
```

So this is a suggestion, not a defect. Older zstd releases printed a differently worded banner (`zstd command line interface 64-bits vX.Y.Z`), where field 5 is not the version, and Ubuntu LTS images can carry those. Since this line is only the summary echo, a wrong string is cosmetic -- but a format-independent parse costs nothing:

```sh
ok "zstd $(zstd --version | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+' | head -1)"
```

I have not verified the old banner myself, so treat that half as a possibility rather than a finding.

**One question I could not settle, and am not asking you to fix here.** `install-macos.sh` falls back to the same Ollama script (`brew install ollama || curl -fsSL https://ollama.com/install.sh | sh`), so the same extraction step exists on that side. Whether a stock macOS always has `zstd` on PATH I did not measure, so I am not claiming a gap -- just noting that the mirror side of this deserves its own look, on our side, not yours.

Push the helper change and this goes to a verify pass on the delivery chain. Installer changes do not get merged without one here, which is about the file, not about your patch.

arnoldvakaria:
Thank you for the review! I made the changes and tested them. Currently I do not have a configured Mac device (but later on I would like to test on it - so then I could solve the mac update what you mentioned)

Here are the installations logs:

linux_install_fixed_zstd_in_WSL_03
linux_install_fixed_zstd_in_WSL_04
linux_install_fixed_zstd_in_WSL_05

Szotasz:
Verify lefutott a cda25c3a fejen (Samu). Mindharom pont rendben: (1) a pkg_install_noninteractive fallback || true-val nem allitja meg a telepitot, ha a csomag nem elerheto; (2) a dnf/yum ag ugyanazon a helperen fut, mint a fo fuggoseg-telepites, az egy-hely invarians ep; (3) mindket uj hivas idoben a bun/ollama telepito elott all. A verzio-parse a regi zstd bannert is jol olvassa. Koszonjuk a gyors javitast -- a merge kovetkezik.

arnoldvakaria:
Én köszönöm :)

PR #937, opened 2026-08-08
fix(stuck-modal-guard): add 'limited' state for session quota exhaustion

mogganhun:
## Summary

- `classify_pane` misclassified a quota-limit pane as `stuck`: the usage-limit banner hides the idle footer and has no live turn — the same surface signature as a modal. The guard queued Escape attempts and a respawn against a pane that only needed to wait for the quota reset (bigme incident 2026-08-08 15:35).
- New `limited` state in `classify_pane`, checked after `busy` and before `idle`. Matched via the observed banner text (`hit your session limit`, bigme 2026-08-08) plus the canonical variants from `src/model-fallback.ts` USAGE_LIMIT_RX.
- `decide_action`: `limited → hold` — Escape and respawn never fire.
- `alert_limited()`: rate-limited (1h) owner notification with the real reason and reset time extracted from the pane, replacing the generic `/mcp modal` message.
- `run_guard`: calls `alert_limited` when `state == limited`, before `decide_action`.
- Test section (g): quota-pane fixture (observed text + 5-hour variant) + `decide limited → hold` (two firstseen values). 23/23 tests pass.
- `slarti-rendszer-ellenorzes/SKILL.md`: §0 added — check for quota line on pane before guessing modal.

## Test plan

- [ ] `bash scripts/__tests__/stuck-modal-guard.test.sh` — all 23 pass
- [ ] Verify `classify` mode: `echo "You hit your session limit · resets 5:50pm" | bash scripts/stuck-modal-guard.sh classify` → `limited`
- [ ] Verify `decide limited 0 1000` → `hold`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #938, opened 2026-08-08
feat(context-restart-gate): proactive /clear gate with fail-closed live-work detection

mogganhun:
## Summary

- New module: `src/context-restart-gate.ts` -- pure, 0-model-token gate logic for proactive `/clear`
- Gate fires at 400k context tokens (absolute, configurable); FAIL-CLOSED on every unmeasurable signal
- Six conditions block the clear: pane not idle, usage-limit banner (reuses #937), hard guard interlock, live child processes, pending outbound agent_messages, open ledger question, live task state
- `/clear` sent via `withSessionSendLock` (send lane, per PANEWRITERS805)
- Persistent-block alert to bigme after 2h continuous blocking
- 32 unit tests; 0 TypeScript errors; existing context-guard and pane-writers tests unaffected

## What this is NOT

- Not wired to `store/context-restart-gate.json` or `settings.json` yet -- opt-in per agent, separate Janos approval step
- Not a replacement for the hard context-guard (actPct 90% / hardPct 97%) -- complements it by acting much earlier

## Verified scenarios (bigme's requirement)

**GATE BLOCKS** -- `decideGate` test: context 400k, pane idle, but `pendingOutboundCount: 1` (dispatched sub-agent work):
```
action: 'block', reason: 'pending-outbound-messages (1 within 2h cutoff)'
```

**GATE OPENS** -- `decideGate` test: context 400k, pane idle, no children, no outbound, no open question, no task state:
```
action: 'allow', reason: 'context 400000 >= 400000, all gate conditions clear'
```

## Known limitation (documented in runner)

Child-process detection walks `pane_pid → (shell →) claude → children`. Short-lived `exec()` children are filtered by `CHILD_MIN_AGE_S=3`. Very short background Bash (< 3s) is missed; the 5-min retry window catches it unless it finishes by then. `null` from `ps` failure → fail-closed.

## Test run

```
✓ src/__tests__/context-restart-gate.test.ts (32 tests)
✓ src/__tests__/context-guard.test.ts (42 tests)
✓ src/__tests__/pane-writers-under-send-lock.test.ts (10 tests)
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/__tests__/context-restart-gate.test.ts
+++ b/src/__tests__/context-restart-gate.test.ts
@@ -1,0 +1,263 @@
+import { describe, it, expect } from 'vitest'
+import {
+  decideGate,
+  DEFAULT_THRESHOLD_TOKENS,
+  DEFAULT_STALE_CUTOFF_MS,
+  DEFAULT_PERSISTENT_BLOCK_ALERT_MS,
+  normalizeGateConfig,
+  DEFAULT_GATE_CONFIG,
+  type GateInputs,
+  type GateConfig,
+} from '../context-restart-gate.js'
+
+// Fully-clear inputs: context at threshold, pane idle, no dispatched work, no
+// open question, no task state, hard guard idle, child processes measured false.
+// decideGate on these inputs with firstBlockedAt=null MUST return 'allow'.
+const NOW = 1_700_000_000_000
+const CLEAR_INPUTS: GateInputs = {
+  nowMs:                NOW,
+  contextTokens:        DEFAULT_THRESHOLD_TOKENS,
+  paneState:            'idle',
+  paneUsageLimited:     false,
+  hardGuardPhase:       'idle',
+  pendingOutboundCount: 0,
+  hasStaleOutbound:     false,
+  hasChildProcesses:    false,
+  hasOpenQuestion:      false,
+  hasLiveTaskState:     false,
+}
+const ENABLED: GateConfig = { ...DEFAULT_GATE_CONFIG, enabled: true }
+
+function decide(inputs: Partial<GateInputs>, firstBlockedAt: number | null = null, cfg = ENABLED) {
+  return decideGate({ ...CLEAR_INPUTS, ...inputs }, cfg, firstBlockedAt)
+}
+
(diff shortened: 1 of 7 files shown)


mogganhun (commented):
Reviewed and verified on my side (bigme). GitHub will not let me press Approve on a PR from my own account, so this comment is the sign-off.

Three rounds, and every finding came from measuring this host's live process tree rather than from the PR's own tests. The tests were green in all three rounds while the feature was broken, because the fixtures modelled the assumed process shape, not the real one.

What was found and fixed:

1. The original child-process check would have returned true forever: the MCP servers (`npm exec gmail-mcp-server@1.0.30` and the telegram plugin's `bun run --cwd .../plugins/cache/...`) are direct children of the claude process and live for the whole session. The gate would never have opened.
2. The first fix assumed the pane PID is always the claude process. Measured on all eight live sessions: true for six, but `bigme-worker` and `bigme-worker-fast` have bash on the pane with claude as its child. Treating the pane PID as claude there made the real claude look like infrastructure, which flips the failure into a false allow (a /clear on a session that is working). Now resolved by comm.
3. The 0.85 age ratio loosened with session length: at 6890s it classified anything older than 5857s as infrastructure, including a long-running subagent. Measured reality is a 1-3 second gap, so an absolute 60s delta is both generous and stable.
4. Age alone was not sufficient, because this system restarts channel plugin servers mid-session (`channel-mcp-reconnect.ts`). A reconnected server is a young child and would have blocked the gate permanently. Now covered by command-pattern detection.

Final verification against the live tree: `npm exec gmail-mcp-server@1.0.30` matches the package name extracted from .mcp.json, the telegram plugin matches on /plugins/cache/, and my own Bash tool child (a shell-snapshot invocation) matches neither, so it correctly counts as work.

One design correction was mine, not the author's: I originally specified that an existing task-state record must block. That record is precisely what survives a restart, so it signals an open thread, not work in flight. Narrowed to a 10-minute freshness window.

Not wired into settings.json. Enabling it live needs the owner's approval.

PR #960, opened 2026-08-12
fix(context-restart-gate): completion reports are not dispatched work

mogganhun:
## The problem

The gate treats an agent's pending/delivered outbound messages as work in flight and will not open while any exist. Closing an inbound message auto-creates an `[Eredmény] msg_id:<n> status:<s>` message back to the sender (the `PUT /api/messages/:id` route), and `getDispatchedPendingStats` counted those as well.

Nobody answers a completion report, so it never reaches `done`: it sits at `delivered` forever. An agent that diligently closes its inbox therefore builds up **permanent** blockers against its own soft restart.

Observed on 2026-08-12 on the main agent at 596k context: the gate reported

> A(z) "bigme" agens kapuja 121 perce folyamatosan blokkolt. Ok: pending-outbound-messages (11 within 2h cutoff) ... ellenorizd hogy nincs-e elakadt munka.

Several of those 11 were the agent's own acknowledgements. There was no stuck work to find, but the alert asked the owner to go looking for it, and the restart the gate exists to perform stayed blocked.

## The fix

`getDispatchedPendingStats` excludes messages whose content starts with the completion-report prefix, via a bound parameter rather than interpolation.

The prefix becomes an exported `COMPLETION_REPORT_PREFIX`. The notification writer already used the same literal (to avoid ping-pong when a delegator writes back), so both sides now reference one constant and cannot drift apart.

Stale counting is unchanged otherwise: a genuinely dispatched task still blocks, and `hasStale` still reports old dispatches beyond the cutoff.

## Verification

- `tsc --noEmit` clean.
- `db.test.ts`: 30 passing, including two new cases. A real dispatched task counts; adding a completion report leaves the count unchanged; the query stays scoped to the agent's own outbound.
- Built in an isolated worktree off `origin/develop`.

Follows #959, which fixed the other half of the same family: a blocking clock that outlived its session.

### DIFF
--- a/src/__tests__/db.test.ts
+++ b/src/__tests__/db.test.ts
@@ -21,6 +21,9 @@ import {
   deletePendingTaskRetryById,
   markPendingTaskRetryAlert,
   clearPendingTaskRetryAlert,
+  createAgentMessage,
+  getDispatchedPendingStats,
+  COMPLETION_REPORT_PREFIX,
 } from '../db.js'
 import { DB_FILENAME } from '../config.js'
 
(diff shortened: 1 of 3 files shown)


Szotasz:
Independent verify at the pinned head `6f7648a2` (fleet reviewer):

- **Fix confirmed at the source**: `getDispatchedPendingStats` has exactly one consumer (`context-restart-gate-runner.ts:400`), so the exclusion lands precisely where the false block originated. The bound-parameter `NOT LIKE` with the shared `COMPLETION_REPORT_PREFIX` constant (also reused by the route's ping-pong guard) keeps producer and filter on one definition.
- **Tests run and bite**: `db.test.ts` 28 → 30 (both new tests execute). Two semantic mutations, both red: disabling the exclusion (`NEVER_MATCH%`) fails the exclusion test; over-broadening it (`%`) fails both tests — so the suite also pins that REAL dispatched work still counts (the fail-closed direction is preserved).
- **Gate-level regression check**: `context-restart-gate.test.ts` + `context-guard.test.ts` 129/129 green at the PR head.

One **non-blocking** nit for a follow-up (not worth a round-trip before merge): in the second test, `expect(getDispatchedPendingStats('stat-a', NOW, CUTOFF).count).toBe(getDispatchedPendingStats('stat-a', NOW, CUTOFF).count)` compares an expression to itself and can never fail. The `stat-c === 1` assertion carries the test; if you touch this file again, replacing the tautology with a captured before/after pair would pin the "inbound messages are not counted" direction it was meant to cover.

Thanks — the failure mode write-up (an agent's own diligence permanently blocking its soft restart) was accurate and reproduced from the measured case.

PR #968, opened 2026-08-14
fix(context-restart-gate): macOS-blind assumptions kept the gate shut

Vlbbtabs:
On a macOS install the context restart gate could never fire. Six separate places assume GNU/procps behaviour and get BSD `ps` instead, and every one of them fails closed, so a session grows past its threshold with nothing watching it. Found on a live Homebrew macOS install, where an agent had grown to ~500k tokens with the gate reporting itself as enabled.

## The six macOS assumptions

- **`ps -o etimes=` is GNU-only.** BSD `ps` answers `keyword not found`, so the age came back `null`, which the gate treats as "cannot prove this is old" and blocks. Switched to `etime` (`[[dd-]hh:]mm:ss`), which exists on both platforms, parsed by `parseEtimeSeconds` (exported for tests).
- **`ps --ppid` is procps-only.** BSD `ps` rejects the flag outright.
- **`ps -o comm=` prints the bare command name on Linux but the FULL PATH on macOS**, so the equality check against `"claude"` never matched a Homebrew install.
- **tmux is not at `/usr/bin/tmux`** on a Homebrew macOS box, it is `/opt/homebrew/bin/tmux`.
- **MCP child processes were matched against `.mcp.json` patterns only.** A live child whose config entry had since been removed matched nothing, and because a reconnected server is younger than the age filter, it stayed classified as work forever. Replaced with a config-independent shape rule (`isMcpProcess`).
- **An agent with no gate entry of its own inherited `enabled: false`** and so got no gate at all. Added a `_default` fallback key (`DEFAULT_CONFIG_KEY`); an agent-specific entry still wins over it.

## Two counting fixes in the same area

- **Self-addressed messages counted as outstanding dispatched work.** An alert an agent writes to itself is not delegated work and can never come back and clear, so the gate deadlocked against its own persistent-block alert and re-sent it every interval.
- **A `/clear` sent to a session that is mid-turn queues behind the running turn** instead of clearing anything, and the gate recorded a restart that never happened. The runner now proves the session is actually idle first.

## Diagnostic

`scripts/context-restart-gate-doctor.mjs` is read-only: it prints, per agent, why the gate is or is not firing (config, context size, blocking messages, pane state). It never sends a `/clear`.

## Tests

90 tests in `context-restart-gate.test.ts` and 38 in `db.test.ts`, including BSD and GNU `ps` output fixtures for `parseEtimeSeconds` and the `isMcpProcess` shape rule. Full suite on this branch against current `develop`: **3645 passed**, typecheck clean.

### DIFF
--- a/scripts/context-restart-gate-doctor.mjs
+++ b/scripts/context-restart-gate-doctor.mjs
@@ -1,0 +1,129 @@
+#!/usr/bin/env node
+// Context-restart gate doctor: prints, for every agent in the fleet, exactly
+// what the gate sweep sees and what it would decide RIGHT NOW. Read-only --
+// it never sends /clear and never writes gate state.
+//
+// It goes through the runner's own gatherGateInputs()/diagnoseAgent(), so this
+// is the real decision path, not a re-implementation that can drift.
+//
+// Usage:  node scripts/context-restart-gate-doctor.mjs [--json]
+//
+// Exit code 0 always: this is a diagnostic, not a check that can "fail".
+// Read the COVERAGE section -- an agent listed there is one the gate cannot
+// help, which is how a session quietly grows to half a million tokens.
+
+import { execFileSync } from 'node:child_process'
+import { existsSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, join } from 'node:path'
+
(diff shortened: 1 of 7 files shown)


Szotasz:

PR #974, opened 2026-08-14
fix(taskstate): replay on source=clear -- the gate's own restart path dropped it

mogganhun:
## What

`REPLAY_SOURCES` in `src/web/agent-taskstate.ts` did not include `clear`, so a restart performed by the context-restart gate never replayed the task-state record.

The gate restarts an agent by sending `/clear` (`context-restart-gate-runner.ts`), and the harness reports that to SessionStart hooks as `source=clear`. So the restart path the fleet uses in normal operation was the one path that dropped the handover.

## Why it is easy to miss

The other two SessionStart hooks (ledger-replay, daily-log digest) **are** registered for `clear`. The fresh session therefore comes up with chat history and a daily-log digest and looks healthy -- it simply has no done-steps list and no already-delegated list, which is precisely the state that invites re-running or re-delegating finished work.

Observed 2026-08-14 on a live fleet: an agent restarted mid-thread returned without its handover record. The record was intact on disk; the source filter withheld it.

## Also in this PR

`templates/settings.json.template` still carried `"matcher": "compact|resume"`, i.e. narrower than what the server serves since #782. Every agent scaffolded from it was missing `startup` replay as well, so crash respawns lost their record too. Measured on this host: six scaffolded agents had been running that way for 15 days.

## Scope

Nothing else loosens: an unconsumed, non-empty, within-TTL record is still required, and the `consumed` flag still guarantees a single replay. A planned `/clear` right after finishing work replays nothing.

## Verification

- `src/__tests__/agent-taskstate.test.ts`: 21 pass, 3 new (clear replays; a consumed record on clear does not; an unknown source still does not)
- `tsc --noEmit` clean; template parses as JSON
- Live check against a running dashboard: `source=clear|startup|compact` return the injection, an unknown source returns `null`

## Behaviour change (stated explicitly, per review)

After this PR a manual `/clear` behaves differently than it does today. Today a
`/clear` starts the fresh session empty. After this, `clear` is a replay source
like `compact` and `resume`, so an unconsumed, non-empty, within-TTL record is
injected into the fresh session as a task-continuation block.

That is the point of the change -- the context-restart gate restarts by sending
`/clear` -- but it also applies to a `/clear` a human types. Anyone deliberately
clearing to start clean will see the previous task's done-steps and next-action
injected. The `consumed` flag still limits that to one replay, and a record
written after the work finished is empty and replays nothing.

## Matcher convergence (added after review)

The template fix alone reaches new installs only. `ensureAgentHooks` adds missing
commands and syncs timeouts, but never touches the matcher, so an install keeps
the matcher it was created with. Widening the template therefore looked complete
while every running install kept dropping the replay.

`syncHookMatchers` converges a hook group **only when every command in it comes
from the template entry**. A group that also holds a hook we did not put there
keeps its own matcher, since widening it would silently change when that foreign
hook runs.

Kept in this PR rather than split out: it is the same defect and the fix is inert
without it. Happy to split if you would rather review it separately.

## Coverage test (added after review)

`REPLAY_SOURCES` is now exported and the template matcher is asserted to **cover**
it, per source, not to equal it. An equality assertion passes today and drifts
silently the next time a source is added to one side only -- which is how the
replay went dead in the first place. Three negative cases lock the convergence
rule: idempotent once converged, a group with a foreign command untouched, an
unrelated group untouched.

Verification after the review round: `npm run build` clean, and
`session-start-matcher-sync.test.ts` + `agent-taskstate.test.ts` +
`hook-path-guard.test.ts` = 48 pass.

Note for anyone reproducing: the suite refuses to run in a live install and must
not run from `/tmp` either -- the hook-path guard rejects `/tmp`-prefixed roots,
so the gate tests go falsely red there. A worktree under `$HOME` is green.

### DIFF
--- a/src/__tests__/agent-taskstate.test.ts
+++ b/src/__tests__/agent-taskstate.test.ts
@@ -38,8 +38,17 @@ describe('shouldReplayTaskState', () => {
   it('replays on startup too (crash respawn mid-task)', () => {
     expect(shouldReplayTaskState(rec(), 'startup', NOW + 1000)).toBe(true)
   })
+  // The context-restart gate restarts by sending /clear, so source=clear is the
+  // MOST common restart this record exists for (2026-08-14). Pinned: excluding
+  // it silently emptied every gate-initiated restart.
+  it('replays on clear (the context-restart gate restarts with /clear)', () => {
+    expect(shouldReplayTaskState(rec(), 'clear', NOW + 1000)).toBe(true)
+  })
   it('does NOT replay an unknown source', () => {
-    expect(shouldReplayTaskState(rec(), 'clear', NOW + 1000)).toBe(false)
+    expect(shouldReplayTaskState(rec(), 'prompt-input-exit', NOW + 1000)).toBe(false)
+  })
+  it('does NOT replay a consumed record on clear either', () => {
+    expect(shouldReplayTaskState(rec({ consumed: true }), 'clear', NOW + 1000)).toBe(false)
   })
   it('does NOT replay an empty record on startup either', () => {
     const empty = rec({ doneSteps: [], alreadyDelegated: [], nextAction: '', pendingDecision: '', summary: 'idle' })
--- a/src/__tests__/session-start-matcher-sync.test.ts
+++ b/src/__tests__/session-start-matcher-sync.test.ts
@@ -1,0 +1,99 @@
+// Two guards for the SessionStart task-state replay, both aimed at the same
+// failure: the hook is registered, but not for the sources that matter.
+//
+//   (a) Coverage: the template's SessionStart matcher must COVER REPLAY_SOURCES.
+//       Asserted as containment, not equality -- an equality assertion passes
+//       today and drifts silently the next time a source is added to one side
+//       only, which is exactly how the replay went dead the first time.
+//
+//   (b) Convergence: an install created with an older matcher must converge on
+//       the template's matcher on the next scaffold run. Adding a source to the
+//       template alone reaches new installs and never the running ones, so the
+//       fix looks complete while the fleet stays broken.
+import { describe, it, expect } from 'vitest'
+import { readFileSync } from 'node:fs'
+import { join, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+import { REPLAY_SOURCES } from '../web/agent-taskstate.js'
+import { syncHookMatchers } from '../web/agent-scaffold.js'
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const ROOT = join(__dirname, '..', '..')
+const TEMPLATE = join(ROOT, 'templates', 'settings.json.template')
+
+const REPLAY_HOOK = 'taskstate-replay.py'
+
+function templateSessionStartMatcher(): string {
+  // The template carries {{PLACEHOLDER}} tokens that are not valid JSON values
+  // on their own; they sit inside strings, so JSON.parse handles the file fine.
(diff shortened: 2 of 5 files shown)


Szotasz:
Thanks for this one. The direction is right and the diagnosis is correct: the gate's own restart path did drop the replay on `source=clear`, and that is worth fixing.

We are holding the merge on one measured gap, and it is not about the change you made. It is about who the change reaches.

**The template matcher never propagates to installs that already exist.** `ensureAgentHooks` writes hooks for an agent but does not synchronise the matcher, so an install that is already running keeps the matcher it was created with. Your PR fixes the template, which means new installs get the fix and existing ones do not. Concretely: the startup replay that broke on 2026-07-27 is still dead on every running install today, from this same cause, and this PR as written would not revive it.

That makes the fix look complete while the fleet stays broken, which is the worst of both outcomes, because nobody goes back to check a merged PR.

What we would like before merging, all three small:

1. **Matcher sync in `ensureAgentHooks`**, so an existing install converges on the next run instead of staying on its creation-time matcher.
2. **A subset test**: assert that the template matcher covers `REPLAY_SOURCES`. A plain equality test would pass today and drift silently the next time someone adds a source, so please assert containment rather than equality.
3. **Say the behaviour change out loud** in the PR body: after this, a manual `/clear` behaves differently than it does today. We want that written down, not discovered later by whoever debugs it.

No rush on our side and no scope creep expected beyond those three. If you would rather split point 1 into its own PR because it touches a different file, that is fine by us, just say so and we will track them together.

One note so the timing is not a surprise: we have a release going out on Tuesday, and changes to the restart and scheduling paths are merging after it rather than before. So even once this is updated, the merge will land after Tuesday. That is about our release window, not about the quality of the patch.

mogganhun:
Thank you, the matcher gap is a fair catch and it was the more important half of the bug. All three are in, pushed as `0be273a`.

**1. Matcher sync.** `syncHookMatchers` in `agent-scaffold.ts`, called from the existing-event branch of `ensureAgentHooks` alongside the timeout sync. It converges a hook group **only when every command in it comes from the template entry**; a group that also holds a hook we did not put there keeps its own matcher, because widening it would silently change when that foreign hook runs. I kept it in this PR rather than splitting it out, since the template change is inert without it, but say the word and I will split it.

**2. Subset test.** `REPLAY_SOURCES` is now exported, and `session-start-matcher-sync.test.ts` asserts the template matcher **covers** it per source rather than equalling it, exactly for the drift reason you gave. Three negative cases lock the convergence rule: idempotent once converged, a group holding a foreign command untouched, an unrelated group untouched.

**3. Behaviour change.** Written into the PR body under its own heading: after this, a manual `/clear` replays the record where today it starts the fresh session empty, and that applies to a human-typed `/clear` too, not just the gate's.

One measurement that supports your point better than my patch did. On this host all seven installs already carry the full matcher, so the replay is alive here. That is not the converge path working, though. The settings files were rewritten wholesale during unrelated isolated-config work: the shared file's mtime is the same second as the template edit, and the agent files were regenerated two days later. Without that accident this host would still be on the old matcher, which is precisely the case you are describing.

A note for anyone reproducing the suite: it refuses to run in a live install, and it must not run from `/tmp` either, because the hook-path guard rejects `/tmp`-prefixed roots and seven gate tests go falsely red. A worktree under `$HOME` is green: 48 pass across the three affected files, `npm run build` clean.

Understood on the Tuesday release window, no rush from this side.

Szotasz:

Szotasz:
Szia, es elnezest a hosszu csendert. 19 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

PR #976, opened 2026-08-14
feat(context-restart-gate): wake the fresh session after /clear

mogganhun:
## What

After the gate sends `/clear`, it stops. The fresh session has the injected SessionStart context but no reason to speak -- a Claude Code session produces a turn only when something prompts it. So a gate-restarted agent sits mute, and the handover record's "next action", written for exactly this moment, is read by nobody.

Measured 2026-08-14: an agent restarted at 22:24 stayed silent until its owner asked "well?" at 22:47. Nothing crashed, the restart worked, the context arrived. There was simply nothing to trigger the first turn.

This also quietly defeats the gate's own purpose in the unattended case: restarting early is meant to keep the agent working cheaply, and a mute agent does not work.

The hard context-guard already solves this on its own path (`inject-resume` + `resumePrompt`). This is the same idea on the soft path.

## Design

- The "is it due" rule is pure and unit-tested (`decideWake`); the runner only performs delivery.
- The debt is **persisted** (`pendingWakeAt` in the gate run-state), not an in-memory timer, so a dashboard restart between the `/clear` and the nudge still leaves the fresh session woken on the next sweep.
- 25s grace covers boot + SessionStart hooks; delivery additionally waits for an idle pane. A busy pane **defers** to the next sweep rather than typing over live work.
- Past 30 minutes the nudge is **dropped**: "you just restarted" is false half an hour later.
- Negative age (clock skew) reads as `wait`, never `drop`.
- The prompt states explicitly that "no work in flight" is a complete state, so the nudge cannot become a source of invented work.

## Verification

- `src/__tests__/context-restart-gate.test.ts`: 70 pass, 8 new (`decideWake` boundaries incl. clock skew and overrides, plus prompt-content guards)
- `tsc --noEmit` clean
- Live delivery test on a running fleet **without** a real restart: arming `pendingWakeAt` produced the nudge in the agent's session on the next sweep, and the debt cleared -- which the code only does on `outcome === 'sent'`

### DIFF
--- a/src/__tests__/context-restart-gate.test.ts
+++ b/src/__tests__/context-restart-gate.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import {
+  gateWakePrompt,
   isInfrastructureChild,
   findClaudePidInTree,
   extractMcpPackageNames,
@@ -11,7 +12,10 @@ import {
 } from '../web/context-restart-gate-runner.js'
 import {
   decideGate,
+  decideWake,
   nextBlockClock,
+  WAKE_DELAY_MS,
+  WAKE_MAX_AGE_MS,
   DEFAULT_THRESHOLD_TOKENS,
   DEFAULT_STALE_CUTOFF_MS,
   DEFAULT_PERSISTENT_BLOCK_ALERT_MS,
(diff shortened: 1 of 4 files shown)


Szotasz:

PR #978, opened 2026-08-14
fix(isolated-config): keep hooks the isolated dir alone defines

mogganhun:
## What

The target-only-**key** rescue from #861 cannot protect `hooks`. The shared `~/.claude/settings.json` almost always defines that key, so the shared object wins wholesale and every hook that exists **only** in the isolated `settings.json` is dropped -- on every main-agent start, silently. The agent boots fine; the hook simply never runs again.

Measured 2026-08-14 on a live install: a `UserPromptSubmit` hook was registered into `.channels-config/settings.json` at 21:20 and was gone by the next start. Nobody noticed for 26 hours -- during which the operator had been told the feature was live, and a week-long measurement was believed to be running.

Same failure shape as the `statusLine` losses (07-28, 07-30, 08-03) that motivated #861; hooks just fall outside the key-level fix.

## How

Merge per hook **entry**, identified by what it runs (command line, or `prompt` for `agent` hooks):

- shared groups stay first and unchanged -- shared remains the source of truth, matching key-merge semantics
- an entry present in both appears once, in its shared position
- an isolated-only entry is appended to the group with the same `matcher`, or as a new group when there is none
- an event the shared file never mentions comes along whole
- kept entries are logged by command, never silently

## Trade-off (stated, not hidden)

This cannot distinguish "removed from shared" from "never in shared", so an entry deleted upstream keeps running for that agent until it is also removed from the isolated file. Additive is the safe direction here: the failure it replaces is silent loss.

## Verification

- 32 tests pass: 5 new unit tests (`mergeIsolatedHooks`, `hookEntryId`), 2 new provisioning tests, plus the existing `main-agent-isolated-config` / `isolated-channel-config` suites
- `tsc --noEmit` clean
- Live check on the affected install: a hook written only into `.channels-config/settings.json` survived a real provisioning run, and the new log line named it

### DIFF
--- a/.testout
+++ b/.testout
@@ -1,0 +1,439 @@
+
+ RUN  v2.1.9 /home/janos/wt-978
+
+ ✓ src/__tests__/context-guard.test.ts (58 tests) 62ms
+ ✓ src/__tests__/process-lock.test.ts (63 tests) 71ms
+ ✓ src/__tests__/pane-state.test.ts (217 tests) 132ms
+ ✓ src/__tests__/governance-gates.test.ts (76 tests) 78ms
+ ✓ src/__tests__/stuck-tool-call.test.ts (35 tests) 39ms
+ ✓ src/__tests__/prompt-injection-defense.test.ts (58 tests) 37ms
+ ✓ src/__tests__/context-restart-gate.test.ts (62 tests) 51ms
+ ✓ src/__tests__/mcp-list-parser.test.ts (59 tests) 108ms
+ ✓ src/__tests__/token-usage.test.ts (23 tests) 215ms
+ ✓ src/__tests__/remote-enroll-core.test.ts (44 tests) 61ms
+ ✓ src/__tests__/federation-lifecycle.test.ts (18 tests) 165ms
+ ✓ src/__tests__/federation-config.test.ts (27 tests) 74ms
+[23:52:56.143] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:52:56.596] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:52:57.098] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:52:57.762] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:52:58.448] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:52:59.441] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:53:00.406] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:53:01.698] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+ ✓ src/__tests__/fleet-transfer.test.ts (17 tests) 7417ms
+   ✓ encrypt/decrypt round-trip > decrypts to the original plaintext 2145ms
+   ✓ encrypt/decrypt round-trip > throws on wrong password (GCM auth tag mismatch) 1364ms
+   ✓ encrypt/decrypt round-trip > produces a non-trivially-parseable blob that differs from plaintext JSON 644ms
+   ✓ importFleet: encrypted wrapper detection > returns error DiffReport when encrypted but no password given 663ms
+   ✓ importFleet: encrypted wrapper detection > returns error DiffReport on wrong password (no file writes) 1296ms
+   ✓ importFleet: encrypted wrapper detection > succeeds (dry-run) with correct password 1286ms
+ ✓ src/__tests__/bridge-enroll.test.ts (13 tests) 137ms
+ ✓ src/__tests__/template-identity-hygiene.test.ts (10 tests) 196ms
+ ✓ src/__tests__/channel-mcp-reconnect.test.ts (27 tests) 29ms
+[23:53:03.846] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:53:04.565] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+ ✓ src/__tests__/channel-coordinator.test.ts (31 tests) 66ms
+ ✓ src/__tests__/db.test.ts (30 tests) 132ms
+[23:53:06.439] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+[23:53:06.814] [33mWARN[39m (978396): [36mbreak-glass password reset via bearer token[39m
+    [35musername[39m: "alice"
+  prune: python3 /tmp/scratchpad/staleness-guard.py
+boot-hook-prune: pruned 1 stale hook(s) from /tmp/boot-prune-test-h4lOY4/.claude/settings.json (backup: /tmp/boot-prune-test-h4lOY4/.claude/settings.json.bak)
+boot-hook-prune: 1 stale hook(s) removed across 1 file(s)
+  prune: python3 /opt/removed-checkout/scripts/hooks/staleness-guard.py
+boot-hook-prune: pruned 1 stale hook(s) from /tmp/boot-prune-test-56Ings/.claude/settings.json (backup: /tmp/boot-prune-test-56Ings/.claude/settings.json.bak)
+boot-hook-prune: 1 stale hook(s) removed across 1 file(s)
+ ✓ src/__tests__/hook-path-guard.test.ts (18 tests) 279ms
+ ❯ src/__tests__/installer-start-and-fallback.test.ts (26 tests | 1 failed) 10503ms
+   × the ERR-trap abort is real (guards the premise of the tests above) > an unguarded capture aborts, and bash blames the enclosing `fi` 31ms
+     → expected 'TRAP:5' to be 'TRAP:6' // Object.is equality
+   ✓ install-macos.sh -- launchd units must be verified, not assumed > never trusts load/bootstrap alone -- it always kickstarts 2058ms
+   ✓ install-macos.sh -- launchd units must be verified, not assumed > returns the pid once the unit is actually up 2051ms
+   ✓ install-macos.sh -- launchd units must be verified, not assumed > returns NOTHING when the spawn stays pended, so the caller can fail loudly 2060ms
+   ✓ install-macos.sh -- launchd units must be verified, not assumed > returns NOTHING for a crash-looping unit whose first pid was transient 2107ms
+   ✓ install-macos.sh -- launchd units must be verified, not assumed > does not abort the installer when launchd refuses to start the unit 2071ms
+[23:53:07.587] [32mINFO[39m (978396): [36mdashboard user created[39m
+    [35musername[39m: "alice"
+ ✓ src/__tests__/claude-config-dir.test.ts (41 tests) 28ms
+ ✓ src/__tests__/serve-file-cache.test.ts (25 tests) 50ms
+ ✓ src/__tests__/channels-never-started-backoff.test.ts (12 tests) 356ms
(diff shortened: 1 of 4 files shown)


Szotasz:
The bug is real and the direction is right. You also locked the invariant structurally rather than only testing the happy path, which we appreciate and do not see often.

Two things block the merge, and the first one is the serious one.

**1. The hook replacement can leave a zombie that fails closed, permanently.**

Our live governance hook embeds a version-pinned node path and ends in `test -x || exit 2`. That is deliberate: if the interpreter is missing, the hook refuses rather than silently letting the call through. It is safe today because the hook can heal itself, and the way it heals is by telling the operator to restart the dashboard, which rewrites the entry.

With this change, a stale entry retained in an isolated config would no longer be rewritten by that path. So after a routine node upgrade, the old pinned path stops resolving, the hook exits 2, and every `PreToolUse` call for every isolated-config agent is blocked. Not degraded, blocked, with no self-healing left. A version bump on an unrelated component would take out the agents, and the cause would be several steps away from the symptom.

Either of these would close it, whichever you prefer:

- filter out entries whose command is no longer runnable before merging them, or
- supersede by basename, so a newer entry for the same hook replaces the old one instead of both surviving.

Please also add a test that walks the upgrade path: an isolated config holding an entry that points at a path which no longer exists must not survive the merge.

**2. Temp and worktree paths would become permanent.**

`mergeIsolatedHooks` does not check where an entry points. Anything picked up from a temporary directory or a git worktree gets written into the durable config and outlives the directory it came from. Please reject entries outside the expected roots. We are asking for this now because it is much cheaper than cleaning it out of many installs later, and we have exactly that mess in one of our own configs already.

Neither point is about the idea. Keep the change as designed and add the guards.

One note on timing so it is not a surprise: we have a release on Tuesday, and changes on this path are merging after it rather than before. So even once this is updated, the merge lands after Tuesday. That is our release window, not a comment on the patch.

mogganhun:
Both guards are in, pushed as `dbd21c1`. The zombie is the one I had not thought through, and you are right that it is the serious half: retaining an entry is exactly what removes it from the path that heals it.

**1. Stale entries no longer survive the merge.** I did both of the options you offered, because they cover different halves of the same failure:

- `isDurableIsolatedHook` drops an entry if **any** absolute path in its body does not resolve right now. That is the interpreter case directly: after a node upgrade the pinned path is gone, so the entry is dropped rather than carried forward to fail closed forever.
- The merge also **supersedes by basename**: if the shared side already provides a hook whose script basename matches, the isolated variant does not come along, even when both paths resolve. That covers the case where the old path still exists but is no longer the one the install uses.

**2. Transient paths rejected.** A script path outside the install root or the operator's `~/.claude` does not come along. Worth noting for the review: "somewhere under `$HOME`" would not have caught this, since a git worktree lives there too, so the test is the install root specifically.

**Tests walk the upgrade path**, as asked: a pinned interpreter that no longer resolves, a script removed by an upgrade, a `/tmp` root, a worktree copy, and supersede-by-basename with **both** paths resolving so the drop can only come from that rule. The positive case is locked too: a durable isolated-only hook the shared file never mentions is still kept, which is the behaviour this PR exists for.

One deliberate non-reuse: I did not import `isUnsafeHookCommand` from `agent-scaffold`, since that would pull the scaffold's module graph (and `runAgent` with it) into process startup. Say the word if you would rather have the single implementation and take the import.

Verification: `npm run build` clean; `isolated-hook-merge` + `main-agent-isolated-config` 23 pass. Full suite 267 of 269 files pass; the two failures are `memory-performance` (Ollama unreachable on this host) and one bash ERR-trap case in `installer-start-and-fallback`, both of which fail identically with my changes stashed.

Understood on the Tuesday window.

Szotasz:

Szotasz:
Szia, es elnezest a tizenkilenc napos csendert. A verify lefutott a mai developen: **GO, egy feltetellel.**

## A premissza el

A developen ma sincs `hookEntryId`, sincs izolalt-hook osszefesules -- a fogalom sem letezik ott. A leirt tunet (az izolalt fajl SAJAT hookjai minden fo-agens inditasnal eltunnek, es a tunet a CSEND) tehat ma is fennall.

## Ami kulon jo, es ezt eloszor mondom ki

Nem egyszeruen atmasolod az izolalt hookokat, hanem **ket, altalad megmert hibamodot zarsz ki**: a `/tmp`-ben lako atmeneti hookot, es a "zombi, ami fail-closed" esetet.

A masodikat a **mi valos adatunkon** is igazoltuk. Az izolalt fajlunkban pontosan egy ilyen bejegyzes all, ilyen alakban:

```
test -x "/opt/homebrew/Cellar/node@22/22.22.3/bin/node" || ... ; node .../egress-gate.mjs
```

Az ut ma letezik, tehat a predikatum atvinne. Egy `node@22` frissites utan viszont kiesne -- pontosan az a vedelem, amit igersz, es nem elmeletben.

## A feltetel: egy vak folt, megmerve a SZALLITOTT fuggvennyel

Az `isDurableIsolatedHook` a hook torzseben talalt minden abszolut utat letezes-vizsgalja, es a szkript-vegzodesueket tartos gyokerhez koti. Lefuttattuk a **te fuggvenyedet** (nem egy ujraimplementaciot) a gepunk valodi hook-torzsein:

```
KEEP = 35 | DROP = 15   (osszesen 50)
```

Mind a 15 eldobott ugyanaz a csalad:

```
python3 "$CLAUDE_PROJECT_DIR/scripts/hooks/<valami>.py"
```

A regex a valtozo UTANI reszt (`/scripts/hooks/x.py`) abszolut utnak latja, `.py`-ra vegzodik, es nem esik a tartos gyokerek ala, tehat DROP. Literal-utas hook egyetlen egy sem esik ki.

**Es rogton szukitem is, mert kulonben tulzo lenne:** a predikatum CSAK az izolalt fajlbol atvitelre jelolt bejegyzesekre fut, a mi izolalt fajlunk mind a 11 hookja pedig literal utat hasznal -- tehat **ma egyetlen bejegyzes sem esne ki**.

**A keres attol ervenyes, hogy a `$VAR`-stilus nalunk 15-szor fordul elo a 35 literal mellett, es a trackelt template MAGA is igy ir.** Vagyis egy operator teljesen termeszetes modon regisztralna izolalt-only hookot ebben az alakban -- es akkor a guard NEMAN eldobna. Pontosan azt a csendet allitana elo, ami ellen ez a PR keszult.

## Amit kerunk, konkretan

A letezes-vizsgalat elott a torzsben levo `$VAR` / `${VAR}` alakot vagy oldd fel, vagy az ilyen utat hagyd ki a vizsgalatbol -- ne dobjunk el olyasmi miatt, amit statikusan nem tudunk feloldani. A tartos-gyoker ellenorzes maradhat a literal utakra.

## Egy meresi megjegyzes, hogy a szam ertelmezheto legyen

A `durableScriptRoots()` a `PROJECT_ROOT`-bol szarmazik, az pedig `join(__dirname, '..')` -- tehat a MODUL helyebol, nem kornyezeti valtozobol. Egy worktree-ben futtatva emiatt a nyers eredmeny 42 dropot adott; az a worktree artefaktuma, nem a fuggvenye. A fenti 15 az a szam, ami akkor all elo, ha a predikatum ugyanazt a szerkezetet latja, mint elesben.

Ha a `$VAR`-kezelessel megvagy, szolj, es soron kivul ranezek.

---

mogganhun:
Kesz a `$VAR`-kezeles, felkuldve: `eefa79c`.

## Amit valasztottam, es miert nem a masikat

A ket felkinalt ut kozul a **kihagyast** valasztottam, nem a feloldast. A `$CLAUDE_PROJECT_DIR` agensenkent mas es csak a hook FUTASAKOR kotodik, tehat barmit helyettesitenek ide, az egy tippeles lenne, ami bizonyossagnak olvasodik -- es ha rosszul tippelek, megint egy mukodo operator-hookot dobok el nemán. Amit nem latok, azt bekem hagyom, nem itelem el.

A felismeres **pozicionalis**: egy valodi abszolut ut token elejen all, tehat elotte szokoz, idezojel, `=`, `(` vagy a torzs eleje van. Ha a vezeto `/` elott szo-karakter, `}` vagy `)` all, akkor a per-jel mast folytat -- `$HOME`, `${VAR}`, `$(dirname "$0")` -- es az az ut nem literalis.

## A literalis fel nem lett puhabb, es ezt teszt orzi

Ez volt a sajat aggalyom a valtoztatassal szemben: ha a feloldhatatlan felet atengedem, ne lehessen vele a tobbi ellenorzes ala bujni. Harom uj teszt pontosan ezt meri:

- a valtozobol jovo utas bejegyzes atmegy (mind a negy szokasos alakban),
- de **elbukik**, ha ugyanabban a torzsben egy literalisan megnevezett interpreter mar nem letezik (a zombi esete valtozatlanul zar),
- es **elbukik** akkor is, ha egy literalis `/tmp` ut all a valtozos mellett.

## Meres a sajat elo adatunkon

Az izolalt configunkban **15 hook-bejegyzes** van, ebbol **1** all a valtozos alakban. A javitas elott pontosan az az egy esett volna ki -- egy `$CLAUDE_PROJECT_DIR`-rel hivott kimeno-szoveg kapu.

## Amit a futtatasom NEM lat

A teljes suite: 3624 zold, 3 bukas -- mindharom **baseline**, a patch nelkul is bukik ugyanugy (ket embedding-teszt egy elo Ollama idozitesere, es egy bash ERR-trap teszt). Ezt osszehasonlitassal ellenoriztem, nem feltetelezem: `git stash` elotte-utana, ugyanaz az eredmeny. A tipusellenorzes tiszta.

Amit a merésem nem lat: a te configod alakjat. A 15/1 arany a MI telepitesunk, nalatok mas lehet -- ha a te oldaladon a 15 dropbol tobb is valtozos alaku volt, az a szam most csokkenni fog, de ezt csak te tudod megmerni.

Szotasz:
Thanks for this, and I owe you an apology for the delay: you opened it on 14 August and it sat for 33 days. That is on us, not on the PR.

First, the finding is real. You measured a genuine gap: hooks that only the isolated config directory defines were being dropped, and nothing said so. We reproduced it.

What happened in the meantime is the part you could not have seen. We closed the same gap from the other side in #1307, which is already merged, and the two changes take opposite positions on the same question:

- This PR treats the isolated config file as a hook surface: hooks it alone defines should survive an additive merge.
- #1307 treats the isolated config file as *not* a hook surface: it deletes the key and excludes it from inheritance.

Both are defensible, but they cannot both be true in the same code path, and we had to pick one. We went with #1307, for one reason: the isolated directory is provisioned, not authored. Anything written into it by hand is expected to be overwritten on the next provisioning round, so a hook that lives only there is a hook that silently disappears the next time the agent is re-provisioned. Making it survive the merge would have made that surface look durable when it is not.

That said, your PR pointed at something #1307 got wrong, and we are fixing it: the stripping was **silent**. A hook that is neither global nor project-defined now gets logged by name when the provisioner removes it, so the case you hit produces a visible line instead of nothing. That is tracked on our side and lands after the global-prune work.

So: I am closing this PR, not because the bug was not real, but because the fix landed on the other side of the same decision. Concretely, the branch also conflicts with `develop` now, so rebasing it would mean re-litigating the policy rather than resolving text.

If you disagree with the direction, say so here and I will reopen the discussion. And if you hit the case again after the logging change ships, that is a bug report I want.

mogganhun:
Thanks for the detailed write-up, and the delay needs no apology from our side.

We agree with the direction and we are not asking you to reopen it. The reasoning behind #1307 is stronger than ours: the isolated directory is provisioned rather than authored, so a hook that lives only there looks durable and is not. Our PR would have made that appearance permanent, which is the worse failure of the two. And the part that actually cost us time, the silence, is the part you are fixing, so the finding did its job even though the patch did not land.

One practical question, asked as a user of the project rather than as dissent. With the isolated config no longer treated as a hook surface, what is the supported home for a hook that should apply to exactly one agent and not to the whole fleet? Global settings apply everywhere, and the project settings file is shared by every agent working in that repo. We run a small fleet where a single agent occasionally needs its own guard, and we would rather build toward whatever you intend than re-create the same expectation somewhere else.

If the honest answer is that there is no such surface yet and the intended pattern is a project-level hook that scopes itself by agent id inside the script, that is a perfectly usable answer, we just want to know which way to build.

Happy to move this to a separate issue if a closed PR is the wrong place for it.

PR #979, opened 2026-08-15
fix(stuck-tool-call-watcher): an arriving message must not open the gate on stale evidence

mogganhun:
## What breaks

The idle-prompt guard is the only thing holding back a residual `<verb> for Ns` footer, and it stops applying the instant an inbound channel message is injected into the prompt box: `detectPaneState` then reads `typing`, not `idle`.

## Evidence (live install, 2026-08-15, owner-visible)

```
14:52:05  counter frozen at 49s -> skipped, "pane is at the idle prompt"
14:56:05  same -> skipped
15:00:05  same -> skipped
15:03:06  owner's Telegram message lands in the prompt box
15:04:05  recover: "stagnant past freeze threshold + idle wedge profile"
15:04:09  session respawned with --continue
```

The evidence that triggered the respawn predated the message by twelve minutes. The CPU-profile guard did not catch it either, because a freshly-arrived message is exactly the state where the turn has not started and CPU is still low. The respawn took the not-yet-processed message with it; only a separate ledger-drain safety net recovered it. The owner noticed and asked why a 49-second stall caused a restart.

## Fix

1. **Parked-input guard.** A parked channel block is not wedge evidence: it means the session is about to be driven, and that case already belongs to `stuck-input-watcher`, which has its own escalation (Enter -> clear+re-inject -> respawn). Skip recovery and clear the stale spell so the residual cannot re-arm on the next poll. Placed **before** the CPU-profile guard, for the reason above.

2. **Alert wording.** The owner-facing message rendered `lastSeconds` as a duration, which reads as a 49-second hair trigger. That number is the frozen counter value; the acting threshold is `freezeSeconds`. The message now states both.

## Verification

- `tsc --noEmit` clean
- `vitest run src/__tests__/stuck-tool-call.test.ts` -> 37/37 green
- Two new wiring-contract assertions: the guard exists and is ordered before the CPU guard, and the old duration wording is rejected. Anchored on the call site rather than the exported signature, since the latter also matches `confirmsWedgeProfile(cpuPercent` and would make the ordering assertion vacuous.

### DIFF
--- a/src/__tests__/stuck-tool-call.test.ts
+++ b/src/__tests__/stuck-tool-call.test.ts
@@ -360,6 +360,35 @@ describe('stuck-tool-call-watcher wiring contract', () => {
     expect(watcherSrc).toMatch(/confirmsWedgeProfile\(/)
   })
 
+  it('skips recovery while an inbound channel message is parked in the prompt (2026-08-15)', () => {
+    // Owner-observed false positive: the idle-prompt guard is the only thing
+    // holding back a residual footer, and it stops applying the moment an
+    // inbound message is injected (detectPaneState reads 'typing', not 'idle').
+    // Measured that day: counter frozen at 49s and correctly skipped as
+    // residual at 14:52/14:56/15:00; the owner's message landed 15:03:06; at
+    // 15:04:05 the guard no longer applied and the pane was respawned, taking
+    // the not-yet-processed message with it. A parked channel block belongs to
+    // stuck-input-watcher, so this watcher must stand down.
+    expect(watcherSrc).toMatch(/parkedChannelInput\(pane\)\s*!=\s*null/)
+    // Ordering matters: the parked-input guard must be evaluated BEFORE the
+    // CPU-profile guard, otherwise a freshly-arrived message (turn not started,
+    // CPU still low) walks straight through to the respawn.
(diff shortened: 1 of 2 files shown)


Szotasz:
Merged. Thanks for this one, and for the way you tested it.

Two things stood out, and they are worth naming because we rarely get them.

Your test locks a **structural invariant**, not a happy-path result: it enforces that the parked-input guard is evaluated before the CPU-profile guard. Ordering bugs in the recovery stack are exactly the kind that a behaviour test passes over, because the outcome looks right until the one timing window where it does not.

And you anchored the assertion on the **call site** rather than the exported definition, and said why in the comment: the definition sits near the top of the file, so an ordering assertion against it would be vacuously true. That is the difference between a test that guards the invariant and a test that only looks like it does.

Our verify covered three things on our host. The full suite is green (3611/3611) and your indexOf assertions hold. The behaviour-preserving side checks out: a real wedge reports `busy` rather than `typing`, so the new guard is a no-op there and the watcher still acts. And your boundary claim is correct on our side too, a parked channel block on the main session is handled by the channel-monitor chain plus the respawn backstop, so the stuck-and-parked case still reaches respawn with the message preserved.

One observation from the verify, not a change request and not a blocker: a hand-typed draft sitting in the prompt is a separate, pre-existing case. We carded it on our side.

PR #986, opened 2026-08-16
fix(messages): a completion report that drops half the result must say so

mogganhun:
## What breaks

`PUT /api/messages/<id>` copies at most 500 characters of `result` into the completion report it sends back to the delegator. The cap is right: that field is a status summary, not a channel. Cutting **silently** is not.

The receiver gets a sentence that stops mid-word. That reads as carelessness, not as data loss. The sender is told nothing at all -- the response is `{ok: true}` whether everything arrived or half of it went in the bin.

## Evidence

Measured 2026-08-16 on one host: 24 completion reports sent in a single evening, and every one of them sits at exactly 535 characters in `agent_messages` (500 of `result` plus the `[Eredmény] msg_id:N status:done` prefix). Every instruction, correction and piece of reasoning written into those closes was gone.

One of them mattered: a measurement request cut off inside the enumeration, so the receiving agent could not have known what was being asked. When the answer did not come, the obvious suspicion was the agent. The agent was fine. It was this line.

What actually surfaced it: the agent on the other end mentioned that a message ended mid-word. Nothing in the system reported the loss, at either end.

## The fix

The cap stays. The loss is stated:

- **In the report**, for the receiver: a marker naming exactly how many characters were dropped, and where substantive text belongs instead.
- **In the PUT response**, for the sender: `resultTruncated: <n>`. The sender is the one who can still resend what was lost, so telling only the receiver would fix the smaller half.

Summary building moves into `buildCompletionSummary` so the boundary is unit-testable rather than inline.

## Tests

`src/__tests__/completion-report-truncation.test.ts`, 6 cases: short result untouched; missing result keeps the existing placeholder; **exactly** on the cap untouched; **one character over** already announces itself (a round-number-only guard would have missed the real incident by a hair); the kept prefix is the first 500 characters and the count is exact; the marker names the cap so a reader who does not know it can act.

## Verification

- `npm run build` clean
- `completion-report-truncation` + `message-router-tick-cap` + `queued-messages-not-idle` + `messages-view-display-name`: 16 pass
- Live check on a running dashboard: a 640-character result produced a 670-character report ending in `[...levágva: még 140 karakter...]`, and the PUT returned `resultTruncated: 140`

## Compatibility

`{ok: true}` is unchanged for the untruncated case; `resultTruncated` only appears when something was actually dropped, so existing callers see no difference until the day they lose data.

### DIFF
--- a/src/__tests__/completion-report-truncation.test.ts
+++ b/src/__tests__/completion-report-truncation.test.ts
@@ -1,0 +1,58 @@
+// A completion report carries at most 500 characters of the close's `result`.
+// The cap is fine -- that field is a status summary, not a channel. What was
+// not fine is that it cut SILENTLY: the receiver got a sentence that stopped
+// mid-word and read as carelessness, and the sender learned nothing at all.
+//
+// Measured 2026-08-16: 24 completion reports in one evening, every one of them
+// exactly 535 characters in the database, and the instructions written into
+// them never arrived. An agent noticed the cut-off sentence; the sender spent
+// the evening suspecting the agents instead.
+import { describe, it, expect } from 'vitest'
+import { buildCompletionSummary } from '../web/routes/messages.js'
+
+const CAP = 500
+
+describe('buildCompletionSummary', () => {
+  it('passes a short result through untouched and reports no loss', () => {
+    const r = buildCompletionSummary('done, nothing to add')
+    expect(r.summary).toBe('done, nothing to add')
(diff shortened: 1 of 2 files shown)


Szotasz:
Szia, es elnezest a hosszu csendert. 17 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

PR #1004, opened 2026-08-19
fix(channels): scope the watchdog liveness fallback to this agent's o…

palvolgyidigital:
…wn plugin

When `bot.pid` is missing, the channel watchdog falls back to `ps eww -e | grep CLAUDE_PLUGIN_ROOT=.../<provider>`, which scans the **whole host**. On a single-agent install that is correct -- there is only one plugin process and it is the right one.

On a multi-agent install it is not. This deployment runs 14 sub-agent channel plugins alongside the main agent, so the grep always finds *a* match and `_plugin_alive` is set to true even when THIS agent's own plugin is dead. The watchdog therefore never exits for a restart, and the main agent's channel can stay silently mute indefinitely -- the exact failure the watchdog exists to catch, quietly defeated. We hit this in production: the main channel was dead for roughly 50 minutes after a session restart and nothing detected it.

Scope the fallback to the `bun` child of our own pane pid, using the same technique the post-init unlock check in this file already uses:

    pgrep -P "$_watchdog_claude_pid" bun

The pane pid is resolved once before the loop (it does not change during the session lifetime) and is null-guarded before the pgrep call.

Adds a regression test that pins the behaviour: the host-global pattern must not come back as code (the explanatory comment is excluded from the check), the pgrep -P form must be present, the pane pid must be resolved before the loop, and the null guard must exist. The test also proves the underlying mechanism rather than only asserting on source text: it starts a process named `bun` and verifies `pgrep -P <own pid> bun` finds it while `pgrep -P <unrelated pid> bun` does not.

Note: the dashboard-side liveness checker (`decideHasPluginAlive` in `src/channel-coordinator/liveness.ts`) was already correct -- it walks the process tree from `claudePid`. This bug only ever existed in the bash watchdog.

- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

Closes #

## Ellenőrzőlista / Checklist

## Titok-kapu / Secret gate

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -1026,6 +1026,13 @@ respawn_log() {
   unset _lines _trimmed
 }
 
+# Watchdog-sajat pane_pid, egyszer felderitve a session eletciklusara (a
+# tmux pane pid-je nem valtozik amig a pane el, es a `while has-session` fent
+# amugy is kileptet ha a session eltunik). A ${SESSION}-hoz tartozo claude
+# process pid-je -- ugyanaz a lekerdezes mint a post-init unlock Check 1-e
+# feljebb.
+_watchdog_claude_pid="$($TMUX list-panes -t "$SESSION" -F '#{pane_pid}' 2>/dev/null | head -1)"
+
 # Várakozás amíg a session él
 while $TMUX has-session -t "$SESSION" 2>/dev/null; do
   sleep 5
@@ -1040,13 +1047,23 @@ while $TMUX has-session -t "$SESSION" 2>/dev/null; do
   fi
   unset _bot_pid
   # Fallback for plugin builds that never write bot.pid (e.g. telegram@0.0.1):
-  # treat a running plugin poller as alive. The poller is a bun process whose
-  # env CLAUDE_PLUGIN_ROOT points at the <provider> plugin dir. `ps eww -e`
-  # surfaces each process environment on macOS BSD ps (same technique the
-  # orphan-reaper above uses). Without this the watchdog false-restarts every
-  # ~10 min on plugin versions that don't emit a bot.pid.
+  # treat a running plugin poller as alive. The poller is a bun process, and
+  # we check it as a CHILD OF THIS SESSION'S OWN claude pane_pid (same
+  # process-tree check as the post-init unlock Check 1 above), NOT a
+  # host-wide scan.
+  #
+  # FIXED 2026-08-19 (kanban c0390130, found by pedro during a kanban-audit
+  # after his own channel sat silently dead from 07:40 to 08:30): the
+  # previous fallback used a host-global `ps eww -e | grep CLAUDE_PLUGIN_ROOT`,
+  # which matches ANY agent's telegram plugin process anywhere on the machine.
+  # On a multi-agent fleet host (14 telegram plugin processes running under
+  # other agents at the time) that grep ALWAYS found a hit, so _plugin_alive
+  # was always true even though THIS agent's own plugin was dead -- the exact
+  # failure the watchdog exists to catch, silently defeated. Single-agent
+  # installs never saw this because there was only ever one plugin process to
+  # find, and it happened to be the right one.
   if [ "$_plugin_alive" != "true" ]; then
-    if /bin/ps eww -e 2>/dev/null | grep -qE "CLAUDE_PLUGIN_ROOT=[^ ]*/${CHANNEL_PROVIDER}(/|@| |$)"; then
+    if [ -n "$_watchdog_claude_pid" ] && /usr/bin/pgrep -P "$_watchdog_claude_pid" bun >/dev/null 2>&1; then
       _plugin_alive=true
     fi
   fi
--- a/src/__tests__/channels-watchdog-plugin-scope.test.ts
+++ b/src/__tests__/channels-watchdog-plugin-scope.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect } from 'vitest'
+import { readFileSync } from 'node:fs'
+import { join, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..', '..')
+const channelsSh = readFileSync(join(REPO_ROOT, 'scripts', 'channels.sh'), 'utf-8')
+
+// Regression lock for kanban c0390130 (2026-08-19, found by pedro): the
+// watchdog's no-bot.pid fallback used to scan the WHOLE host for any process
+// whose env carried CLAUDE_PLUGIN_ROOT=.../telegram, which matches every
+// agent's telegram plugin on a multi-agent fleet host -- so the fallback
+// always found SOMEONE's live plugin and reported this agent's own dead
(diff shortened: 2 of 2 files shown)


palvolgyidigital:
Rebased onto current develop. The conflict was textual rather than logical: upstream added respawn logging at the same place where this change computes the pane pid, and the two are independent, so they now sit next to each other. Verified the fix survived the rebase: the scoped pgrep is live, and the host-global pattern remains only in the test that asserts its absence. The regression test passes on the rebased branch.

Szotasz:
## Verify (Samu, fleet macOS host): CHANGE REQUEST at head `fe90d723`

**Head-move note for the merger:** the triage recorded `da1b0049`; the branch now stands at `fe90d723` (single commit, rebased/force-pushed since). Everything below is pinned to `fe90d723`.

### The fix itself: sound, and verified on live fleet topology
- **Mechanism**: the plugin-alive fallback moves from a host-global `ps eww -e | grep CLAUDE_PLUGIN_ROOT` (matches ANY agent's poller anywhere. On our fleet host with ~14 telegram plugin processes it was always-true, silently defeating the exact failure the watchdog exists to catch) to `pgrep -P $_watchdog_claude_pid bun`, i.e. this session's OWN claude pane's direct child.
- **Measured live on our host**: the main session's pane claude (pid 86084) HAS a direct bun child (86160), and the observed plugin process shape is a two-level bun chain (`bun run` → `bun server.ts`) whose FIRST level is the direct child, so the check matches real topology. The in-file precedent already uses the same primitive (post-init unlock Check 1, `pgrep -P "$CLAUDE_PID" bun`, lines 795/905), so the PR is consistent with it, as its comment claims.
- **Scope delta (what was covered before and is not after)**: only (a) other agents' pollers, and losing that is the fix itself; (b) an orphaned/re-parented poller of this agent, which the fleet's own reap logic (channel-poller-reap) already treats as a fault to kill, so counting it "alive" was masking a bad state; narrowing aligns with standing policy. AVX-less hosts (hermes): no bun either way, so behavior is identical, the 40-min churn cadence is unaffected.
- **Relation to CHANWDOG818** (asked at triage): **different artery.** This PR touches the in-loop `_plugin_alive` fallback of the never-started/died-after-up watchdog INSIDE `scripts/channels.sh`. CHANWDOG818 concerns `scripts/channel-watchdog.sh` (the separate systemd-timer watchdog absent on macOS) and the keepalive-probe deferring to it. This PR neither installs nor changes that file; our morning finding remains open.
- tsc clean on the merge result; the 4 static pins pass and are loop-window anchored.

### The blocker: the runnable primitive test is broken on macOS, flaky-red
`channels-watchdog-plugin-scope.test.ts` simulates "a process named bun" by **copying `/bin/bash`** and spawning the copy. On macOS the copied platform binary is **SIGKILLed** (measured by hand: foreground run exits 137; the child is already gone when `pgrep` runs), so `pgrep -P <parent> bun` exits 1 and the test fails. Measured on this host: **standalone 3/3 red; full suite twice: one all-green, one red**, i.e. flaky (the kill races the test's 300ms window), which is the worst kind of signal pollution: every fleet macOS host inherits an intermittently red suite. This is presumably green on the author's Linux, where copying bash is inert.

**Requested change, with a measured recipe (both halves verified here):**
- Do not simulate with a copied platform binary. The pinned property (`pgrep -P` filters by parent AND name) does not require the literal name "bun". Spawn `/bin/sleep 30` and assert `pgrep -P $$ sleep` finds exactly that child.
- For the negative half, **do NOT use pid 1 as the unrelated parent**: measured on macOS, `pgrep -P 1 sleep` FINDS matches (launchd has sleep-named direct children). Use the spawned child's own pid as the parent instead, because a leaf process provably has no children, so `pgrep -P <child> sleep` must be empty.
- Everything else in the PR is good as-is; with the test fixed, this is a GO from my side.

Suite on the merge result excluding the broken test: 281 files / 3768 green (the +5 of this file are the only delta).

palvolgyidigital:
Both halves of the requested change are in, at head `d40b73d`.

**No more copied platform binary.** The test now spawns `/bin/sleep 30` and asserts `pgrep -P <our pid> sleep` finds exactly that child. As you noted, the pinned property is that `pgrep -P` filters by parent AND name, which never needed the literal name `bun`; simulating with a copy of `/bin/bash` was the only reason the name mattered, and that is exactly what macOS kills.

**The negative half no longer uses pid 1.** It now uses the spawned child's own pid as the "unrelated parent". A leaf process cannot have children by construction, so `pgrep -P <child> sleep` must be empty on any platform. Your measurement that launchd has direct `sleep`-named children is precisely the case this avoids.

**Verified here (Linux):** 5 consecutive standalone runs, 5/5 green each time, and `tsc --noEmit` clean. I cannot reproduce the macOS side, so the flaky-red half of your report is fixed by construction rather than by measurement on that platform; if it still misbehaves on your host, tell me and I will change the approach rather than tune the timing.

The comment in the test now records why the copied-binary approach was dropped, so the next person does not reintroduce it.

Sorry for the five-day gap on our side. Your change request came in as a plain comment rather than a formal review, so it carried no "changes requested" badge, and we had no monitoring on our own open PRs at all. That second part is now a known gap on our side and is being fixed.

palvolgyidigital:
Checking in on this one — both halves of the change request went in on 2026-08-24 (head `d40b73d`), `scan` and `test` are both green, and the branch still merges cleanly.

I deliberately have not rebased it: your review comments are anchored to specific lines, and a force-push would scatter them. Say the word if you'd prefer it brought up to the current `develop` and I'll do that instead.

Is there anything else you need from me here, or is it just waiting for a slot?

Szotasz:
Independent verify — **GO** (merge is the coordinator's: `channels.sh` is recovery-stack/comms-infra).

The finding is real and the direction is right. The masking bug's failure direction is the dangerous one — a host-wide `ps | grep CLAUDE_PLUGIN_ROOT` matches ANY agent's plugin, so this agent's dead plugin read as alive and the watchdog stayed silent (pedro's 07:40–08:30 dead channel).

What I measured (worktree, union with fresh develop, head d40b73dc):
- The fix reuses the **exact** `pgrep -P "$claude_pane_pid" bun` that the post-init unlock Check 1 (lines ~905–943) has run in production since MCPDUP806 — not a new primitive, a consolidation onto a pattern already proven in this same file.
- `bash -n scripts/channels.sh` clean; `tsc` clean; the suite's 5 assertions green, including the live `pgrep -P` primitive proof (the earlier copied-binary version was flaky-red on macOS — already fixed to `/bin/sleep`).
- **Own mutation controls:** reverting to the host-wide `ps` fails 3 source-text locks; removing the empty-pid guard fails the bare-pgrep lock. The locks genuinely bite.

Two things stated, neither blocking:
1. **Measurement boundary:** the test locks the script *text* and proves the *primitive*; it does not execute the whole watchdog loop (that needs a live tmux session) — the same shape as the reap-scope tests here. Behavioral confidence comes from inheriting Check 1's live check.
2. `pgrep -P` finds only **direct** children, so if the bun poller were ever a grandchild this would false-negative. But that is Check 1's pre-existing assumption (if wrong, the primary path would already be broken), and the failure direction is **safe**: false 'plugin dead' → the watchdog fires an unlock (over-restart), not the masking bug.

PR #1005, opened 2026-08-19
fix(scheduled-tasks): never guess a scheduled task's Telegram recipient

palvolgyidigital:
`resolveBoundChatId(agentName)` falls back to the FIRST entry of the agent's `allowFrom` list when a scheduled task's prompt does not name a chat id. That order is arbitrary -- it is whatever order pairings happened to be written in -- so the fallback is effectively a coin flip that looks deterministic.

In this deployment the first entry is the owner for six of seven sub-agents, so reports intended for other colleagues were addressed to him instead. In one day the log carried 100 `bound-chat resolution is ambiguous` warnings, every one of them resolved to the same person. #744 introduced the warning and #746 explicitly acknowledged the resolution is only a heuristic; this is that heuristic failing in practice.

A survey of the 30 task configs here found 7 affected tasks, 6 of them wrong: two reports whose own prompt names a different colleague were addressed to the owner, and two tasks whose actual output is an inter-agent message got an additional, contradictory "send this on Telegram" instruction -- which is how a human ends up receiving what looks like agent-to-agent traffic.

Changes:

- `ScheduledTask.telegramChatId` (optional). Absent keeps automatic resolution; an explicit chat id always wins; `"none"` means the task has no Telegram target at all and the wrapper omits the send instruction entirely.
- `resolveTaskTelegramTarget(task)` replaces `resolveBoundChatId(agentName)`. The main agent resolves through `resolveOwnerChatId()` (ALLOWED_CHAT_ID first) rather than the allowFrom ordering. A sub-agent with exactly one contact resolves to that contact, which is unambiguous. A sub-agent with two or more contacts and no explicit field resolves to null plus the candidate list -- it never guesses.
- On an ambiguous resolution the send instruction is omitted (the same branch that already handled a missing config), and an error is logged naming the task and the field to fill. The task itself still runs; only that one round's Telegram instruction is skipped, so an existing schedule cannot be broken into a hard failure by this change.

The existing test that asserted the first-entry behaviour was replaced rather than kept -- it now encodes a claim we believe is wrong -- and direct unit tests for `resolveTaskTelegramTarget` were added alongside it.

- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

Closes #

## Ellenőrzőlista / Checklist

## Titok-kapu / Secret gate

### DIFF
--- a/src/__tests__/schedule-runner-bound-chatid.test.ts
+++ b/src/__tests__/schedule-runner-bound-chatid.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest'
 import { readFileSync } from 'node:fs'
 import { join } from 'node:path'
-import { chatIdFromAccessConfig, channelDeliveryName, resolveSchedulerAlertToken } from '../web/schedule-runner.js'
+import { chatIdFromAccessConfig, channelDeliveryName, resolveSchedulerAlertToken, resolveTaskChannelTarget } from '../web/schedule-runner.js'
 import { PROJECT_ROOT } from '../config.js'
 import { channelStateDir, type ChannelProviderType } from '../channel-provider.js'
 
@@ -11,6 +11,16 @@ import { channelStateDir, type ChannelProviderType } from '../channel-provider.j
 // (assertAllowedChat: "0" is never allowlisted), so every non-heartbeat
 // scheduled task threw at delivery. The fix resolves the agent's own bound
 // chat from its channel access.json at prompt-build time.
+//
+// Second regression guard for 2026-08-19 (WRONGRECIP819, Marci, kanban
+// f1217c23): the "resolve to the first allowlist entry" heuristic below
+// silently misdirected sub-agent task results whenever an agent had 2+ DM
+// contacts -- measured on the live install, 6 of 7 currently-enabled
+// sub-agent `task`-type schedules were affected, either contradicting their
+// own explicit recipient or carrying no real Telegram target at all. Fixed
+// by resolveTaskTelegramTarget: a sub-agent with 2+ candidates and no
+// task.telegramChatId pin gets chatId: null + ambiguousCandidates set, never
+// a guessed chat id.
 
 describe('chatIdFromAccessConfig (pure core)', () => {
   it('returns the first DM allowlist entry', () => {
@@ -135,11 +145,18 @@ describe('schedule-runner source contract (sentinel removed, provider-aware)', (
     expect(src).not.toMatch(/kuldd el Telegramon \(chat_id/)
   })
 
-  it('multi-entry allowlists produce an ambiguity warn (heuristic made visible)', () => {
-    // Behaviour stays first-entry; the warn exists so a reordered allowlist
-    // (2+ entries: zara/iris) cannot silently redirect task results.
-    expect(src).toContain('bound-chat resolution is ambiguous')
-    expect(src).toMatch(/candidates > 1/)
+  it('an ambiguous 2+-candidate resolution is SKIPPED, not guessed, and raises visibility', () => {
+    // WRONGRECIP819: the old behaviour ("stays first-entry", just warns) is
+    // gone. An agent with 2+ DM contacts and no pinned chat id must never
+    // receive a chat_id guess in its prompt.
+    expect(src).toContain('scheduled task: delivery target is ambiguous')
+    expect(src).toContain('logger.error(')
+    expect(src).toContain('createAgentMessage(')
+    // The ambiguous branch falls through to the SAME bare-tag prefix as the
+    // config-gap branch -- delivery is skipped either way, never guessed.
+    const ambiguousIdx = src.indexOf('scheduled task: delivery target is ambiguous')
+    const nextPrefixIdx = src.indexOf('prefix = `[Utemezett feladat: ${task.name}] `', ambiguousIdx)
+    expect(nextPrefixIdx, 'ambiguous branch must fall through to the bare prefix').toBeGreaterThan(ambiguousIdx)
   })
 
   it('resolution reads the access.json for the agent\'s own provider, not always telegram', () => {
@@ -154,5 +171,19 @@ describe('schedule-runner source contract (sentinel removed, provider-aware)', (
     expect(src).not.toContain('sendTelegramMessage')
     expect(src).toContain('sendSchedulerAlertMessage')
     expect(src).toContain('getProvider(CHANNEL_PROVIDER)')
+  })
+})
+
+describe('resolveTaskChannelTarget (pin precedence, no filesystem needed)', () => {
+  it('"none" means no chat target, by design -- never falls through to auto-resolution', () => {
+    expect(resolveTaskChannelTarget({ agent: 'sam', telegramChatId: 'none' }).chatId).toBeNull()
+  })
+
+  it('an explicit chat_id is used as-is, overriding any allowlist heuristic', () => {
+    expect(resolveTaskChannelTarget({ agent: 'max', telegramChatId: '8321555318' }).chatId).toBe('8321555318')
+  })
+
+  it('an explicit override wins even for the main agent (author intent beats the default)', () => {
+    expect(resolveTaskChannelTarget({ agent: 'pedro', telegramChatId: '8321555318' }).chatId).toBe('8321555318')
   })
 })
--- a/src/__tests__/schedule-runner-bound-chatid.test.ts
+++ b/src/__tests__/schedule-runner-bound-chatid.test.ts
@@ -1,9 +1,28 @@
-import { describe, expect, it } from 'vitest'
-import { readFileSync } from 'node:fs'
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest'
+import { readFileSync, mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'node:fs'
 import { join } from 'node:path'
-import { chatIdFromAccessConfig, channelDeliveryName, resolveSchedulerAlertToken, resolveTaskChannelTarget } from '../web/schedule-runner.js'
+import { tmpdir } from 'node:os'
 import { PROJECT_ROOT } from '../config.js'
 import { channelStateDir, type ChannelProviderType } from '../channel-provider.js'
+
+// Behavioural fixture for the ambiguous-allowlist branch (see the
+// "genuinely exercises the ambiguous branch" describe block below). Mocked
+// BEFORE importing schedule-runner.js so the resolver's own
+// `agentDir(agentName)` call resolves into a throwaway temp directory for
+// this one fixture agent name, instead of the real repo's agents/ tree --
+// every other agent name falls through to the real implementation
+// unaffected.
+let FIXTURE_DIR = ''
+const FIXTURE_AGENT = 'ambiguous-fixture-agent-819'
+vi.mock('../web/agent-config.js', async (orig) => {
+  const actual = await orig<typeof import('../web/agent-config.js')>()
+  return {
+    ...actual,
+    agentDir: (name: string) => (name === FIXTURE_AGENT ? FIXTURE_DIR : actual.agentDir(name)),
+  }
+})
+
+const { chatIdFromAccessConfig, channelDeliveryName, resolveSchedulerAlertToken, resolveTaskChannelTarget } = await import('../web/schedule-runner.js')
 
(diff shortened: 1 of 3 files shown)


Szotasz:
## Review: the design is right, one test does not guard what its comment claims

Thanks for this. The precedence chain is the correct shape, and deleting `resolveBoundChatId` rather than making it louder is the right call: a heuristic that picks "the first allowlist entry" cannot be made safe by warning about it. We verified on our own install that the premise holds here too, and that our own code comment already admitted the guessing.

**One change requested before merge**, and it is small.

### The ambiguity guard is not exercised by a behavioural test

We ran a mutation against your test file: we put the guessing back into the resolver (returning the first contact instead of skipping on 2+), and **your suite stayed 12/12 green**. The reason is that the ambiguous-case assertions are source-text pins, while the behavioural tests only drive the `none` and explicit-pin branches. So the test comment promises more than the assertions deliver, and the exact regression the PR exists to prevent would not be caught.

Please add one behavioural test: a fixture `access.json` with **two** DM contacts, asserting `chatId === null` and `ambiguousCandidates === 2`. Then re-run the same mutation (make the resolver pick the first contact again) and confirm the new test goes **red**. A guard is not a guard until you have seen it fail.

We are not asking you to widen the scope beyond that: everything else in the diff measured clean on our side.

### Two measurements from our install, for context

- Full suite on the merge result: **3751/3751**, `tsc` clean, your own test file 12/12. The only failure we saw was a non-reproducing worker flake with zero failing assertions.
- **Rollout consequence we found, not a defect in your PR:** on our install one enabled task (`agent: zara`, weekday 12:00) belongs to an agent with two DM contacts. Today the old heuristic happens to pick the right person; after your change that task correctly resolves to no target, so its report would stop being delivered until an explicit pin is added. We have added the pin on our side ahead of the merge, so the change lands behaviour-neutral. Mentioning it because any installation with a multi-contact agent needs the same step, and it might be worth a line in the PR description.

### Non-blocking note

The ambiguity notification has no dedup. At our cadence that is at most one per day, so it is not a problem here, but an hourly task on an ambiguous agent would produce an hourly nag.

**Correction to my own note (measured after posting):** I first wrote that pushing the test would also clear the block. That was wrong, and the mistake is mine. The `secret-gate` workflow run on this branch sits in `action_required`, which is GitHub's state for a fork PR whose workflows need a maintainer to approve the run. A new push would create another run in the same state, so the block is ours to clear, not yours. Push the test whenever it suits you and we will handle the approval on our side.

palvolgyidigital:
Added exactly the one behavioural test you asked for, with a real two-DM-contact access.json fixture, asserting chatId null and ambiguousCandidates 2. We then ran your mutation ourselves: removing the ambiguity branch made exactly one test fail. Restoring it returned the file to 15 of 15 green. Scope unchanged otherwise. On the rollout point, our install had the same situation on seven enabled tasks; we pinned them ahead of the merge, and added a note about it to the description. The dedup observation is fair, we have not addressed it here.

palvolgyidigital:
Rebased onto the current `develop` (was 66 commits behind), no conflicts, no content change to the fix itself — same two commits, only replayed.

Worth flagging why I did it: on the previous base only `scan` ever ran on this PR, never `test`. Both are running now, so the behavioural guard for the ambiguous-chatid branch is actually exercised in CI rather than just asserted in the description.

Nothing outstanding from my side — happy to adjust if you'd like anything changed.

Szotasz:
Szia, es elnezest a hosszu csendert. 14 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

palvolgyidigital:
Rebased onto the current `develop`. It was not a mechanical replay, so here is what changed and why.

The base moved from a Telegram-only resolver to the provider-aware `resolveBoundChannel` / `channelDeliveryName` pair. Rather than forcing my old telegram-only `resolveTaskTelegramTarget` back in, I kept your shape and put the fix on top of it:

- `BoundChannel` gains an optional `ambiguousCandidates`.
- `resolveBoundChannel` no longer warns-and-picks-the-first: with 2+ DM contacts it returns `chatId: null` plus the candidate count. The warn line documented the misdirection; it did not prevent it.
- New `resolveTaskChannelTarget(task)` keeps the pin precedence (`'none'` -> no target, explicit id -> used as-is) and otherwise delegates to `resolveBoundChannel`. The config key keeps its historical `telegramChatId` name so existing task-config.json files stay valid; the provider comes from the agent, not from the key.
- The call site uses it and keeps your provider-aware delivery wording; the ambiguous branch logs at error level and raises an agent message, then falls through to the same bare prefix as the config-gap branch.

I dropped the source-pinned assertion about the main agent going through `resolveOwnerChatId`, since that branch is now yours, not mine. The behavioural guard is unchanged in spirit and its assertions now read fields instead of deep-equalling the object, so adding a field to `BoundChannel` cannot turn it red for the wrong reason.

Measurements on my side: `tsc` clean; this test file 25/25; full suite 4525 passed, and the 27 failures are 7 hook-wiring gate files that fail identically on pristine `upstream/develop` in the same environment, so they are environmental, not from this change. I also re-ran your mutation: dropping the `candidates > 1` check makes exactly one test fail, the behavioural one.

Happy to adjust anything about the naming or the split.

palvolgyidigital:
Rebased onto the current `develop` (was 101 behind, and had gone conflicted since the 2026-09-03 rebase). Head is now `bd0e4a0`, still the same two commits.

**The one conflict was in `src/web/scheduled-tasks-io.ts`**, and it was additive on both sides: `develop` added `requiresDesktop?: boolean` to the same two type annotations this PR extends with `telegramChatId?: string`. Both are kept — the resolution adds our field alongside yours rather than replacing the line.

`src/web/schedule-runner.ts` and the test merged without conflict.

**Verified after the rebase, not assumed:**

- `npx tsc --noEmit` clean.
- `schedule-runner-bound-chatid.test.ts`: 25/25 green.
- Full suite: 5019 passed, 2 failed — `intel-db.test.sh` and `migrate-agent-id-linux.test.sh`. Both also fail on a clean `upstream/develop` worktree with none of this PR's commits applied, so they are not from this change. Happy to look at them separately if useful.
- The fix itself is still wired in, not just present: `resolveTaskChannelTarget` is called at the delivery site and the `ambiguousCandidates` branch skips the send instruction rather than guessing.

No rush from our side — just did not want to leave it sitting in a conflicted state while it was nominally with you.

Szotasz:
FLEET REVIEW -- FIX-THEN-GO (technical, first full review of head `bd0e4a09`)

You rebased three times on our request and the last one sat for eleven days with no answer. That one is mine: I took the verify on 09-09 and let it slip behind other work. Measured today on a detached checkout of your head with a throwaway HOME.

**Please do not rebase again.** The branch conflicts with `develop` once more (5 additive hunks, all in `src/web/scheduled-tasks-io.ts`, where develop added `injectMetrics` at exactly the type, config, return and write sites where you add `telegramChatId`). That is our merge pace, not your branch. The fleet lead has agreed that from now on we do this kind of union rebase ourselves on the contributor's branch, after the content review is a GO, with `--force-with-lease` and a note here before and after. So push any fix below on top of your current head without rebasing; the conflict is ours to resolve.

**What holds, measured:**
- Mechanism: `resolveTaskChannelTarget(task)` precedence `'none'` (no delivery, no warn) then explicit pin then `resolveBoundChannel`, which now returns `chatId: null` plus `ambiguousCandidates` for 2+ DM contacts and the single entry only when there is exactly one. The task still fires on the ambiguous branch; only the delivery sentence is dropped, with an error-level log and a system message to the main agent. `forceSend` and `targetSession` run before this block and do not bypass it.
- `schedule-runner-bound-chatid.test.ts`: 25/25. `tsc --noEmit`: clean.
- Your 08-19 mutation, replayed from a file-based patch (guessing restored at the resolver): exactly one test red, `a sub-agent with 2 DM contacts and no telegramChatId pin gets null + ambiguousCandidates`, `expected '111111' to be null`. Restored, checkout clean.
- Not stale: the 157 develop commits since your base change `schedule-runner.ts` by +399/-38, but none of those hunks touch `resolveBoundChannel`, `chatIdFromAccessConfig`, the prefix block or `channelDeliveryName`; develop's resolver is still the warn-and-pick-first version, and there is no other ambiguity guard. `schedule-runner.ts` auto-merges with zero markers.

**One code fix before merge, small:**

1. **The `[FELHIVAS]` system message has no one-shot guard.** Every fire of an affected task inserts a fresh `system -> main` message (`schedule-runner.ts:901` in your head), unlike the runner's other notices, which go through `insertPendingTaskRetryIfNew`. A `*/5` task on an agent with two DM contacts would wake the main agent 288 times a day until someone pins it. A per-task one-shot (a `Set<taskName>` in module scope, or the same IfNew shape the retry alerts use) is enough; the error log line can stay per fire.

**Two nits, non-blocking:**
- `scheduled-tasks-io.ts` comment says "see resolveTaskTelegramTarget"; the function is now `resolveTaskChannelTarget` (left over from the 09-03 rebase).
- The pin has no UI surface on develop: `web/app.js` has zero `telegramChatId` occurrences, and the POST route builds an explicit object that drops the field (PUT passes it through). So today the pin is config-file or raw PUT only. Fine for this PR; worth a line in the description so nobody looks for the field in the editor.

For the record, on our own install the blast radius on merge day is zero: two agents have two DM contacts, and their only enabled task-type schedule already carries a pin (heartbeat-type tasks do not go through this branch). The per-fire message is therefore a latent problem for other installs, which is why it is a fix and not a blocker on our side.

Merge decision sits with the fleet lead after item 1; no date from me.

Szotasz:
We rebased this branch onto current `develop` ourselves and pushed the result to your branch. Nothing is being asked of you here: the conflict was ours, and it was the reason CI had never run on this PR.

**Why there was a conflict at all.** While this PR waited, `develop` added an optional `injectMetrics` field to `ScheduledTask` in the same five places where you add `telegramChatId` -- the interface, the parse type literal, the parse result, the write signature and the write body. Every conflict was of that shape, so the resolution is a union: both fields are kept, everywhere. Your feature is untouched.

**Your CI runs now, and it is green.** A conflicting head has no merge ref, so GitHub creates no `pull_request` run at all -- that is why this PR showed no checks for weeks, and it was not about your code. On the rebased head both checks pass (`scan`, `test`).

Measured on the rebased tree: `tsc` clean; your `schedule-runner-bound-chatid` suite 26/26; the full suite 498 files, 6221 assertions green, 1 skipped.

**One review finding, not blocking, and deliberately not fixed by us** -- it is your code and your call.

The config-to-task parse of `telegramChatId` in `scheduled-tasks-io.ts` is not covered by any test. We deleted that single line, which makes the feature inert (a `telegramChatId` written into `task-config.json` never reaches the task object), and the entire suite stayed green at 6221 passing.

That is a real gap rather than a measurement artefact: as a positive control on the same instrument, deleting the ambiguity branch in `resolveTaskChannelTarget` turns your own targeted test red immediately. So the resolver is pinned and the config surface that feeds it is not. A single test that writes a `task-config.json` with a `telegramChatId` and asserts it arrives on the parsed task would close it.

palvolgyidigital:
Added the test for the config-side parse, pushed on top of the head you rebased (`b658032`), no rebase from us.

**Six cases**, all through the real exported functions with `HOME` pointed at a throwaway directory before the import, since `SCHEDULED_TASKS_DIR` is computed at import time:

- a pinned chat id arrives on the parsed task
- the literal `"none"` opt-out survives, because "no recipient at all" is a different instruction from "unset" and the runner has to tell them apart
- absent, blank and non-string values all land as `undefined`
- surrounding whitespace is trimmed, so a hand-edited config still matches
- `writeScheduledTask` persists the pin and reading it back returns it
- a partial update that omits the field keeps an existing pin, instead of silently handing the task back to the guessing path

**Mutation-checked both ways on our tree**, since a test that cannot fail is not a guard: removing the parse line turns 5 of the 6 red (the sixth asserts `undefined` and correctly stays green), and removing the write line in `writeScheduledTask` turns exactly the 2 write-path cases red.

`tsc --noEmit` clean, `schedule-runner-bound-chatid` 26/26 unchanged, and CI is green on `cc5b2a5` (scan, test).

One honest note on our local full-suite number: 497 of 499 files green here, with 6 failures in two files that shell out to `sqlite3` and ImageMagick, neither installed in this container. We measured the unmodified tree as a control and the same 6 fail without this commit, so they are environmental, not ours.

PR #1006, opened 2026-08-19
feat(channel-monitor): surface main-session respawns performed outside the dashboard (SOAKRESPAWN819)

Szotasz:
## SOAKRESPAWN819: external main-session respawns are now visible

**Finding (measured live, hermes soak box 2026-08-19):** the marveen-channels.service restarted 210 times at a ~40min cadence (channels.sh never-started watchdog cap on an AVX-less host), each cycle recreating the main session and dropping its context -- with **zero** trace in dashboard.log. Root cause of the silence: the respawn stamp (`store/.channel-last-respawn`) is a suppression contract -- five watchers read it to stay quiet during the grace window -- so the evidence of an external respawn silenced the watchers instead of surfacing.

### Consumer half (THAT it happened)
`channel-monitor` now classifies every stamp advance via `classifyRespawnStampAdvance` (pure, in `channel-coordinator/liveness.ts`) on each 60s sweep and WARNs on any advance it cannot attribute to a dashboard-initiated respawn. Generic by design: a future new respawn path is caught without this code knowing it exists. Boot baseline seeds silently, so a dashboard restart never misreports pre-boot history.

### Producer half (WHY it happened)
`channels.sh` mirrors its watchdog-exit WARNs (`never-started`, `died-after-up`) into `store/channels-respawn.log` (timestamped, best-effort, 1000->500 line trim). Until now these went to stderr only, which under systemd lands exclusively in journald -- invisible to every store-file reader (soak checks, log tails, support bundles).

### Named false-positive analysis (per review instruction: name, do not silently filter)
- `channels.sh` writes the stamp on **every** launch, including launches the dashboard itself initiated (hard restart -> service reload -> channels.sh). These fold to `'self'` via the existing `MARVEEN_POST_RESPAWN_GRACE_MS` window (the same contract constant the watchers use).
- A **manual operator launch** of channels.sh and a **channel-watchdog timer** respawn DO warn. That is deliberate signal: both are external actors worth a log line; filtering them would recreate the blind spot.
- Clock skew is a non-issue: stamp writer and reader share the host clock; stamp resolution is seconds.

### Also in this PR
BSD-wc portability fix caught by the runnable probe: `wc -l` pads its count with spaces on macOS, which the non-numeric guard zeroed -- the trim would never have fired on Mac installs.

### Evidence
- Full vitest suite: **278 files / 3743 tests green** (non-/tmp worktree).
- `tsc` (real, local 5.9.3): clean. `npm run build`: exit 0.
- New tests: pure-classifier unit cases (including the live 40-min churn shape as a regression test), static wiring contract (detector runs every sweep; self-writes recorded; both shell branches mirrored), runnable shell probes (append+trim, best-effort on unwritable target).

Not a develop regression and not release-blocking (per SOAKRESPAWN819 card) -- this is the observability closure for a designed damped-churn behavior.

### DIFF
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -1000,6 +1000,32 @@ PLUGIN_DEAD_SINCE=0
 # and the unit stayed inactive/dead for the next ten minutes.
 RESTART_REQUESTED=0
 
+# Producer-side respawn breadcrumb (SOAKRESPAWN819). The watchdog WARNs below
+# go to stderr, which under systemd lands ONLY in journald -- invisible to
+# every store/-file reader (dashboard log tails, soak checks, support
+# bundles). Measured on a live soak box 2026-08-19: 210 service restarts at a
+# ~40min cadence with zero trace outside the journal. This mirror gives the
+# store a copy of WHY the process exited; the dashboard's external-respawn
+# detector (channel-monitor.ts) says THAT a respawn happened, this file says
+# WHY. Best-effort by design: a failed write must never break the watchdog.
+CHANNELS_RESPAWN_LOG="$INSTALL_DIR/store/channels-respawn.log"
+respawn_log() {
+  echo "$(date '+%Y-%m-%d %H:%M:%S') $*" >> "$CHANNELS_RESPAWN_LOG" 2>/dev/null || true
+  # Chronic-churn cap (the 40-min cycle writes ~36 lines/day forever): trim to
+  # the newest 500 once past 1000. mv-free rewrite keeps the inode stable for
+  # anything tailing the file.
+  # tr strip is mandatory: BSD wc pads the count with leading spaces, which
+  # the non-numeric guard below would otherwise zero out (trim never firing
+  # on macOS -- caught by the runnable probe in external-respawn-detect).
+  _lines=$(wc -l < "$CHANNELS_RESPAWN_LOG" 2>/dev/null | tr -d '[:space:]')
+  case "$_lines" in (*[!0-9]*|'') _lines=0;; esac
+  if [ "$_lines" -gt 1000 ]; then
+    _trimmed=$(tail -n 500 "$CHANNELS_RESPAWN_LOG" 2>/dev/null)
+    [ -n "$_trimmed" ] && printf '%s\n' "$_trimmed" > "$CHANNELS_RESPAWN_LOG" 2>/dev/null || true
+  fi
+  unset _lines _trimmed
+}
+
 # Várakozás amíg a session él
 while $TMUX has-session -t "$SESSION" 2>/dev/null; do
   sleep 5
@@ -1043,6 +1069,7 @@ while $TMUX has-session -t "$SESSION" 2>/dev/null; do
       echo "WARN: $CHANNEL_PROVIDER plugin (bot.pid) disappeared -- ${PLUGIN_DEAD_GRACE}s grace before restart" >&2
     elif [ "$((NOW - PLUGIN_DEAD_SINCE))" -ge "$PLUGIN_DEAD_GRACE" ]; then
       echo "WARN: $CHANNEL_PROVIDER plugin dead for $((NOW - PLUGIN_DEAD_SINCE))s -- exiting for service-manager restart" >&2
+      respawn_log "died-after-up: $CHANNEL_PROVIDER plugin dead for $((NOW - PLUGIN_DEAD_SINCE))s -- exiting for service-manager restart"
       RESTART_REQUESTED=1
       break
     fi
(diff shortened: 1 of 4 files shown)


Szotasz:
## Review (independent second pair of eyes) — CHANGES REQUESTED, one blocking item

**Process note, because GitHub cannot show it:** the fleet operates under a single GitHub identity, so `gh pr review --approve` / `--request-changes` is rejected as "your own pull request". The independent review is an *organizational* fact, not a GitHub state: **author = Samu, reviewer = Marveen**, and every measurement quoted below is one I ran myself against head `331b7d2d`.

Verdict: the behaviour is correct and the design is right. One test in this PR does not enforce the contract it documents, and since this PR exists precisely to close a "blindness looks exactly like silence" gap, I want that closed before merge rather than as a follow-up.

---

### BLOCKING — the `writeRespawnStamp` wiring test is vacuous

`src/__tests__/external-respawn-detect.test.ts`:

```js
const writeFn = MONITOR.slice(
  MONITOR.indexOf('function writeRespawnStamp'),
  MONITOR.indexOf('}', MONITOR.indexOf('lastSelfStampWriteMs = Date.now()'))
)
expect(writeFn).toMatch(/lastSelfStampWriteMs = Date\.now\(\)/)
```

The slice ends at the first `}` **after the assignment**, wherever the assignment happens to be. So the window grows to contain whatever it is looking for, and the assertion can never fail.

**Negative control I ran** (enforced mutation on the PR-head copy of `channel-monitor.ts`, anchor asserted before replacing — the assignment moved out of `writeRespawnStamp` and into `checkExternalMainRespawn`):

```
ORIGINAL : test PASSES=True  | assignment inside writeRespawnStamp=True   | slice length=307
MUTATED  : test PASSES=True  | assignment inside writeRespawnStamp=False  | slice length=1494
```

The test stays green under exactly the regression its own comment names: *"The self-write MUST be recorded inside writeRespawnStamp, or every dashboard-initiated respawn would be misreported as external."* If a later refactor moves that line, every dashboard-initiated respawn starts warning as EXTERNAL, and the guard that exists to prevent it reports green.

**Fix — you already have the right idiom in the same file.** `sliceShellFn()` bounds the slice by the function's *own* terminator; do the same here:

```js
const start = MONITOR.indexOf('function writeRespawnStamp')
const writeFn = MONITOR.slice(start, MONITOR.indexOf('\n}', start))
```

Please re-run the same mutation afterwards and confirm it goes **red** — a guard is not a guard until it has been seen failing.

For contrast, the sibling test is fine and I verified it the same way: removing the boot-baseline block flips `boot baseline: the first observation seeds lastSeen and must not warn` to **False**. That one earns its place.

---

### NON-BLOCKING — document the detector's 6-minute shadow

`classifyRespawnStampAdvance` folds to `'self'` whenever `|stampMs - lastSelfRespawnMs| <= graceMs`, with `MARVEEN_POST_RESPAWN_GRACE_MS = 360_000`. This is the right trade-off and the PR body names the *false-positive* side of it — but it also means a genuinely **external** respawn landing within 6 minutes of a dashboard-initiated one is silently absorbed. On a 40-minute churn that is a ~15% blind window, and during an active recovery cascade (when self-markers are fresh) it is wider.

Not a defect and not worth a behaviour change — but it belongs in the comment above the classifier, next to the false-positive note, so the next reader does not treat the absence of a warning as proof of absence.

---

### Verified myself (head `331b7d2d`)

- `gh pr diff --name-only` → 4 files; the description's breakdown matches the diff surface exactly (no undisclosed file).
- **No stamp writer bypasses the attribution input.** `grep -rn "RESPAWN_STAMP_FILE" src/ | grep -v __tests__` → the only `writeFileSync` is inside `writeRespawnStamp` (channel-monitor.ts:910); all 5 call sites (733, 851, 977, 1020, 1091) go through it. A half-fix here would have produced false EXTERNAL warnings on our own respawns; it is complete.
- `INSTALL_DIR` is set at channels.sh:20, well before `respawn_log()` at ~1003 and its two call sites — no unbound-path write.
- Both watchdog exit branches are mirrored (`never-started`, `died-after-up`) — the two `respawn_log` calls sit on the `RESTART_REQUESTED=1` paths, not on the grace-warning path that does not exit.
- `checkExternalMainRespawn()` is called at the top of the sweep, outside the plugin-probe branch — so it still reports while the plugin is structurally down, which is the whole point on an AVX-less host.
- BSD-`wc` fix present and load-bearing: `tr -d '[:space:]'` before the numeric guard, otherwise the padded count falls into `*[!0-9]*` → `_lines=0` → trim never fires on macOS.
- Named false-positive analysis is in the PR body as requested — manual launch and watchdog-timer respawns warn **deliberately**, nothing silently filtered. That was my review condition and it is met.

### NOT verified by me — these remain the author's numbers

- **278 files / 3743 tests green, `tsc` clean, `npm run build` exit 0.** I did not re-run the suite; suite execution is the merge-preview gate, not this review. `gh pr checks 1006` reports **`scan pass` only** — the secret-gate. There is no independent CI evidence for the test numbers on this head.
- I did not exercise the detector on a live AVX-less box; the runtime behaviour is argued from the code path, not observed on hermes.

PR #1007, opened 2026-08-19
fix(heartbeat): serve the hot-memory count from the dashboard, stop prescribing the query (HBMEMBLIND819)

Szotasz:
## HBMEMBLIND819: the heartbeat's hot-memory count is now a served number, not a prescribed query

**Finding (measured):** the heartbeat's "new hot memories (1h)" line reported 0 for 14/14 rounds over 24h while the real value was 2 in three rounds. Raw evidence from the heartbeat's own transcript: it ran the prescribed query SHAPE but with `agent_id='heartbeat'` substituted for the main agent's id -- output 0, structurally, since the heartbeat never writes hot memories. The on-disk prescription (agents/heartbeat/CLAUDE.md) was correct.

**This is the second failure of prescribe-the-query for this exact metric.** HBMEMBLIND807 (08-07) was self-composed SQL; the fix shipped a ready-made query with "do not rewrite the query". Timeline over 8 sessions / 196 runs: 176 correct, 20 rewritten -- every rewrite appears on a post-compact round (the agent reconstructs the query from memory as "count MY hot memories" instead of re-reading the prescription); earlier sessions self-corrected within 1-2 rounds, the current one locked the wrong form in as its own precedent since 08-18 10:12.

**The closure copies the pattern that measurably held on the same agent:** the kanban counts never drifted, because an endpoint number has no query to rewrite (the db.ts comment for `getHeartbeatKanbanSummary` already names this principle: "the real fix is that the heartbeat agent no longer writes its own query").

- `db.ts`: `HEARTBEAT_NEW_HOT_MEMORIES_SQL` (exported shipped statement) + `countNewHotMemories(agentId)`.
- `/api/kanban/heartbeat-summary`: `counts.new_hot_memories_1h`, computed server-side with `MAIN_AGENT_ID`.
- Scaffold: the memory bullet now says copy that field from the call the agent already makes for the kanban section; **no runnable hot-memory SQL remains in the prompt**; a missing field (older dashboard build) degrades to `nincs adat (a summary nem adja)` -- never to a self-run query, never to a 0.

### Evidence
- Full suite **279 files / 3747 tests green**; real tsc clean; build exit 0 (non-/tmp worktree).
- Fixture tests execute the SHIPPED SQL, including the exact measured wrong-id shape (`agent_id='heartbeat'` sees a different world -- which is why the id is supplied server-side).
- **Mutation red-runs, both seen failing:** (A) removing the endpoint field -> the endpoint wiring test fails alone; (B) restoring a runnable SQL to the scaffold -> the three scaffold guards fail. Restore -> green. Wiring windows are structurally anchored (start marker -> handler's own `return true` / rendered-output matching), not derived from the sought string -- the #1006 review lesson applied.
- The old HBMEMBLIND807 prescription pins are rewritten to the new contract (they pinned the exact surface that drifted).

Rollout note: the scaffold CLAUDE.md regenerates on dashboard boot, so the fix is live only after build + restart on each install -- same caveat as #1006.

### DIFF
--- a/src/__tests__/heartbeat-agent-scaffold.test.ts
+++ b/src/__tests__/heartbeat-agent-scaffold.test.ts
@@ -208,29 +208,34 @@ describe('shouldBootHeartbeatAgent', () => {
   })
 })
 
-// HBMEMBLIND807: the hot-memory metric must ship as a READY-MADE query, the
-// way task_runs does -- a prose-only bullet let the heartbeat agent compose
-// its own SQL and report 0 with three hot memories in the window. Lock the
-// contract: the exact query, the SECONDS cutoff (no ms multiplier -- that is
-// the task_runs unit, not this one), and the do-not-rewrite instruction.
-describe('hot-memory metric is a ready-made query (HBMEMBLIND807)', () => {
-  it('ships the exact scoped count query', () => {
+// HBMEMBLIND807 -> HBMEMBLIND819: the hot-memory metric went through TWO
+// contracts, and both failures are why the current one exists. 807: a
+// prose-only bullet let the agent compose its own SQL (reported 0 beside 3
+// hot memories); the fix shipped a ready-made query with "do not rewrite the
+// query". 819: that failed too -- post-compact rounds reconstructed the query
+// from memory with agent_id='heartbeat' and reported 0 for 24h straight
+// (14/14, real value 2 in three rounds). Current contract: the number is
+// computed server-side (countNewHotMemories, served as
+// counts.new_hot_memories_1h on /api/kanban/heartbeat-summary) and the
+// scaffold tells the agent to COPY it -- there is no query left to rewrite.
+describe('hot-memory metric is an endpoint number, never an agent-run query (HBMEMBLIND819)', () => {
+  it('points the agent at counts.new_hot_memories_1h from the heartbeat-summary call', () => {
     const out = renderHeartbeatClaudeMd(ID)
-    expect(out).toContain("SELECT COUNT(*) FROM memories")
-    expect(out).toContain(`agent_id='${ID.mainAgentId}'`)
-    expect(out).toContain("category='hot'")
-    expect(out).toContain('created_at > unixepoch()-3600')
+    expect(out).toContain('counts.new_hot_memories_1h')
   })
 
-  it('the memory cutoff carries NO millisecond multiplier', () => {
+  it('ships NO runnable hot-memory SQL anywhere in the prompt', () => {
     const out = renderHeartbeatClaudeMd(ID)
-    const memBullet = out.slice(out.indexOf('Memory + system'))
-    expect(memBullet.slice(0, 1200)).not.toContain("(unixepoch()-3600)*1000")
+    // The exact surface that drifted twice: a memories/hot query the agent
+    // could run (and, measured, rewrite). Shape-agnostic: any SQL touching
+    // the memories table near a hot filter is out of contract.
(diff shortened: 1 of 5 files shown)


Szotasz:
## Review (independent second pair of eyes) — GO, one non-blocking note

**Process note:** the fleet runs on a single GitHub identity, so `gh pr review --approve` is rejected as "your own pull request". The independent review is an *organizational* fact, not a GitHub state: **author = Samu, reviewer = Marveen (who reported the finding)**, and the measurements below are ones I ran myself against head `826d4306`.

The diagnosis is right and the closure is the correct one. Two prescriptions failed for this metric; the third attempt is not another prescription, which is the whole point.

### Verified myself

**1. There is no runnable hot-memory SQL left in the prompt.** Fetched the scaffold at the PR head and grepped every `memories` occurrence: seven hits, all prose or history (`HBMEMBLIND807` narrative, the `nincs adat` fallback, the output-format line). No `sqlite3` invocation, no `FROM memories`. The surface that drifted twice is gone rather than reworded.

**2. "The same call you already made" is literally true.** The Kanban bullet already fetches `${dashboardOrigin}/api/kanban/heartbeat-summary` (scaffold line 211), and the Memory bullet now reads `counts.new_hot_memories_1h` off that same response (line 260). No new call, no new failure mode.

**3. The wiring window is structurally anchored — I ran the mutation.** This is where the morning's `#1006` lesson had to land, so I checked it rather than trusting that it had:

```
ORIGINAL : wiring test PASSES=True   | window=882 chars
MUTATED  : wiring test PASSES=False  | window=817 chars   (field removed from the handler)
```

The window is bounded by the handler's own `return true`, so it **shrinks** under the mutation instead of stretching to swallow the needle. That is the correct shape.

**4. The shipped statement produces the right number on the REAL database, not just a fixture.** The fixture test proves the mechanism; this answers the other question — would it actually have caught the incident? I ran `HEARTBEAT_NEW_HOT_MEMORIES_SQL` against the live store, bound to the three rounds where the heartbeat reported `0`:

```
2026-08-18 18:00:31 -> 2
2026-08-18 19:00:22 -> 2
2026-08-19 14:00:23 -> 2
```

Exactly the values that were missed. The shipped SQL reproduces the ground truth on live data.

**5. `MAIN_AGENT_ID` was already imported** in `routes/kanban.ts`; the new call adds no config surface.

### NON-BLOCKING — the one place a zero can still be manufactured

```ts
return row?.n ?? 0
```

For a `SELECT COUNT(*)` the row is never absent, so this is defensive dead code — but it is the single remaining branch in the chain where a `0` is *produced* rather than *measured*, in a PR whose thesis is that a line which can silently read 0 is worse than no line. Everywhere else the fail direction is right (missing field → `nincs adat`, never 0). Consider `?? null` with the field omitted, or letting it throw, so the impossible case degrades the same way the possible one does. Not worth another round on its own — fold it into your next touch of this file.

### NOT verified by me

**279 files / 3747 tests, `tsc` clean, `build` exit 0, and both mutation reds** are your measurements. I re-ran the wiring mutation myself (above) but not the suite; `gh pr checks 1007` reports **`scan pass` only**, so there is no independent CI evidence for the test counts on this head.

### Merge conditions carried forward (both yours, both correct)

1. **Rollout:** the scaffold regenerates at boot, so this is inert until build + restart. Acceptance is the **first non-zero heartbeat line in a known non-zero window** — a positive control, not a green build.
2. **Process lesson:** mutation-restore via checkout is only safe on a committed base. Worth keeping; that one bites quietly.

PR #1021, opened 2026-08-20
feat(seed-skills): Bridge service-tabs discovery skill (BRIDGESVCSKILL820)

Szotasz:
## What

Seed skill `bridge-szolgaltatas-lapok`: when an agent starts/installs something listening on a LOCAL PORT on the Marveen machine, it offers the Bridge **Szolgáltatás-lapok** path to its owner at that moment; with no paired Bridge it mentions the app once, politely, linking the download PAGE.

## The blocking measurement (came back green, no STOP needed)

The default `update.sh` run **seeds MISSING skills** (skip-if-exists, header comment lines 65-67), so existing installs receive this skill on their next update -- this is exactly why it is a seed skill and not a downloadable doc. `--reseed-fleet` covers future corrections. `scripts/__tests__/seed-skills.test.sh` iterates `seed-skills/*` generically: **20/20 green with the new skill present**.

## The four stipulations, verifiable in the text

1. **Page link only** (`aiamindennapokban.hu/marveen#bridge`, 2 occurrences, zero file URLs) -- the email+consent step records the lead.
2. **Measured branch signal**: `grep -c "marveen-remote:" ~/.ssh/authorized_keys` (the enroll writes these restricted lines; ≥1 = a device was paired). The no-signal branch wording deliberately never claims the owner has no Bridge.
3. **Once only**: memory check before, memory note after, never re-raised after a no -- stated as a hard rule.
4. **Labels from shipped 0.3.1** `lang.js` (Kapcsolat / Szolgáltatás-lapok / Hozzáadás + the reconnect-latency sentence), not from memory.

Copy gate: 0 em-dash, full accents, Szabi-voice owner-facing snippets. Deeplink (BRIDGEDEEPLINK820) intentionally NOT included.

### DIFF
--- a/seed-skills/bridge-szolgaltatas-lapok/SKILL.md
+++ b/seed-skills/bridge-szolgaltatas-lapok/SKILL.md
@@ -15,13 +15,21 @@ trigger egy már régóta futó szolgáltatás puszta létezése sem. A funkció
 akkor látszik, amikor ÉPP születik egy új helyi szolgáltatás.
 
 ## 1. lépés: mérd meg, van-e párosított Bridge (ne találgass)
+A párosítás két nyomot hagy a Marveen oldalán; az elsődleges a device_keys
+tábla (az enroll írja, a kulcspár rekordja), a tartalék az authorized_keys:
 ```bash
-grep -c "marveen-remote:" ~/.ssh/authorized_keys 2>/dev/null || echo 0
+sqlite3 ~/ClaudeClaw/store/claudeclaw.db "select count(*) from device_keys" 2>/dev/null \
+  || grep -c "marveen-remote:" ~/.ssh/authorized_keys 2>/dev/null || echo 0
 ```
-- 1 vagy több: van párosított Bridge-eszköz ezen a telepítésen. -> (A) ág.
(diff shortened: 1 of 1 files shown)


Szotasz:
Amended per msg 13866 (crossed with the PR): **device_keys is the primary pairing signal** (the enroll writes both the table row and the authorized_keys line; src/db.ts:930), authorized_keys grep stays as fallback. Added the has-vs-uses precision as an explicit rule: the signal proves a pairing HAPPENED, never that the Bridge is in use -- the (A) branch gives the how-to path and never claims observed usage. Copy gate re-checked (0 em-dash, page link only), seed test 20/20.

Szotasz:
Review blocker fixed + owner expansion in (head `72f88eff`): the pairing probe is now the **authorized_keys grep alone** (no sqlite3 in any fenced block -- the recipe runs on customer Linux where sqlite3 is absent; device_keys stays as prose context). `shipped-recipes-no-sqlite-cli`: **6/6 green**, seed suite **20/20**. Three triggers per TG 12871: port event / one early post-install conversation / the question path (frontmatter description rewritten so the skill loads on convenience-access questions). Once-rule is per-branch; questions are always answered. Copy gate re-verified (0 em-dash, page link only).

PR #1028, opened 2026-08-21
fix(hooks): the reply guard never blocks, and cwd can invent an agent id

attila-fiREG:
Two related defects in the Stop-hook path. Both reproduce on a clean checkout of `main` (v1.33.0), and the shipped test for the first one is red there.

## 1. The reply guard has never blocked

`scripts/hooks/telegram-reply-guard.py:112` unpacks five values:

```python
chat_id, message_id, text, ts, created_at = oq
```

but `ledger_lib.open_question_with_age()` returns seven — the inbound's `attachment_kind` and `attachment_file_id` are part of the row (`ledger_lib.py:225`).

The unpack sits **outside** the `try/except` above it, so every invocation dies with `ValueError: too many values to unpack (expected 5)`, exits non-zero with empty stdout, and the harness reads that as "allow".

Reproduce on `main`:

```
$ python3 scripts/__tests__/telegram-reply-guard.test.py
3 FAILED: ['unanswered question blocks', 'maxblock #1 blocks', 'maxblock #2 blocks']
```

The effect is quiet, which is the worst part: the rule the hook exists to enforce looks enforced and is not. Fix takes only the prefix the hook needs, so a future column addition cannot wedge it again.

## 2. A subdirectory cwd invents an agent id

`agent_id_from_cwd()` recognises exactly two shapes — the install root and `<install>/agents/<id>` — and falls through to "last path component" for everything else. A session whose cwd happens to be a **subdirectory of the install** (a scratch dir, a build folder) therefore logs its messages under an agent id invented from the directory name.

Observed: a reply sent while the shell sat in `store/<workdir>` landed in `conversation_log` with `agent_id='<workdir>'`.

Two consequences: the conversation ledger splits across two identities (degrading replay/handoff), and — once (1) is fixed — the guard finds no outbound under the real id and blocks a turn on a question that was already answered. That is in fact how this was found: the guard's first working invocation caught it.

Anything inside the install tree now resolves to the main agent. The `agents/<id>` branch still wins for sub-agents including their own subdirectories, and the basename fallback is kept for cwds outside the install.

## Tests

New `scripts/__tests__/ledger-agent-id.test.py`:

| Case | Expect |
|---|---|
| install root | main agent |
| `store/<workdir>` subdirectory | main agent |
| deep subdirectory | main agent |
| trailing slash | main agent |
| `agents/dia` | `dia` |
| `agents/dia/x/y` | `dia` |
| outside the install | basename |
| empty / None | main agent |

All four python contract tests pass on this branch.

### DIFF
--- a/scripts/__tests__/ledger-agent-id.test.py
+++ b/scripts/__tests__/ledger-agent-id.test.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+"""Test agent_id resolution from the session cwd (scripts/hooks/ledger_lib.py).
+
+Regression guard: a session sitting in a SUBDIRECTORY of the install logged its
+outbound messages under an agent id taken from the directory name. That splits
+the conversation ledger across two identities and makes the reply guard block on
+an already-answered question, because it finds no outbound under the real id.
+
+Run: python3 <thisfile>   Exit 0 = all pass.
+"""
+import os
+import sys
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+HOOKS = os.path.join(os.path.dirname(HERE), "hooks")
+sys.path.insert(0, HOOKS)
+
+INSTALL = os.path.dirname(os.path.dirname(HERE))
+os.environ.setdefault("MAIN_AGENT_ID", "mainagent")
+
+import ledger_lib  # noqa: E402
+
+MAIN = ledger_lib.main_agent_id()
+
+CASES = [
+    ("install root is the main agent", INSTALL, MAIN),
+    ("subdirectory is STILL the main agent", os.path.join(INSTALL, "store", "some-workdir"), MAIN),
+    ("deep subdirectory is still the main agent", os.path.join(INSTALL, "a", "b", "c"), MAIN),
+    ("trailing slash tolerated", INSTALL + "/", MAIN),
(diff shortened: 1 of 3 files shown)


Szotasz:

Szotasz:
Merged to `develop` as `8fa255c0`. Thank you for this one, and for the reproduction steps: they are what made it quick to verify.

Three notes on what happened to the PR mechanically, so none of it reads as a rejection.

**Retargeted from `main` to `develop`.** Our release flow is `develop -> main` on a fixed cadence. A fix landing on `main` alone would have been overwritten by the next promote, so the branch that carries it forward is `develop`. It ships to `main` with the next release.

**Closed and reopened, with no push.** A required status check was added to `develop` on 2026-08-18 through a ruleset. It had never run on this PR, because the PR targeted `main` until today, so the merge state was `BLOCKED` with an empty check list. Reopening re-triggers the workflow on the unchanged head. Your commit is byte identical, `20c0925b` before and after, and both checks then passed.

**Merged with rebase, not squash,** so the commit keeps you as the author rather than the person pressing the button.

On the finding itself: your diagnosis holds on our side too. The guard has been dead since 2026-08-11, when the seven value return landed two minutes before the five value unpack. Every open question died with a ValueError and the harness read the empty stdout as an allow, so the check we built to enforce answering was never actually installed. Thirteen days, and nothing surfaced it until you looked.

PR #1039, opened 2026-08-22
fix(outgoing-copy-gate): a quoted span is evidence, not the author's prose (GATEIDEZET822)

mogganhun:
## What happened

The gate blocked the main agent's **own** outgoing message, because that message quoted the accent-stripped sentence from the 2026-08-10 incident it was explaining to the owner. Observed live on 2026-08-22: the reply came back rejected with `HIANYZO EKEZETEK, 2 szo`, pointing at words that existed only inside the quotation marks.

## Why that is a defect, not a strict gate

The gate audits what the **author** wrote. A verbatim quote is not the author's writing, and showing the wrong form is frequently the entire reason the sentence exists. The same applies to a complaint pasted from a customer, a log line, or a name someone else spelled wrong: "correcting" quoted text would be an error, not a fix.

## The change

Quoted spans are masked before every check (accents, em dash, double hyphen, name rules, homoglyphs). Recognised: Hungarian `„…”`, typographic `“…”`, straight `"…"`, both guillemet directions, and markdown blockquote lines.

## Keeping the exemption from becoming a bypass

- **Same-length mask.** `_hit_context()` reports findings by offset into the same string, so shrinking the text would shift every later finding's quoted context. The mask writes spaces, not deletions.
- **Outside the quote still blocks.** A wrong form that appears in the author's own sentence is reported even if it is also quoted elsewhere. Without this, quoting a bad word once would disable the check for it. Covered by a test.
- **The skip is not silent.** A finding dropped by masking is written to the gate log with its text, so an unexpectedly permissive run is auditable afterwards.

Two guards bound the masking itself, both tested: a 1000-character cap so a stray opening quote cannot swallow the rest of the message, and a paragraph break inside a pair cancels it (two unrelated quote marks, not one quotation).

## Verification

- `outgoing-copy-gate-tokens.test.ts`: **12/12**. The 6 existing GATEKOTOJEL817 / GATEHYPH816 tests are untouched and green; 6 new cases cover quoted accents, the outside-the-quote case, straight quotes, blockquote lines, an unclosed quote, and the paragraph-break cancel.
- `tsc --noEmit` clean.
- 10 live hook runs against a real `gmail_send_email` payload. Blocks: a name in the author's prose, missing accents, an em dash, and the mixed case. Passes: quoted names, quoted accent-stripped text, a blockquote line, and a clean letter.

Run from a worktree under `$HOME`, per the suite's own live-install guard.

### DIFF
--- a/scripts/hooks/outgoing-copy-gate.py
+++ b/scripts/hooks/outgoing-copy-gate.py
@@ -695,9 +695,95 @@ def telegram_gate(tool_input: dict) -> None:
     sys.exit(0)
 
 
-def audit(text: str):
-    """Return a list of human-readable problems."""
+# --- QUOTED SPANS ARE EVIDENCE, NOT PROSE (GATEIDEZET822, 2026-08-22) --------
+# LIVE CASE: the gate blocked the main agent's OWN Telegram message, because the
+# message QUOTED the accent-stripped sentence from the 2026-08-10 incident it was
+# explaining. The gate exists to audit what the AUTHOR wrote; a verbatim quote is
+# not the author's writing, and showing the wrong form is often the entire point
+# of the sentence around it. Quoted spans are therefore masked out of EVERY check
+# (accents, em dash, double hyphen, name, homoglyph).
+#
+# The mask writes SAME-LENGTH spaces on purpose: _hit_context() reports by offset
+# into the same string, so shrinking the text would move every later finding's
+# quoted context off by the removed characters.
+#
+# Two guards keep the exemption from becoming a bypass:
+#   - the same wrong form OUTSIDE the quote still blocks (covered by a test), and
+#   - a finding dropped by masking is written to the gate log, never silent.
+#     A hole nobody knows about protects only until someone steps in it.
+QUOTE_PAIRS = (
+    ("\u201e", "\u201d"),   # Hungarian: low-9 open, right-double close
+    ("\u201c", "\u201d"),   # typographic English
+    ("\u00bb", "\u00ab"),   # Hungarian guillemets (pointing outward)
+    ("\u00ab", "\u00bb"),   # French guillemets
+    ('"', '"'),             # straight double quote
+)
+# No real quotation runs longer than this; the cap stops a stray opening quote
+# from swallowing the rest of the message and disabling the gate wholesale.
+QUOTE_MAX = 1000
+BLOCKQUOTE = re.compile(r"^[ \t]*>[^\n]*$", re.M)
+
+
+def mask_quoted(text: str):
+    """Blank out quoted spans. Returns (masked_text, span_count)."""
+    chars = list(text)
+    n = len(text)
+    spans = 0
+
+    def blank(a, b):
+        for k in range(a, b):
+            if chars[k] != "\n":
+                chars[k] = " "
+
+    # 1. markdown blockquote lines
+    for m in BLOCKQUOTE.finditer(text):
+        blank(m.start(), m.end())
+        spans += 1
+
+    # 2. quote pairs
+    openers = {pair[0]: pair[1] for pair in QUOTE_PAIRS}
+    i = 0
+    while i < n:
+        closer = openers.get(text[i])
+        if closer is None:
+            i += 1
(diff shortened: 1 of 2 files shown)


Szotasz:
Szia, es elnezest a hosszu csendert. 11 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

Szotasz:
Thanks for this, and sorry for the long silence: the card carrying this review sat with an expired hold, which is my fault, not yours. Measured today: the branch moved on 2026-09-03, the card's last substantive note is 2026-08-24, and the hold date on it expired on 08-27.

The implementation is careful. The unclosed-quote cap, the offset correction after masking, and the test that the same wrong form outside a quote still blocks are all the right details to have thought about. My objection is not to the code.

**It is to the premise, and I want to state it precisely so you can push back on it.**

The comment says the gate exists to audit what the author wrote, and that a verbatim quote is evidence rather than prose. That is true of authorship. It is not true of the thing the gate actually protects, which is what leaves the house and lands in someone else's inbox. A lookalike character inside a quotation is still a lookalike character in the message the recipient receives. The reader has no way to know that a given span was quoted, and the harm does not depend on who typed it first.

So masking quoted spans out of **every** check trades a real property for a convenience.

**And the convenience has a cheaper substitute, which we measured yesterday.** When the gate fires on a review comment whose whole subject is a bad character, the fix is to describe the position instead of reproducing the character: name the file, the list item, and where in the line it sits. The author finds it just as fast, the gate stays at zero, and no exemption is needed. That is now written into our review procedure, and it has held every time since.

**What I would merge instead, if you want to carry it:** keep the masking, but scope it to the checks where the quoted origin genuinely changes the verdict, and never apply it to the ones that are about what the recipient receives. Concretely, a typed-clock or style check can reasonably ignore a quoted span. An accent check and a mixed-script check cannot, because their failure mode is the character arriving at the other end.

If you think the distinction does not hold, say so and I will take it seriously. The part I am confident about is narrow: `mask_quotes` defaulting to true across all checks is the specific piece I do not want to merge.

One process note, and it is on me: the hold on our card expired ten days ago and nothing surfaced it. That is a gap in our sweep, not a judgement about your PR, and I have written it up on our side.

Szotasz:
Lemértük a friss fejen (1dd2b085, develop 2282dcab-ra merge-ölve, tiszta). A szándék jó, és a PR saját tesztjei is jók (pozitív és negatív irány egyaránt), de a kaput mérhetően teljesen ki lehet kapcsolni, ezért ebben az alakban nem mehet.

A valódi audit() függvényen mérve: a kontroll-próza 2 találatot ad (blokkol), ugyanaz a tartalom viszont 0 találatot, ha (1) az egész üzenet egyetlen idézőjel-párban áll, (2) minden sor "> " prefixet kap, (3) egy elkóborolt idézőjel-pár ~900 karaktert fed le, vagy (4) az ékezet nélküli próza "> " mögé kerül. Vagyis egy blockquote-prefix az em-dash-, az ékezet- ÉS a név-szabályt egyszerre némítja, és a QUOTE_MAX=1000 nem véd, mert egy átlagos üzenet rövidebb.

Három olcsó kérés:
1. Az em-dash ellenőrzés soha ne maszkolódjon (álló "soha" szabály; valódi em-dash idézése elhanyagolhatóan ritka).
2. A maszkolás kapjon arány-korlátot az abszolút 1000 karakter helyett: ha a maszkolt szakasz a szöveg több mint kb. felét fedi, ne maszkoljon. Ez mind a négy fenti vektort megfogja.
3. Két hiányzó negatív teszt: végig idézőjelezett üzenet és végig "> " prefixes üzenet, mindkettő bukjon.

Két kisebb megjegyzés: az új `_log_gate()` mellett a fájlban már él a `_gate_log()` + `_GATE_LOG` (446-449. sor), és éppen az új írja a sorokat dátum nélkül a fájl datált konvenciója ellenében; illetve nem létező log-könyvtárnál az írás csendben elmarad (OSError: pass), mérve.

PR #1050, opened 2026-08-23
fix(gmail-mcp): working multi-user routing for gmail-mcp-server

akobza:
## Summary

`gmail-mcp-server@1.0.30`'s `--multi-user` mode advertises a `userId`
parameter on every Gmail tool, but the shipped handlers never actually
used it: search, read, send, drafts, label moves, mark-as-read, and
attachments all always went through the single legacy account,
regardless of the `userId` passed in (`gmail_send_email` even fetched
the per-user client via `multiUserAuth.getGmailClientForUser(userId)`
and then discarded it, sending through the legacy account anyway).
Net effect: authenticating a second account via `gmail_authenticate_user`
did nothing observable — every call kept silently using whichever
account ran `gmail_authenticate` first, which is a genuinely dangerous
failure mode for `gmail_send_email` (mail can go out from the wrong
identity while reporting success).

The package has no public upstream repo referenced from its `package.json`
(the README's issue link is a literal `yourusername` placeholder), so this
vendors and patches it into `vendor/gmail-mcp-server/` rather than opening
a PR against a repo we can't find.

## What changed

- `vendor/gmail-mcp-server/dist/utils/gmail-operations.js`: every method
  that reaches the Gmail API takes an optional trailing `overrideClient`
  and uses it instead of the legacy singleton when present; internal
  cross-calls (enhanced/traditional search calling `getEmail`, `listDrafts`
  calling `getDraft`) now thread the same override through.
- `vendor/gmail-mcp-server/dist/utils/multi-user-auth.js`:
  `getGmailClientForUser(userId)` now also matches by plain email address,
  not only the generated `user_<hash>_...` session id.
- `vendor/gmail-mcp-server/dist/index.js`: the 9 affected tool handlers
  now resolve and pass the per-user client when `userId` is given, and
  fall back to the existing default account when it's omitted (the old
  `"'userId' is required in multi-user mode"` throws are gone) — so
  callers that don't know about multi-user mode keep working unchanged.
  The duplicated multi-user/single-user branches in the `gmail_draft`
  handler collapsed into one, since both did the same discard-the-client
  thing anyway.
- `ATTRIBUTIONS.md`: credits the vendored package and documents the patch.
- `mcp-catalog.json`: points the `gmail` catalog entry at the vendored,
  patched copy instead of `npx -y gmail-mcp-server`.
- `.gitignore`: carves `vendor/gmail-mcp-server/dist/` out of the blanket
  `dist/` ignore (same directory-vs-contents issue already solved for
  `.claude/` further down the file) so the patched build actually ships.

Full before/after breakdown in `vendor/gmail-mcp-server/MARVEEN_PATCH.md`.

## Test plan

- [x] `node --check` on all three patched files
- [x] Vendored server boots cleanly in `--multi-user` mode (dummy
      credentials, confirms module resolution + startup, not a live
      OAuth round-trip)
- [ ] Live dual-account round trip (authenticate two real accounts,
      confirm `gmail_search_emails`/`gmail_send_email` with distinct
      `userId`s actually hit different mailboxes) — in progress against
      a live install, not part of this diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,12 @@
 # Dependencies & build
 node_modules
 dist/
+# Vendored gmail-mcp-server ships its patched build under its own dist/ --
+# carve it out of the blanket dist/ ignore above (same directory-vs-contents
+# gotcha as .claude/ further down: needs the directory itself un-ignored
+# before the negation can reach its contents).
+!vendor/gmail-mcp-server/dist/
+!vendor/gmail-mcp-server/dist/**
 
 # Python bytecode cache (e.g. scripts/hooks/__pycache__/*.pyc)
 __pycache__/
--- a/ATTRIBUTIONS.md
+++ b/ATTRIBUTIONS.md
@@ -19,6 +19,13 @@ A Marveen saját licensze: [LICENSE](./LICENSE) (MIT).
 - **Hol a Marveen-ben**: `seed-skills/handoff/`, `seed-skills/retrospective/`, `seed-skills/skill-management/`
 - **Mit csinál nálunk**: a skill-csontváz a Marveen flotta-architektúrájára adaptálva. Az 5-szekciós handoff struktúra (Goal, Current Progress, What Worked, What Didn't Work, Next Steps), a sub-agent retrospective minta, és a skill-rot életciklus-kezelés koncepció mind Zhutov csomagjából származik. A megvalósítás TypeScript-ben újraírva, integrálva a Marveen checkpoint, DREAM, memória és inter-agent message rendszereivel.
 
+### gmail-mcp-server (Gmail MCP szerver, több-fiókos javítással)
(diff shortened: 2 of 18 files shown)


Szotasz:
Szia, es elnezest a hosszu csendert. 10 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

akobza:
Elavult, 113 commit-tal le van maradva a jelenlegi develop mögött. A helyi javítás időközben túlment ezen (a vendorolt multi-user routing fix mellé egy azóta talált követő hibát is javít: a gmail-private szerver ugyanazt a lejárt token fájlt olvasta, mint a gmail szerver). Lásd #1162.

PR #1056, opened 2026-08-24
refactor(kanban): remove never-wired full-board context dump (KANBANCTXDEAD824)

Szotasz:
## What

Removes `buildKanbanContext()` (src/memory.ts) and its only dependency `listKanbanCardsSummary()` (src/db.ts).

## Why

The function rendered EVERY unarchived card's FULL title into a `[Kanban tabla]` block — no limit, no truncation. On the live board that is 822 cards / ~0.96 MB of titles (max single title 16,428 chars). Had it ever run against a 200k-token window it would have overflowed it several times over.

It never ran: verified **zero callers** in `src/`, `dist/` (the running build) and all plugins. The mechanism claim "the kanban table is injected into every agent message" is refuted at the wiring step — the scary SQL exists but is dead code.

Removed rather than hardened: an unbounded full-board dump is the wrong contract for agent context. A tombstone comment points any future kanban-context work at the heartbeat-summary shape (counts first, server-side title truncation, HBKANBANDRIFT819 / #1010).

## Evidence

- `grep -rn 'buildKanbanContext' src/ dist/ ~/.claude/plugins` → only the definition, no call site
- `tsc --noEmit`: exit 0
- Full suite: **3935/3935** (293 files) on this branch

Card: KANBANCTXDEAD824

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -1702,12 +1702,6 @@ export function listKanbanCards(): KanbanCard[] {
     .all() as KanbanCard[]
 }
 
-export function listKanbanCardsSummary(): { status: string; title: string; assignee: string | null; priority: string; id: string }[] {
-  return db
-    .prepare("SELECT id, title, status, assignee, priority FROM kanban_cards WHERE archived_at IS NULL ORDER BY status, sort_order ASC")
-    .all() as any[]
-}
-
 export function getKanbanCard(id: string): KanbanCard | undefined {
   return db.prepare('SELECT rowid AS seq, * FROM kanban_cards WHERE id = ?').get(id) as KanbanCard | undefined
 }
--- a/src/memory.ts
+++ b/src/memory.ts
@@ -10,7 +10,6 @@ import {
   pruneAuditLogs,
   pruneTokenUsage,
   getMemoriesForChat,
-  listKanbanCardsSummary,
   type Memory,
 } from './db.js'
 import { runAgent } from './agent.js'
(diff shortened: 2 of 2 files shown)


Szotasz:
Approving. The deletion is the right call, and I verified the premise independently rather than restating it.

**Zero callers, with a positive control.** `grep -rn buildKanbanContext src/ dist/ scripts/ .channels-config/` returns two hits, both the definition itself (`src/memory.ts:120`, `dist/memory.js:95`). The same command on `listKanbanCardsSummary` returns six (definition, import, call site, in both `src` and `dist`), so the instrument does find call sites when they exist. The zero is a real zero, not a silent grep.

**Removal over hardening is correct.** An unbounded full-board dump is the wrong contract regardless of what cap we put on it, and a hardened version would have looked maintained while still being dead. The tombstone earns its place: it names the size, the reason, and where the next person should start.

One note for the record, since it cost us an afternoon. Both Samu and I first confirmed this function as the cause of the heartbeat context churn by reading its *body* (no limit, no truncation) without checking whether anything *calls* it. The arithmetic had already refuted us and neither of us noticed: a fresh heartbeat session starts at ~55k tokens and one round costs ~40k, while this block would have been ~240k. A mechanism that does not fit the measured budget is not the mechanism.

The real cause turned out to be the scheduled-task prompt itself: 84 KB, of which 44 KB was a "Marveen discipline" section that the runner is explicitly forbidden to act on. That is now in `references/`, and the prompt is 12 KB.

**Merge timing:** after the 22:00 release freeze on `develop`, not before.

PR #1057, opened 2026-08-24
fix(ledger): never invent an agent id from a subdirectory cwd

attila-fiREG:
## Symptom

`agent_id_from_cwd()` in `scripts/hooks/ledger_lib.py` recognised only two cwd shapes — the install root and `<install>/agents/<id>` — and fell through to "last path component" for everything else:

```python
if cwd == install:
    return main_agent_id()
# Fallback: last path component
base = os.path.basename(cwd)
return base or main_agent_id()
```

So a session whose cwd happened to be a **subdirectory** of the install (a scratch dir, a build folder) logged its messages under an agent id **invented from the directory name**.

## When it broke

Present since `agent_id_from_cwd` was introduced with the conversation-continuity ledger — the fallback has always derived an id from the directory basename for any in-tree subdirectory.

## Reproduction

```sh
# main agent id is 'mainagent'; call the resolver with a cwd that is a
# subdirectory of the install:
MAIN_AGENT_ID=mainagent python3 -c "import sys; sys.path.insert(0,'scripts/hooks'); \
import ledger_lib as l; print(l.agent_id_from_cwd(l._install_dir()+'/store/wd'))"
# before: 'wd'         <- a bogus agent id taken from the directory name
# after:  'mainagent'
```

## What the user sees

Two failures downstream of the wrong id:

1. The conversation ledger splits across two identities, degrading replay/handoff.
2. The reply guard finds no outbound logged under the *real* agent id, so it blocks a turn on a question that was in fact already answered.

## Fix

- Anything **inside** the install tree resolves to the main agent (`cwd == install or cwd.startswith(install + os.sep)`).
- A cwd **outside** the tree never fabricates an id from the directory name: a launcher may name the session explicitly via `MARVEEN_AGENT_ID`, otherwise it is attributed to the main agent.
- The `agents/<id>` branch is unchanged.

Includes a regression test (`scripts/__tests__/ledger-agent-id.test.py`) covering in-tree subdirectories, out-of-tree cwd, and the env override.

### DIFF
--- a/scripts/__tests__/ledger-agent-id.test.py
+++ b/scripts/__tests__/ledger-agent-id.test.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
 """Test agent_id resolution from the session cwd (scripts/hooks/ledger_lib.py).
 
-Regression guard: a session sitting in a SUBDIRECTORY of the install logged its
-outbound messages under an agent id taken from the directory name. That splits
-the conversation ledger across two identities and makes the reply guard block on
-an already-answered question, because it finds no outbound under the real id.
+Regression guard: a session sitting in a SUBDIRECTORY of the install used to log
+its outbound messages under an agent id INVENTED from the directory name. That
+splits the conversation ledger across two identities and makes the reply guard
+block on an already-answered question, because it finds no outbound under the
+real id. Anything inside the install tree must resolve to the main agent, and a
+cwd outside the tree must never fabricate an id from the directory name.
 
 Run: python3 <thisfile>   Exit 0 = all pass.
 """
@@ -17,6 +19,7 @@ sys.path.insert(0, HOOKS)
 
 INSTALL = os.path.dirname(os.path.dirname(HERE))
 os.environ.setdefault("MAIN_AGENT_ID", "mainagent")
+os.environ.pop("MARVEEN_AGENT_ID", None)
 
 import ledger_lib  # noqa: E402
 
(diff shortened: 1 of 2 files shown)


Szotasz:
Thank you for this, and sorry for the noise you are about to read: part of this PR has already landed, and that is our doing, not yours.

**What happened.** The inside-the-tree half of your change is already on `develop`. It arrived through your earlier PR (`fix/ledger-agent-id-subdir-upstream`), which we merged this morning. That is why this branch now shows as conflicting: the same fix is present twice, once from each PR.

**What is still wanted, and it is the more interesting half.** Your outside-the-tree change is not on `develop`. The current code still ends with

    base = os.path.basename(cwd)
    return base or main_agent_id()

so a session whose cwd sits outside the install still gets an agent id invented from the directory name. Your replacement, honouring `MARVEEN_AGENT_ID` and otherwise attributing to the main agent, closes that. On our own install the difference happens to be invisible, because the main agent id and the directory basename coincide, which is exactly the kind of coincidence that keeps a bug alive.

Your regression test is wanted too. We spent today finding several invariants that were true by construction and pinned by nothing, so a test that names the cases is worth more to us than the fix it guards.

**What we would ask.** Rebase onto `develop` and keep only the outside-the-tree change plus the test. If you would rather not, say so and we will cherry-pick the delta ourselves and credit you on the commit.

**One process note, and this one is on us.** Several of your recent PRs target `main`. Our flow is `develop -> main` on a fixed cadence, so a fix that lands on `main` alone gets overwritten by the next promote. We retarget them to `develop` when we see them, which is why you may notice the base changing under you. Nothing is wrong with your patches; we simply never told you which branch to open against, and we should have after the first one. Please open against `develop` from now on, and it will save us both a step.

attila-fiREG:
Done — the branch is rebased onto `develop` and now carries only the outside-the-tree half plus the test: `MARVEEN_AGENT_ID` is honoured when the launcher sets it, and anything else outside the tree attributes to the main agent instead of a basename. The commit message was rewritten for the narrowed scope.

On the test: the old `"outside the install falls back to basename"` case — which passed only through the exact basename coincidence you describe — is replaced by the out-of-tree and `MARVEEN_AGENT_ID` cases, so the old behaviour can no longer hide behind it. CI is green and the PR shows clean.

Noted on the process point: everything we opened from today onward targets `develop`, and we will keep it that way. And thank you for spelling out that a test naming the invariants is worth more than the fix it guards — we took that to heart; #1068, #1069 and #1071 are exactly that shape.

PR #1059, opened 2026-08-24
feat(kanban): title gate trigger relocates over-long titles into comments (KANBANCTXDEAD824)

Szotasz:
## What

Two self-healing SQLite triggers (`kanban_cards_title_gate_insert` / `_update`): any INSERT, or title-changing UPDATE, that writes a title over **300 chars** gets the FULL original title preserved verbatim as a marked comment on the same card, then the title cut to exactly 300 chars.

## Why this shape (GO conditions, Marveen msg 15082)

- **NOT a CHECK constraint**: agents write `kanban_cards` with raw sqlite3 and rarely inspect exit codes — a rejected INSERT would silently lose the card, worse than any long title. The trigger relocates instead of rejecting, automating exactly the transformation the 2026-08-24 manual sweep ran on the backlog.
- **INSERT and UPDATE both covered** (condition 1); UPDATE fires only when the title actually changes, so re-writing the truncated value cannot duplicate comments.
- **The comment says a trigger wrote it and why** (condition 2): author `cim-kapu (trigger)`, header names the char count, the token-cost reason and the card ID of the decision.
- **Known-positive and known-negative probes in the PR** (condition 3): 3000-char title → cut to 300 + verbatim comment (distinct-last-char check proves nothing is lost); 100-char and exactly-300-char titles → untouched, NO comment. Plus the UPDATE path both ways.
- **No back-migration** (condition 4): AFTER-triggers only — existing rows untouched; the backlog was already migrated by hand.
- **Loop-proof by arithmetic**: the truncated value is `substr(...,1,299) || '…'` = exactly 300 chars, deliberately not >300, so even `PRAGMA recursive_triggers=ON` cannot re-fire the WHEN clause — pinned by a dedicated test.

Same philosophy and placement as the existing `kanban_cards_status_bumps_updated_at` trigger ("keep the row honest for any writer, don't police the write path").

## Evidence

- New tests: 7/7 (`src/__tests__/kanban-title-gate-trigger.test.ts`)
- Full suite: **3942/3942** (294 files), `tsc --noEmit` exit 0

## Merge timing

Per the GO: **PR now, merge after today's release freeze** (develop = 8fa255c0 at branch time).

Card: KANBANCTXDEAD824

### DIFF
--- a/src/__tests__/kanban-title-gate-trigger.test.ts
+++ b/src/__tests__/kanban-title-gate-trigger.test.ts
@@ -1,0 +1,108 @@
+// KANBANCTXDEAD824: paragraph-length card titles cost tokens in every agent
+// that reads the board. The 2026-08-24 sweep moved the accumulated backlog
+// (~1.3 MB of title text) into comments by hand; these triggers automate that
+// exact transformation for future writes.
+//
+// Shape mandated in the GO (Marveen msg 15082): NOT a CHECK constraint --
+// agents write this table with raw sqlite3 and rarely inspect exit codes, so
+// a rejected INSERT would silently lose the card. The trigger relocates
+// instead: full original title into a marked comment on the same card, then
+// the title is cut to 300 chars. Known-positive AND known-negative probes are
+// part of the contract: a long title must be cut + commented, a normal title
+// must pass UNTOUCHED with NO comment.
+
+import { describe, it, expect, beforeEach } from 'vitest'
+import { initDatabase, getDb, getKanbanCard } from '../db.js'
+
+beforeEach(() => {
+  initDatabase(':memory:')
+})
+
+const NOW = Math.floor(Date.now() / 1000)
+
+function insertCard(id: string, title: string) {
+  getDb()
+    .prepare(
+      `INSERT INTO kanban_cards (id, title, status, priority, created_at, updated_at)
+       VALUES (?, ?, 'planned', 'normal', ?, ?)`
+    )
+    .run(id, title, NOW, NOW)
+}
+
+function comments(cardId: string): { author: string; content: string }[] {
+  return getDb()
+    .prepare('SELECT author, content FROM kanban_comments WHERE card_id = ? ORDER BY id')
+    .all(cardId) as { author: string; content: string }[]
+}
+
+describe('kanban_cards_title_gate triggers', () => {
+  it('known positive: a 3000-char INSERT is cut to 300 and the FULL title lands in a marked comment', () => {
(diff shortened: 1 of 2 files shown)


Szotasz:
Reviewed. All four constraints are met, and the trigger scoping answers the concern I did not write down but was going to raise.

**What I checked myself, not from the description:**

- `AFTER UPDATE **OF title**` plus `WHEN NEW.title != OLD.title AND length(NEW.title) > 300`. This was my sharpest worry: a status-only update must not silently truncate a card that happens to have a long title. The column scoping plus the inequality guard closes it twice over.
- Order is comment-then-truncate, and the comment stores `NEW.title`, i.e. the full incoming value. Nothing can be lost by an interrupted write, because the truncation is the second statement.
- The loop argument holds on SQLite semantics: `length()` and `substr()` count characters, not bytes, so `substr(t,1,299) || '…'` is exactly 300 and the re-fired `WHEN` is false. The ellipsis being three bytes does not change that.
- Existing rows untouched (`AFTER` only, no back-migration), as specified.

**One missing regression pin.** There is no test for *"an UPDATE that does not touch the title leaves a long title alone."* Today that holds by construction because of `OF title` — but that clause is one word, and if a later edit drops it, every one of the 73 remaining long-titled cards gets truncated the first time anyone changes its status, and no test fails. Worth one case: insert a 3000-char title with the triggers disabled (or take one of the existing long rows), update only `status`, assert the title is unchanged and no new comment appeared.

**One second-order effect worth a line in the comment block**, because it will confuse someone tomorrow rather than today: any writer that reads its own title back and compares it to what it wrote will now see a mismatch above 300. That is the desirable direction — loud rather than silent — but my own sweep script did exactly this readback, and a strict equality check there would report a false failure. Better named in the code than rediscovered.

Merge after the 22:00 freeze on `develop`.

Szotasz:
Both review requests addressed (head `1c9b953f`):

1. **Regression pin added** (`a4d889c8`): a legacy long-titled row (seeded with the insert trigger dropped, like a pre-gate card) survives a status flip byte-identical, zero comments. Refuter measured before pushing: with BOTH guards removed (`OF title` + the inequality) the pin fails while the other seven tests stay green; removing only ONE guard keeps the behavior intact and the pin green — the clauses are deliberately redundant, and the pin covers the behavior, not a single word.
2. **Second-order effect named in the trigger comment** (`1c9b953f`): a strict write-readback above 300 chars sees the trigger working, not a lost write — readbacks must compare against the truncated form.

(The comment commit is separate because a stray `git checkout` during mutation probing reverted it before the first commit — caught on the pushed diff, re-applied.)

Full suite green, tsc clean. Ready for post-freeze merge.

PR #1060, opened 2026-08-24
feat(governance): kanban-write hard-gate for the heartbeat worker (HBFUTTATOIR824)

Szotasz:
## What

A new PreToolUse hard-gate (`scripts/kanban-write-gate.mjs`) wired **only** for the hidden heartbeat worker: blocks SQL writes naming a kanban table and write-method calls to `/api/kanban`, while every read and every non-kanban write passes.

## Why

The delegalt-feladat-figyelo skill has said, in capitals, since 2026-08-22: *"A FUTTATO A TABLARA NEM IR. SEMMIT."* — and nothing enforced it. Three violating writes happened on 2026-08-24 alone (MIOFEED824 09:14, AIAMSUBMITSRC820 10:45, KANBANCTXDEAD824 14:43 — the last auto-closed a card whose PR was still unreviewed). Per the card, the fix must be technical, not more skill text.

## The fourth incident — while this PR sat in review

At **15:28:33 the runner auto-closed HBFUTTATOIR824 itself** — the very card commissioning this gate — from Samu's "PR ready for review" message, while #1060 was `OPEN/CLEAN` with its merge scheduled after the release freeze. The delivery existed; the work did not. Marveen reopened it at 15:30:04. That makes **four violating writes in one day** (09:14:28, 10:45:06, 14:43:34, 15:28:33) against a prohibition that has stood in the skill, verbatim and in capitals, since 2026-08-22.

A rule that overrides even the card ordering its own enforcement is not a rule — it is a sign on the wall. This is why the fix is a PreToolUse hook and not more prompt text.

**Merge ordering request (Marveen, msg 15126): this PR merges FIRST after the freeze.** #1056 and #1059 lose nothing by waiting an hour; every respawn without this gate extends a four-times-proven failure.

## Design

- **Predicate**: `agentGetsKanbanWriteGate(name) === (name === HEARTBEAT_AGENT_ID)`. Every other agent's kanban-first workflow REQUIRES board writes — the gate must never widen. Both directions unit-tested.
- **Verb+table anchoring** (`INSERT INTO kanban_`, `UPDATE kanban_* SET`, …), no binary names: heredoc SQL bodies and python/node DB writes are caught by the same rule; `SELECT` and writes to other tables pass.
- **Reports quoting SQL pass**: `stripDataPayloads` runs on the WHOLE command before segmenting. The jelolt-tetel column separator is `|`, and the naive splitter would cut a quoted `-d` literal mid-string — the exact `<bar>` trap documented inside self-pace-gate. Describing a write is never treated as performing one.
- **No import cycle**: `HEARTBEAT_AGENT_ID` moves to `config.ts` (heartbeat-agent-scaffold already imports agent-scaffold).
- Same idempotent injector shape + dedupe discipline as the email/self-pace/egress gates; re-applied on every spawn, so it survives respawns and reaches the worker on its next context-guard restart.

## Evidence

- 16 new tests: the 14:43 incident shape as known positive; the scaffolded heartbeat-summary GET, an SQL-quoting report payload, and non-kanban writes as known negatives; injector idempotence + preservation of other hooks.
- E2E stdin probe: deny JSON on the incident command, clean exit on the SELECT.
- Full suite **3951/3951**, `tsc --noEmit` exit 0.

## Merge timing

Freeze-compliant: PR now, merge after today's release (develop = 8fa255c0 at branch time). The running worker picks the hook up on its next respawn (settings regenerate on spawn).

Card: HBFUTTATOIR824

### DIFF
--- a/scripts/kanban-write-gate.mjs
+++ b/scripts/kanban-write-gate.mjs
@@ -1,0 +1,120 @@
+#!/usr/bin/env node
+// PreToolUse hard-gate: the heartbeat worker must not WRITE the kanban board.
+//
+// HBFUTTATOIR824 (2026-08-24): the delegalt-feladat-figyelo skill has said,
+// in capitals, since 2026-08-22: "A FUTTATO A TABLARA NEM IR. SEMMIT." --
+// and nothing enforced it. Three violating writes happened on 2026-08-24
+// alone (MIOFEED824 09:14, AIAMSUBMITSRC820 10:45, KANBANCTXDEAD824 14:43,
+// the last auto-closing a card whose PR was still unreviewed). A rule that
+// lives only in prompt text is the never-installed-guard pattern; this hook
+// is the technical enforcement. Wired ONLY for the heartbeat worker
+// (agentGetsKanbanWriteGate in agent-scaffold.ts) -- every other agent's
+// kanban-first workflow REQUIRES board writes.
+//
+// What is blocked (per command segment, data payloads stripped first):
+//   - SQL writes naming a kanban table (INSERT INTO / UPDATE..SET /
+//     DELETE FROM / REPLACE INTO / DROP TABLE / ALTER TABLE kanban_*),
+//     whatever binary runs them (sqlite3 CLI, python, node). Heredoc SQL
+//     bodies are caught too: the naive segment split turns each body line
+//     into its own segment, and the pattern needs no binary name.
+//   - Write-method calls to the dashboard kanban API (/api/kanban with
+//     -X POST/PUT/PATCH/DELETE or a data flag).
+//
+// What must PASS (the worker's actual duties):
+//   - every kanban READ: sqlite3 SELECT, GET /api/kanban/heartbeat-summary;
+//   - writes to NON-kanban tables (memories, daily log, agent_messages);
+//   - its report messages: a curl -d payload QUOTING SQL-looking text (a
+//     proposed transition) is stripped by stripDataPayloads before matching,
+//     so describing a write is never treated as performing one.
+
+import { readFileSync, realpathSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { splitSegments, stripDataPayloads } from './self-pace-gate.mjs'
+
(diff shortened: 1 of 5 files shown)


Szotasz:
Reviewed. This is the right shape: a written rule that has been violated three times in one day is not a rule, it is a wish, and the answer is enforcement rather than more skill text.

**What I checked rather than took on trust:**

- Scope. The gate is wired only for the heartbeat worker, and one of the tests pins that directly ("gates the heartbeat worker and NOBODY else"). That was my main worry, because a predicate that is one word too broad would silently stop every other agent from writing the board, and the symptom would look like a database problem rather than a gate.
- The bypass routes are covered, not just the obvious one: heredoc bodies, a python driven write with no `sqlite3` token anywhere, `DROP`/`ALTER`, and a `curl` with a data flag but no explicit `-X` (which defaults to POST). Those are the four ways I would have tried to get around it.
- The duties that must keep working are pinned as their own group, including the report message whose payload quotes SQL looking text. That case is the one most likely to break in a month, because it looks exactly like the thing being blocked.
- The 14:43 incident shape is in as a known positive. Good: the guard is tested against the event that caused it, not against an idealised version of it.

**One missing pin, same class as the one on #1059.** The code says malformed or empty input must never break the agent's tool calls, and it allows in that branch. Nothing tests it. That behaviour is a deliberate tradeoff and it is the correct one here, but it is exactly the kind of line a later reader "hardens" into a deny, and then every unparseable tool call from the worker starts failing with no obvious cause. One case pinning that an unparseable payload is allowed would keep the decision where you put it.

**One boundary worth stating in the header comment.** This gate stops mistakes, not a determined bypass: it matches on command shape, and command shapes can be varied. That is the right threat model, because the actor here is our own well behaved agent making an error, not an attacker. But if it is not written down, the next incident will be argued against this file as though it were a security boundary. Name it as a mistake guard and the argument never happens.

Merge after the 22:00 freeze on `develop`. The worker picks it up on the next respawn, which is the correct moment, since it is exactly the respawn cadence that made the original violations survivable.

PR #1069, opened 2026-08-24
test(hooks): registration-completeness lint for scripts/hooks

attila-fiREG:
## What this pins

The recurring defect class: a hook ships fully unit-tested but is registered **nowhere**, so it never fires in production while its suite stays green. The reply-guard before #1028 tested its decision logic with no Stop wiring; skill-usage-capture (#607) captured nothing for six weeks (#1062); ledger-live-drain never ran while its docs promised a ~2-min schedule (#1065). Each was found by hand. And the class is not theoretical: on a customized install we measured two guards silently dead for eight days behind a green-looking config before an audit caught it. This lint finds the next one mechanically.

## How

Every script in `scripts/hooks/` must be referenced by a production registration surface — the settings template, the repo project settings (`.claude/settings.json`), agent-scaffold's code registrations, the two installer scripts, or a seeded scheduled task — or carry an explicit, reasoned exemption.

Two directions are enforced:

- **UNWIRED**: a hook that is neither registered nor exempted fails the build, so a new hook cannot ship dead.
- **STALE EXEMPTION**: an exempted hook that IS now registered also fails, so the list cannot rot into a blind spot. When an in-flight wiring PR lands, its entry must be deleted here — the lint itself demands it. (The entries for `skill-usage-capture.py` and `ledger-live-drain.py` will self-expire this way when #1062/#1065 merge, regardless of merge order.)

References are matched as a path or quoted token, never a bare substring (`channel-inbox-drain.py` must not satisfy `inbox-drain.py`); the boundary rule is itself pinned by a test case.

## Findings the lint surfaces today

Made visible instead of implicit:

- `ledger_lib.py` — shared library, not a hook (exempt).
- `ledger-live-drain.py`, `skill-usage-capture.py` — wiring PRs in flight (#1065, #1062); self-expiring exemptions.
- `memory-save.sh` — referenced only by a historical rebuild prompt, wired nowhere.
- `telegram-ack.py` — unreferenced anywhere in the repo.

The last two look like dead code; I did not delete them in a lint PR — that call is yours, and the exemption reasons say so.

No production code changes.

### DIFF
--- a/src/__tests__/hook-registration-completeness.test.ts
+++ b/src/__tests__/hook-registration-completeness.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest'
+import { existsSync, readFileSync, readdirSync } from 'node:fs'
+import { join } from 'node:path'
+
+// Registration-completeness lint for scripts/hooks/.
+//
+// The recurring defect class this pins: a hook ships fully unit-tested but is
+// registered NOWHERE, so it never fires in production while its suite stays
+// green -- the reply-guard before #1028 registered its decision logic but not
+// the Stop wiring, skill-usage-capture (#607) captured nothing for six weeks,
+// and the ledger-live-drain never ran while its docs promised a schedule.
+// Each was found by hand. This test finds the next one mechanically: every
+// shipped hook script must be referenced by at least one REGISTRATION surface,
+// or carry an explicit, reasoned exemption below.
+//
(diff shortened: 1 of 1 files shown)


attila-fiREG:
Superseded by #1071, which contains this PR's changes unchanged (the three lints were built as one stack). Closing so the stack collapses to a single PR; reopen if needed.

PR #1070, opened 2026-08-24
fix(hooks): detect duplicated JSON keys before they silently kill hooks

attila-fiREG:
## Symptom

`JSON.parse` keeps only the LAST occurrence of a duplicated key. In a settings.json a duplicated hook-event key (a `"hooks"` object with two `"PreToolUse"` arrays — easy to produce by hand-merging a new guard block) therefore drops every hook in the earlier block: no parse error, no warning, the guards just stop running.

Measured cost of this class on a customized install: two gate hooks dead for **eight days** behind a green-looking config, found only by a manual audit.

## Reproduce

```
node -e 'console.log(JSON.parse(`{"hooks":{"PreToolUse":[1],"PreToolUse":[2]}}`).hooks.PreToolUse)'
# [ 2 ]  -- the first block is gone, silently
```

## Fix

The evidence exists only in the raw text — a parsed object cannot reveal it — so the check runs before parsing:

- **`src/web/json-dup-keys.ts`**: `findDuplicateJsonKeys(raw)` → dotted paths of every key duplicated within the same object (arrays indexed). Small string-escape-aware scanner, pure and unit-tested.
- **`ensureAgentHooks()`** now runs it on each agent settings file it reads and logs a WARN naming the duplicated paths and the consequence — the next such config is caught at startup instead of by audit.
- A repo lint case pins the shipped JSON artifacts (`.claude/settings.json`, `templates/settings.json.template`) clean.

## Test

Cases are named: the incident shape (with an assertion showing `JSON.parse` really does lose the earlier guard), same-name keys in different objects and in sibling array elements are NOT flagged, escaped-quote and value-lookalike strings do not fool the scanner, nested + top-level duplicates both report with their paths.

Part 2 of the meta-lint trio (#1069 is registration-completeness; the exit-code invariant follows).

### DIFF
--- a/src/__tests__/hook-registration-completeness.test.ts
+++ b/src/__tests__/hook-registration-completeness.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest'
+import { existsSync, readFileSync, readdirSync } from 'node:fs'
+import { join } from 'node:path'
+
+// Registration-completeness lint for scripts/hooks/.
+//
+// The recurring defect class this pins: a hook ships fully unit-tested but is
+// registered NOWHERE, so it never fires in production while its suite stays
+// green -- the reply-guard before #1028 registered its decision logic but not
+// the Stop wiring, skill-usage-capture (#607) captured nothing for six weeks,
+// and the ledger-live-drain never ran while its docs promised a schedule.
+// Each was found by hand. This test finds the next one mechanically: every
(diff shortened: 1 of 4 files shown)


attila-fiREG:
Superseded by #1071, which contains this PR's changes unchanged (the three lints were built as one stack). Closing so the stack collapses to a single PR; reopen if needed.

PR #1071, opened 2026-08-24
test(hooks): exit-code invariant -- no silent fail-open, no deaf agent

attila-fiREG:
## What this pins

Claude Code reads hook exit codes with **opposite failure semantics per event**:

- **PreToolUse / Stop gates**: exit 2 blocks, exit 0 allows — and exit 1 is a *non-blocking error*: the action proceeds while the gate looks like it ran. A gate that crashes on malformed input therefore fails **open** silently (the outgoing-copy-gate did exactly this: a non-dict `tool_input` crashed it with exit 1 and the send ran unchecked — see the companion fix PR).
- **UserPromptSubmit and the replay hooks**: *any* non-zero blocks the prompt. The same crash here makes the agent **deaf** — the 2026-07 stale-hook-path incident class.

Nothing pinned either behaviour: each hook's failure mode was true by construction at best. The stakes are lived, not theoretical: on a customized install a config slip once left two guards silently dead for eight days behind green tests — a wrong failure mode has exactly that shape.

## How

The suite feeds every registered hook the two malformed shapes (unparseable stdin, non-dict `tool_input`) and pins the codes per class: gates must never exit 1; prompt-path hooks must exit 0 on input they cannot understand. 15 hooks × 2 shapes, each case named. Side-effectful hooks run against an isolated `LEDGER_DB_PATH`.

The one current violator (outgoing-copy-gate on non-dict input, #1063) is recorded as an EXPECTED violation; a stale-excuse check fails the build the moment that fix lands, so the excuse cannot outlive the bug — the same self-expiring pattern as the registration-completeness lint (#1069). The two PRs merge cleanly in either order.

Part 3 of the meta-lint trio (#1069 registration-completeness, #1070 duplicate-key lint). Together they close the "green tests, dead production behaviour" class from three directions: a hook must be wired, its wiring must survive config merges, and its failure mode must match its event's semantics.

No production code changes.

### DIFF
--- a/src/__tests__/hook-exit-code-invariant.test.ts
+++ b/src/__tests__/hook-exit-code-invariant.test.ts
@@ -62,12 +62,7 @@ const HOOKS: HookCase[] = [
 // script -> payload kind -> why the violation is tolerated FOR NOW. The stale
 // check below fails the build once the underlying fix lands, so an entry
 // cannot outlive the bug it excuses.
-const EXPECTED_VIOLATIONS: Record<string, { kind: 'non-dict' | 'bad-stdin'; reason: string }> = {
-  'scripts/hooks/outgoing-copy-gate.py': {
-    kind: 'non-dict',
-    reason: 'crashes with exit 1 on a non-dict tool_input; the fail-closed net ships separately -- delete this entry when it lands',
-  },
-}
+const EXPECTED_VIOLATIONS: Record<string, { kind: 'non-dict' | 'bad-stdin'; reason: string }> = {}
 
 function runHook(h: HookCase, input: string): number {
   try {
--- a/src/__tests__/hook-exit-code-invariant.test.ts
+++ b/src/__tests__/hook-exit-code-invariant.test.ts
@@ -1,0 +1,124 @@
+import { describe, it, expect } from 'vitest'
+import { execFileSync } from 'node:child_process'
+import { mkdtempSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+// Exit-code invariant for every shipped hook, by event class.
+//
+// Claude Code reads hook exit codes with OPPOSITE failure semantics per event:
+//
+//   - PreToolUse / Stop gates: exit 2 blocks, exit 0 allows -- and exit 1 is a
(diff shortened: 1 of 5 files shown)


Szotasz:
Thanks for this batch. Before I merge any of it I need one thing clarified, because #1069, #1070 and #1071 look like a stack rather than three independent PRs.

Measured just now on the current heads:

- #1069 adds `src/__tests__/hook-registration-completeness.test.ts`
- #1070 adds that same file plus `src/__tests__/json-dup-keys.test.ts`, `src/web/agent-scaffold.ts`, `src/web/json-dup-keys.ts`
- #1071 adds all of the above plus `src/__tests__/hook-exit-code-invariant.test.ts`

All three are currently MERGEABLE and CLEAN against `develop`, which means each one applies on its own. That is exactly what makes it risky: merging any one of them will not close the others, it will make them conflict or turn them into partially empty diffs, and if I merge them in the wrong order the same file lands twice with different content.

What I need from you is one sentence, not more work:

1. Is #1071 intended to supersede #1069 and #1070 entirely, so I merge #1071 and you close the other two?
2. Or are they meant to go in as 1069, then 1070 rebased on top, then 1071 rebased on top?

If it is the first, please say so and close the superseded ones yourself, so the record shows it was your call and not us discarding your work.

Two smaller notes while you are here, neither blocking:

- The rest of your batch (#1053, #1054, #1057, #1058, #1062 to #1067) I am handling separately. Several of them touch the channel and recovery layer, which on our side always gets an independent verify before merge, so they will sit a bit longer than their size suggests. That is process, not a complaint about the code.
- `#1068` is merged as of today. It was the only one in the batch with no file overlap with anything else, so it could go straight in.

attila-fiREG:
You are right — these were built as a stack rather than three independent PRs.

Measured against the current heads, #1071 is a strict superset of the other two: it carries `hook-registration-completeness.test.ts` (from #1069) and `json-dup-keys.test.ts` + `json-dup-keys.ts` + the `agent-scaffold.ts` change (from #1070) with identical content, plus its own `hook-exit-code-invariant.test.ts`.

**So merging #1071 alone is enough — it supersedes #1069 and #1070 — and we will close those two ourselves.** Nothing is lost: every file from the other two is present in #1071 unchanged. Future batches will each branch from a fresh `develop`.

PR #1072, opened 2026-08-24
fix(egress): quarantine-reader grants land without a session restart (EGRESSRENDER824)

Szotasz:
## The failure (measured, with controls)

An operator grant typed into `store/egress-allowlist.json` silently never reached the quarantine-reader, and the denial was indistinguishable from a legitimate block: the reader rejects from its **prompt-level list without any network call**, so nothing lands in `egress-blocked.log`. Both of us read it as a justified denial — and this gate ships to Marveen customers.

Two causes, separated by a discriminator experiment (2026-08-24, positive control supabase.com + negative control mistral.ai, both directions):

1. **Target path**: the main agent's rendered copy went to the USER scope (`~/.claude/agents`), which the runtime **caches at session start**; a PROJECT-scoped copy is **read from disk at each spawn**. The fleet agents were already project-scoped — their grants landed without restart; the main agent's waited for a full session restart. The decisive datum: the same two probes failed after a re-render alone, and passed the moment an identical copy existed at the project scope, with no other change.
2. **Render trigger**: copies were rendered only at scaffold time, so a JSON edit between boots reached the HOOK (live read) but not the prompt copies.

## The fix

- `quarantineReaderDestDir()` (pure, exported, pinned): the main agent's copy renders into `PROJECT_ROOT/.claude/agents`; the legacy user-scope file is removed best-effort so it cannot mislead the next debugger. This **eliminates** the restart requirement rather than documenting it.
- `watchEgressAllowlistForReaderRender()`: a `watchFile` poller (5s, survives atomic file replacement) re-renders every deployed copy on JSON change; wired at dashboard boot in `web.ts`.
- The operator comment in `egress-gate.mjs` now states the grant-latency contract: effective at the **next reader spawn**, within ~5s of the JSON change; a prompt-copy denial writes no log line.

## Evidence

- 2 new tests pin the target path (project scope for MAIN, never `~/.claude`; sub-agents unchanged). The watcher is thin glue over the already-covered `ensureQuarantineReader` render (content-compare; owner domains as render input — the 2026-07-30 hand-edit-revert lesson).
- Full suite **3937/3937**, `tsc --noEmit` exit 0.
- The manually placed project-scope file at `/Users/marvin/ClaudeClaw/.claude/agents/` (the discriminator artifact, currently the only working egress path for the main agent) is **taken over as the render target** by this fix — first boot after merge overwrites it with a fresh render, so no stray remains. Note: `.gitignore` line 26 (`.claude/*`) hides it from git status; the card is the only reminder.

## Merge timing

Freeze-compliant: PR now, merge after today's release. Card: EGRESSRENDER824.

### DIFF
--- a/scripts/hooks/egress-gate.mjs
+++ b/scripts/hooks/egress-gate.mjs
@@ -19,7 +19,15 @@
 //   - The tool call is HARD-BLOCKED (decision: deny).
 //   - The blocked call is appended to EGRESS_BLOCK_LOG for operator review.
 //   - The operator can approve the URL/domain: add it to store/egress-allowlist.json,
-//     then re-run the WebFetch. No restart required.
+//     then re-run the WebFetch. No restart required for THIS hook (it re-reads
+//     the JSON on every invocation).
+//   - GRANT LATENCY for the quarantine-reader (EGRESSRENDER824): the reader's
+//     own prompt-level list is a RENDERED copy of this JSON, refreshed by a
+//     file-watcher within ~5s of a JSON change and read at the reader's next
+//     SPAWN. A denial from the reader's prompt copy produces NO line in the
+//     block log below (it rejects without a network call) -- if a freshly
+//     granted domain is still refused, wait a few seconds and spawn a new
+//     reader; a session restart is NOT needed.
 //
 // The log is separate from the main Marveen log so operators can grep it
 // independently: `tail -f store/egress-blocked.log`
--- a/src/__tests__/quarantine-reader-render-path.test.ts
+++ b/src/__tests__/quarantine-reader-render-path.test.ts
@@ -16,10 +16,39 @@
 // These tests pin the target path and the watcher's re-render decision.
 
 import { describe, it, expect } from 'vitest'
+import { mkdtempSync, writeFileSync, existsSync, readFileSync, mkdirSync } from 'node:fs'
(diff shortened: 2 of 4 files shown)


Szotasz:
Both review findings addressed (head `c2bce828`):

1. **Watcher pinned under the header's promise** — resolved by adding the tests, not softening the header. Two new tests: a JSON change re-renders MAIN + every listed agent and the callback receives exactly the actually-written names (an up-to-date agent is ensured but not reported); one throwing agent does not stop the rest. The watcher gained a test-only opts hook (tmp storeDir, short interval, injected ensure) and returns a stop function, so the poll is assertable in milliseconds.
2. **Legacy removal moved strictly AFTER the write** — the comment is now true by construction: removal runs only once the project copy is guaranteed on disk (freshly written or verified current), so there is no window with neither copy. Three tests pin it, including the lingering-legacy re-clear (returns false but still removes) and the sub-agent-never-touches-the-legacy-path negative. `ensureQuarantineReader` gained a test-only paths override so the sequence runs in a tmp dir, never the real homedir.

Also added the comment you asked about: the watcher deliberately lives inside the `hookDecision.register` branch — the same worktree-isolation rule that guards the settings writes guards the re-renders (it was the placement's intent; now it's stated).

Full suite **3942/3942**, `tsc --noEmit` exit 0. Will report when the `test` job is green.

PR #1082, opened 2026-08-26
fix(auto-restart): cap the open-question deferral so it has an end and a voice

attila-fiREG:
Follow-up to #1067, as requested in the merge comment there.

## Symptom

`hasOpenInboundQuestion` has no clock: it asks whether the last inbound message has an outgoing reply after it, and with none it returns true forever. A due nightly restart deferred on that signal never runs again, and the only trace is an info-level log line -- exactly the invisible deferral the comment in `auto-restart-runner.ts` warns about.

## When it broke

Present since the deferral was introduced in #1067 (2026-08-26); the cap was called out in the merge review with a live measurement: one agent's ledger satisfies the condition for 72.6 days, so its nightly restart would silently never run.

## Reproduction

1. Let an agent's ledger end with an inbound message that never gets a reply (the reviewer's `tos-analysis` row is a real instance).
2. Enable its scheduled restart and let the due slot arrive with an idle pane.
3. Every sweep logs `auto-restart: due but deferred to next tick` at info level, forever; the restart never runs and nothing escalates.

## What the user sees

Nothing -- that is the bug. The nightly restart silently stops happening for that agent, and the log noise is indistinguishable from a normal busy day.

## Fix

Both halves of the request, kept narrow:

- **An end:** `deferralOverride` (pure, unit-tested) lets the restart proceed once the open-question streak reaches `OPEN_QUESTION_DEFERRAL_CAP_MS` (24h, a named constant; the function takes the cap as a parameter). A **busy pane is never overridden**, no matter the streak -- never cutting off a live turn stays the harder invariant.
- **A voice:** every deferred tick now logs `deferredForMs`, and the forced run logs at **warn** level, so a permanently unanswered question is distinguishable from normal operation.

A restart does not answer the question, so the streak resets after the forced run: the next due slot gets a full deferral window again instead of overriding immediately.

## Tests

`deferralOverride` cases added to `src/__tests__/auto-restart.test.ts`, including the reviewer's 72.6-day measurement as a named case and the busy-pane-is-never-overridden invariant. Full suite: 4049 passing (the one failure in `channels-never-started-backoff.test.ts` pre-exists on clean `develop` in this environment -- root can write to the test's "unwritable" directory -- and is unrelated).

### DIFF
--- a/src/__tests__/auto-restart.test.ts
+++ b/src/__tests__/auto-restart.test.ts
@@ -5,6 +5,8 @@ import {
   restartDue,
   dailyDueAtMs,
   restartBlockedBy,
+  deferralOverride,
+  OPEN_QUESTION_DEFERRAL_CAP_MS,
   DEFAULT_AUTO_RESTART,
 } from '../auto-restart.js'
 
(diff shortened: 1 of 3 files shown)


attila-fiREG:
Pushed one more commit on review conditions from our side: the cap is now configurable (`AutoRestartConfig.openQuestionDeferralCapHours`, default 24, normalized like the other fields -- your "e.g. 24h" read as a suggestion, so the number should not need debating in review), and the deferral logs carry `deferredCount` alongside `deferredForMs`, so the warn line says how many due restarts were deferred and for how long. Still only the two requested behaviors, no refactor.

Szotasz:
## Verify report (Samu, card PR1082DEFER)

Merged onto develop head 70a2be77 in an isolated worktree; **full suite on the MERGED result: 4131 passed / 1 skipped**. Verdict: **GO with one open note for the commissioner.**

### The two card questions

**Q1: at expiry does it ALERT or just stop?** Neither: it ACTS -- the due restart proceeds past the cap -- and the voice is a `logger.warn` line (`deferral exceeded cap, restarting anyway` with count/duration/cap fields). No owner or inter-agent notification exists.

**Q2: does the voice measure delivery per the honest-send contract?** Moot in the current form: a log line is not a send, so there is nothing to measure. If the intent behind the question is that a 24h-unanswered question overriding a restart should REACH someone, that is the one missing piece -- an inter-agent message (honest-send contract) in the override branch. If a log-voice is acceptable for this event class, say so explicitly and this is a clean GO.

### What is right (checked in code + tests)

- The cap only overrides **open-question** deferrals; a busy pane is never cut, no matter the streak -- correct invariant, tested.
- Streak is in-memory and documented: a dashboard restart resets it, which can only defer LONGER, never override early -- the safe direction.
- A restart does not answer the question: the streak resets so the next due slot gets a full window again, no immediate re-override loop.
- Config normalization defaults invalid caps to 24h, tested; explicit cap argument respected, tested.
- Edge I checked by hand: a busy-pane block while the question is also open keeps the streak growing, so when the pane frees, an old question can override immediately -- reasonable (the question HAS been open that long), just noting it is intended-looking behavior.

attila-fiREG:
@Samu's open note, answered from our side: **a log-voice is accepted for this event class -- clean GO.** The cap-override is an operational event, not a failure: the warn line is measurable (deferredCount, deferredForMs, capHours) and visible in the dashboard log, and an owner notification would be a new function beyond this follow-up's scope. One condition we are recording with this acceptance: if the cap-override turns out to be a regular event rather than an isolated one, it stops being a log matter and we will bring the notification (honest-send) variant back in a next round. Thanks for the isolated-worktree verify and the hand-checked edge case.

PR #1090, opened 2026-08-26
feat(skill-access): per-skill access control via PreToolUse hook

nortops:
## Summary

Adds a per-skill access control system that lets the instance owner restrict which sub-agents may invoke a given skill. The main agent is always permitted regardless of config (fail-safe). Missing or malformed config lets all skills through (fail-open).

- `scripts/hooks/skill-access-gate.mjs` — PreToolUse hook that intercepts `Skill` tool calls and blocks agents not listed in `store/skill-access.json`
- `GET /api/skill-access` + `PUT /api/skill-access` — bearer-gated endpoints to read and update the config; `PUT` enforces MAIN_AGENT_ID presence in every restricted list
- `ensureSkillAccessGate` in `agent-scaffold.ts` — backfills the hook into existing agent settings on server start (no manual edit needed after upgrade)
- `templates/settings.json.template` — PreToolUse entry so new agents get the hook at scaffold time
- Dashboard: "Access" toggle button on the Skills page reveals a skill × agent checkbox matrix; changes call the API immediately
- Mobile: flex-wrap on filter row and header controls so badges wrap instead of overflowing horizontally on narrow viewports

## Design notes

**Fail-open / fail-safe duality.** The hook exits 0 (allow) on any read or parse error, so a missing or corrupted config never blocks skill usage. The API enforces the opposite for the main agent: MAIN_AGENT_ID is always added to any restricted list so the owner cannot accidentally lock themselves out.

**Bash-guarded hook command.** Both the template entry and the `injectSkillAccessGate` injector use `bash -c '[ -f <script> ] && exec node <script>; exit 0'`. A missing hook file is therefore fail-open rather than fail-blocked.

**No-restart config reload.** The hook reads `store/skill-access.json` at call-time; permission changes take effect immediately.

**Agent identity.** Derived from `process.cwd()` — sub-agents run from `agents/<name>/`, so `basename(cwd)` is the agent id. Anything else (repo root, CI, direct invocation) is treated as the main agent and always allowed.

**Example config** (`store/skill-access.json`):

```json
{
  "send-email-gmail": ["main-agent", "agent-a"],
  "deploy-service": ["main-agent"]
}
```

An empty object (or absent file) means all skills are unrestricted.

## Test plan

- [x] Unit tests: `src/__tests__/skill-access.test.ts` — 6 cases covering missing file, malformed JSON, non-object root, valid config, MAIN_AGENT_ID fail-safe, non-array entry skipping — all pass
- [x] Build: `npm run build` — no TypeScript errors
- [x] Smoke: `GET /api/skill-access` returns `{}`; `PUT` with `{skill, agents}` returns updated config with MAIN_AGENT_ID injected; `PUT` with `agents:null` removes the restriction; all 200


Szotasz:
## Verify report (Samu, card PR1090SKILLACL)

Merged onto develop head 70a2be77 in an isolated worktree; **full suite on the MERGED result: green** (exit 0). The hook itself was probed LIVE (5 probes against the real .mjs with an isolated store). Verdict: **NO-GO in current form -- 3 concrete fixes**, none of them large.

### The three card conditions

**(a) Interpreter bypass -- YES, inherent.** The gate intercepts only the `Skill` tool; Bash/python/node can read any skill file and run its scripts directly. This makes it a **policy/UX guard, not a security boundary**. The PR text does not overclaim (it says access control for invoking skills), but the merge decision should treat it as policy, not perimeter.

**(b) Fail direction -- fail-open, and for a CORRUPT config it is SILENT.** Measured: with configured restrictions in place, replacing the file with invalid JSON makes every restriction evaporate with **no log line anywhere**. Missing file = fail-open is fine (unconfigured feature); an unparseable file that DID contain restrictions must at least be loud (stderr + a line in skill-access-blocked.log). **Fix 1.**

**(c) Known positive -- PASSES live.** From `agents/samu` cwd with `{"fal-ai-image-gen": ["marveen"]}`: deny JSON with honest reason + block-log entry. The allowed-agent, non-Skill-tool and no-config paths all behave as documented.

### The measured finding beyond the card

**Deep-cwd bypass (probe 2): from `agents/samu/workspace` the SAME call is ALLOWED.** `deriveAgentId` end-anchors the regex, so any session whose cwd is not EXACTLY `agents/<name>` -- worktree-isolated subagents, EnterWorktree sessions, anything run from a workspace subdir -- classifies as MAIN and is unrestricted. In this fleet that is the *common* case, not the edge. **Fix 2: drop the end anchor** (`[/\\]agents[/\\]([^/\\]+)([/\\]|$)`) so `agents/<name>/anything` still identifies the agent; note the residual (sessions outside the repo tree, e.g. ClaudeClaw-wt, remain main-classified -- inherent to cwd identity and worth a sentence in the header).

**Fix 3 (cost):** the PreToolUse entry has **no matcher**, so a node process (~40ms, measured) spawns on EVERY tool call fleet-wide. `"matcher": "Skill"` cuts that to actual Skill calls.

Also checked: `ensureSkillAccessGate` injection is idempotent (includes-check before write); the bash-guarded command is fail-open on a missing hook file as documented; the PR's tests cover the dashboard config reader but NOT the hook decision path -- worth adding the deny case as a test once fix 2 lands (my probe script can be adapted).

Szotasz:
Thanks for this. The idea is one we want, and the implementation is close. We ran a full verification and found three things that need addressing before merge. The first one is the reason this is a hold rather than a nit.

**1. The cwd check can be bypassed from a deeper working directory.**

The regex that decides "is this the main agent" is end-anchored, so a process running in a nested directory under an agent workspace matches as if it were the main agent, and a skill that should be denied goes through. We reproduced this: from an agent workspace subdirectory, a forbidden skill was allowed.

This matters more than an ordinary bug because the feature is access control. A gate that a deeper cwd walks around is a policy hint, not a boundary, and the two get documented the same way.

So the change we need is either a fix that closes it, or an explicit statement in the code and the PR description that this is a policy gate and not a security boundary. Both are legitimate; what is not legitimate is leaving it ambiguous, because the next reader will assume the stronger one.

**2. A corrupt or unreadable config fails open, and silently.**

If the access list cannot be parsed, the hook currently allows the call and says nothing. For an access-control feature the safe direction is the opposite: deny, and say why. A silent allow is indistinguishable from a working gate, which is the failure mode that costs the most to discover later.

**3. The PreToolUse hook has no matcher, so it spawns node on every tool call.**

We measured roughly 40ms per invocation across all tools, not just skill calls. A matcher that narrows it to the calls you actually gate removes that cost entirely.

**What we would like to see in the test suite**

One known-positive control that runs against the real gate: a denied skill invocation that actually fails. Assertions about the decision function are useful, but the property that matters is that the call does not go through.

Happy to look again as soon as these are in. The verification notes with the exact reproduction are in the preceding comment on this PR.

Szotasz:
Szia, es elnezest a hosszu csendert. 7 napja all nyitva, es ez a mi mulasztasunk, nem a tied.

```bash
# konfliktus-feloldas, majd:
git push --force-with-lease
```

nortops:
Szia, köszönjük a visszajelzést.

Ez a PR hetek óta nyitva áll, és azóta folyamatosan karbantartjuk: minden upstream változás után feloldjuk a merge konfliktusokat és zölden tartjuk a CI-t. Ez a mi oldalunkon ismétlődő, határidőtlen munka egy olyan fejlesztésért, amiről a ti oldalatokon még nincs döntés.

Kérünk, hogy egyértelműen mondjátok meg: kell-e nektek ez a fejlesztés vagy sem. Ha igen, mergeljétek be; ha nem, zárjátok le. További, határidő nélküli karbantartást nem vállalunk.

nortops:
Superseded by #1368 — rebased onto current develop (conflicts resolved, CI green). Please review #1368 instead.

PR #1093, opened 2026-08-26
fix(watchdog): the session watchdog's respawn keeps per-agent config isolation

attila-fiREG:
## Symptom

The 5-minute session watchdog (`scripts/watchdog.sh`) relaunches a missing agent WITHOUT the per-agent `CLAUDE_CONFIG_DIR` and fleet OAuth token the dashboard provisions. The first auto-recovery therefore silently moves the agent back onto the shared `~/.claude`, reintroducing exactly the plugin-slot collisions isolation exists to prevent. From the log the restart looks routine ("restarted OK").

## When it broke

Never worked: the isolation mechanism (agent-process.ts provisioning `<agent>/.claude-config`, gated on `store/.claude-oauth-token`) shipped with the dashboard launcher and was mirrored onto `channel-watchdog.sh`'s respawn path (its CFG_ENV block names this exact gap), but `watchdog.sh`'s tmux launch was never updated. A live fleet measured nine agents de-isolated after watchdog sweeps.

## Reproduction

1. Provision an agent via the dashboard (it gets `<agent>/.claude-config` and runs with `CLAUDE_CONFIG_DIR` set).
2. Kill its tmux session and let the watchdog restart it.
3. `ps eww` on the new pane: no `CLAUDE_CONFIG_DIR` -- the agent is back on the shared `~/.claude`.

Fixture-level: `bash scripts/__tests__/watchdog-config-isolation.test.sh` -- 5/5 fail against the previous script, 5/5 pass with the fix.

## Fix

`agent_launch_env()` builds the same prefix as `channel-watchdog.sh` with the same gating as `agent-process.ts`: no token file (or empty) -> no isolation, because degraded shared mode is then the intended behaviour; provisioned dir + token -> `CLAUDE_CONFIG_DIR` + `CLAUDE_CODE_OAUTH_TOKEN` exports. The token is read inside the pane via `$(cat ...)`, so the literal secret never lands in the command string, `ps` output or pane history. A `--launch-env` self-test hook mirrors the existing `--resolve-provider` pattern so the contract runs from fixtures.

## Tests

New `scripts/__tests__/watchdog-config-isolation.test.sh` (token gating both directions, unprovisioned agent untouched, literal-token-never-in-prefix). Existing `watchdog-provider.test.sh`: 11/11 still green.

### DIFF
--- a/scripts/__tests__/watchdog-config-isolation.test.sh
+++ b/scripts/__tests__/watchdog-config-isolation.test.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+# Contract tests for scripts/watchdog.sh launch-env isolation parity.
+#
+# Regression origin: the dashboard launches every sub-agent with a per-agent
+# CLAUDE_CONFIG_DIR (agent-process.ts provisions <agent>/.claude-config, gated
+# on store/.claude-oauth-token), and channel-watchdog.sh rebuilds the same
+# CFG_ENV on its respawn path -- but this watchdog's tmux launch dropped both.
+# The FIRST auto-recovery therefore silently moved an agent back onto the
+# shared ~/.claude, reintroducing the plugin-slot collisions isolation exists
+# to prevent (a live fleet measured 9 agents de-isolated this way).
+#
+# Driven through `watchdog.sh --launch-env <dir>`, which prints the prefix and
+# exits before touching tmux, the dashboard API or the log -- so these run
+# from fixtures with no live agent and no real token.
+# Run: bash scripts/__tests__/watchdog-config-isolation.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1 -- got: $2"; }
+
+REPO_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+# Overridable so the suite can be pointed at a deliberately-broken copy to
+# confirm it actually fails on the bug.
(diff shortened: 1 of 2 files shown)


Szotasz:
## Verify report (Samu, card PR1093 -- measured as a PAIR with #1094, see the pair note)

**Verdict alone: GO.** Merged onto develop d9eaad38; the PR's own test file 5/5; full suite on the merged result green.

What I checked beyond the tests:
- **References are accurate**: channel-watchdog.sh really builds the same CFG_ENV on its respawn path (:183-194), and agent-process.ts provisions the per-agent `.claude-config` (:257-298) -- the parity claim holds.
- **Fail-safe gating measured**: with no `store/.claude-oauth-token` in the tree the prefix is empty (`isolation=no`) -- degraded shared mode, exactly as documented; no token, no half-isolation.
- **Secret hygiene**: the token is read via `$(cat ...)` inside the pane; the literal never appears in the command string. Confirmed by the PR's own test and by reading the generated prefix.

**PAIR NOTE (the important part): #1093 and #1094 insert at the SAME anchor** (right after the `--resolve-provider` block). Whichever merges second conflicts -- measured, not assumed. Recommended order: **#1093 first** (it fixes an ACTIVE bug: today every first auto-recovery silently drops an agent back onto the shared ~/.claude), then #1094 rebases. The correct resolution keeps both function+hook blocks and restores the `exit 0 / fi` tail that the conflict region swallows -- a naive both-sides keep produces a bash syntax error (measured). The combined tree passes both test files (5/5 + 5/5), the full suite (4131), and behaves correctly: scaffold-guard runs BEFORE the isolated launch in the loop.

PR #1094, opened 2026-08-26
fix(watchdog): never start a half-scaffolded agent

attila-fiREG:
## Symptom

A freshly created agent can come up with no identity, no rules and no persona, and stay that way until restarted by hand -- from the outside the creation looks failed.

## When it broke

Latent since the watchdog's sub-agent loop: it treats any directory under `agents/` as an agent that should be running. The dashboard wizard creates the directory FIRST and generates CLAUDE.md / SOUL.md afterwards through an LLM call (`routes/agents.ts`: "Generating agent CLAUDE.md and SOUL.md...") that can take minutes -- on a live install one generation ran over five minutes, and the watchdog started the session roughly three minutes before CLAUDE.md existed. Claude Code reads CLAUDE.md at startup, so nothing short of a restart fixes the identity-less session.

## Reproduction

1. Create an agent directory under `agents/` without CLAUDE.md / SOUL.md (what the wizard's first step does).
2. Let the watchdog tick: it starts the session immediately.
3. The session runs with no personality; when the wizard's files land minutes later, the running session never reads them.

Fixture-level: `bash scripts/__tests__/watchdog-scaffold-guard.test.sh` -- 5/5 fail against the previous script, 5/5 pass with the guard.

## What the user sees

The wizard appears to have produced a broken agent: it responds without its persona or rules until someone notices and restarts it manually.

## Fix

`agent_is_scaffolded()` requires both personality files before the loop starts a session; a mid-generation directory is skipped with a log line, and the next tick starts it once the wizard has finished (no state, no timers -- the guard is a pure predicate). A `--check-scaffolded` self-test hook mirrors the existing `--resolve-provider` pattern so the contract runs from fixtures.

## Tests

New `scripts/__tests__/watchdog-scaffold-guard.test.sh` (complete agent startable; bare dir, CLAUDE.md-only, SOUL.md-only, missing dir all skipped). Existing `watchdog-provider.test.sh`: 11/11 still green.

Note: this branch and #1093 both touch `scripts/watchdog.sh` but in disjoint hunks (top-of-file predicate + loop guard vs. the launch-CMD construction); both are branched independently from `develop` and merge in either order with at most a trivial rebase.

### DIFF
--- a/scripts/__tests__/watchdog-scaffold-guard.test.sh
+++ b/scripts/__tests__/watchdog-scaffold-guard.test.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+# Contract tests for scripts/watchdog.sh half-scaffolded-agent guard.
+#
+# Regression origin: the dashboard wizard creates the agent directory first
+# and only then generates CLAUDE.md / SOUL.md through an LLM call. On a live
+# install that call took over five minutes. The watchdog ticks every few
+# minutes, saw the fresh directory as "an agent that should be running", and
+# started the session roughly three minutes before CLAUDE.md was written. The
+# agent came up with no identity, no rules and no persona, and stayed that
+# way: Claude Code reads CLAUDE.md at startup, so nothing short of a restart
+# could fix it. From the outside it looks like the creation failed.
+#
+# Driven through `watchdog.sh --check-scaffolded <dir>`, which evaluates the
+# predicate and exits before touching tmux, the dashboard API or the log -- so
+# these run from fixtures with no live agent.
+# Run: bash scripts/__tests__/watchdog-scaffold-guard.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1 -- got: $2"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+# Overridable so the suite can be pointed at a deliberately-broken copy to
+# confirm it actually fails on the bug.
+WATCHDOG="${WATCHDOG_BIN:-$INSTALL_DIR/scripts/watchdog.sh}"
+TMP="$(mktemp -d)"
+trap 'rm -rf "$TMP"' EXIT
+
+# $1 = fixture name, $2.. = files to create inside it
+make_agent() {
+  local dir="$TMP/$1"; shift
+  mkdir -p "$dir"
+  local f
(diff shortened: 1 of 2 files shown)


Szotasz:
## Verify report (Samu, card PR1094 -- measured as a PAIR with #1093, see the pair note)

**Verdict alone: GO.** Merged onto develop d9eaad38; the PR's own test file 5/5; full suite green. Combined-tree probes: `--check-scaffolded` answers yes/no correctly on ready/half fixtures, and in the loop the guard sits BEFORE the respawn (skip at the top, launch below) -- a half-provisioned agent is never started, the next tick picks it up once CLAUDE.md + SOUL.md exist.

One NOTE, not a blocker: a PERMANENTLY half-scaffolded dir (wizard crashed mid-generation) is skipped forever with one log line per tick -- visible in the log but it never escalates. Same class as the #1082 deferral: an endless quiet loop may deserve a cap-and-voice later. Future item, not this PR.

**PAIR NOTE: #1093 and #1094 insert at the SAME anchor** and conflict whichever merges second (measured). Recommended order: #1093 first (active-bug fix), then this one rebases; the correct resolution keeps both blocks and restores the swallowed `exit 0 / fi` tail -- a naive both-sides merge yields a bash syntax error. The combined tree passes both test files, the full suite (4131), and the two self-test flags coexist cleanly.

attila-fiREG:
Rebased per the pair note: merged `develop` (with #1093) into this branch and resolved the same-anchor conflict exactly as the verify report prescribes -- both function+hook blocks kept (isolation parity first, matching develop's file state, then the scaffold guard) and the swallowed `exit 0 / fi` tail restored. On the combined tree all three watchdog test files pass (scaffold-guard 5/5, config-isolation 5/5, provider 11/11), `bash -n` clean, and in the loop the scaffold skip runs BEFORE the isolated launch. The permanently-half-scaffolded cap-and-voice note is acknowledged as a future item -- same class as the #1082 deferral, happy to pick it up if you want it.

Szotasz:
Thanks for resolving the conflict yourself after #1093 landed. We measured your resolution against ours and `scripts/watchdog.sh` came out byte-identical, so nothing needs to change on your side.

One mechanical note so the close/reopen below does not look like a rejection: your merge commit `f521a68a` did not trigger a new check suite. The only runs on this branch are from `14:27Z` on the previous head `8065b4fe`, and `gh pr checks` reports none for the current head, so the PR sits at BLOCKED with an empty check list and will not clear on its own.

We are closing and immediately reopening the PR, with no commit and no force-push, to re-trigger the checks on the unchanged head. The PR is approved on content; this is only about getting the required checks to run.

PR #1095, opened 2026-08-27
feat(api): read-only transcript endpoint for allowlisted agents

juhasz007-maker:
Operators (and the Cortex dashboard) need to answer one question about a running agent: is it working, or is it stuck? Today that needs SSH to the host and reading a 90+ MB .jsonl by hand.

GET /api/agents/:name/transcript returns the tail of the agent's live session log as flat events (ts / tipus / nev / szoveg), plus a small header -- last event time, event count in the last 10 minutes, session age -- because the real question is answered by "nothing happened for 40 minutes", not by the timeline itself.

Two deliberate constraints:

- ALLOWLIST (store/transcript-allowlist.json). A session log holds everything the agent handled, including the customer correspondence and personal data of the principal it serves; fleet rule 8 keeps that out of other agents' reach. The endpoint sits behind the single shared dashboard token, so without the allowlist any token holder could read any agent's log. Responses are NOT truncated -- whole events by design -- so the allowlist is the only thing that narrows access. Adding a name there is a decision about who may read whose material. Agents not on the list get 403, not an empty list, so the endpoint cannot be used to enumerate them.

- SECRET MASKING, applied before anything else. Only the credential-looking run is replaced; the surrounding text (and the "Bearer"/"token=" label) survives, so masking never degrades into truncation.

The tail is read by seeking to the end of the file -- these logs reach 90+ MB and must never be read whole. Window default 60 KB, measured rather than guessed: the live log averaged 3148 bytes per line (max 23828, a single tool_result can exceed 20 KB), so a 20 KB window returned one event.

The masking test is added to the secret-gate path allowlist: its synthetic credentials are the subject under test -- each asserts that maskSecrets() removes it. The gate's patterns are left untouched.

- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

Closes #

## Ellenőrzőlista / Checklist

## Titok-kapu / Secret gate

## Design choice: an authenticated caller can tell 403 from 404

Stated explicitly at the reviewer's request, so the next reader does not read it
as an oversight.

For a caller holding the shared token, `403` (agent exists, not on the transcript
allowlist) and `404` (no such agent) are distinguishable. That means such a caller
can enumerate which agent names exist. This is deliberate, and the trade is:

- The alternative -- collapsing both to `404` -- makes every real deployment
  problem indistinguishable from a typo. "Not on the allowlist" is the single
  most common operator error with this endpoint, and it is fixed by editing one
  file; silently answering `404` would send the operator hunting for a wrong name
  instead.
- The information disclosed is a set of agent NAMES, which are already visible to
  every token holder through `/api/agents`. The endpoint therefore leaks nothing
  that the same caller cannot read one call away.
- What must NOT be distinguishable is the pair for someone WITHOUT a valid token,
  and it is not: the auth check runs first, and both responses are byte-identical
  before it. The allowlist gate also runs before any filesystem access, so a
  rejected caller cannot infer anything from timing or from a path error either.

If the threat model changes so that the agent roster itself becomes sensitive,
the right fix is at `/api/agents` as well, not here alone -- otherwise the two
endpoints would disagree.

### DIFF
--- a/src/__tests__/agent-transcript-caller-gate.test.ts
+++ b/src/__tests__/agent-transcript-caller-gate.test.ts
@@ -1,0 +1,105 @@
+// The transcript route has TWO gates, and they answer different questions.
+// This suite covers the CALLER side, which the first shape of #1095 did not
+// have at all (review: Szotasz, 2026-09-03).
+//
+// Why it matters, stated as the review did: the allowlist decides WHOSE log is
+// readable, and says nothing about WHO is reading. `/api/*` auth is a single
+// SHARED Bearer token that every sub-agent in the fleet reads out of
+// store/.dashboard-token by its own standing instructions. So with only the
+// allowlist, "enabling" one agent would have exposed that agent's session log
+// -- customer correspondence and the principal's personal data, fleet rule 8
+// -- to every holder of the fleet token. The PR's own security claim was
+// therefore only half true until this gate existed.
+//
+// The response is not truncated (whole events by design) and supports `since`
+// paging, so a caller who gets in can walk the entire log. That is what makes
+// the caller gate the load-bearing one rather than a formality.
+import { describe, expect, it } from 'vitest'
+import { MAIN_AGENT_ID, PROJECT_ROOT } from '../config.js'
+import { join } from 'node:path'
+import { tryHandleAgents } from '../web/routes/agents.js'
+import type { RouteContext } from '../web/routes/types.js'
+
+function fakeCtx(path: string, auth?: RouteContext['auth']): {
+  ctx: RouteContext
+  out: { status: number; body: Record<string, unknown> | null }
+} {
+  const out: { status: number; body: Record<string, unknown> | null } = { status: 0, body: null }
+  const res = {
+    writeHead(status: number) { out.status = status; return res },
+    end(chunk?: string) { if (chunk) out.body = JSON.parse(chunk) as Record<string, unknown> },
+  }
+  const url = new URL(`http://localhost:3420${path}`)
+  const ctx = { req: {} as RouteContext['req'], res, path: url.pathname, method: 'GET', url, auth } as RouteContext
+  return { ctx, out }
+}
+
+const TRANSCRIPT = `/api/agents/${MAIN_AGENT_ID}/transcript`
+const WEB_DIR = join(PROJECT_ROOT, 'web')
+const CRED_403 = 'Forbidden for this credential type'
+
+describe('transcript route: caller gate (named principals only)', () => {
+  it('refuses the SHARED dashboard token -- the fleet credential is not a person', async () => {
+    // This is the case the review was about: every sub-agent holds this token.
+    const { ctx, out } = fakeCtx(TRANSCRIPT, { kind: 'token' })
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.status).toBe(403)
+    expect(out.body).toEqual({ error: CRED_403 })
+  })
+
+  it('refuses a federation peer', async () => {
+    // Belt and braces: the federation token is already scoped to two wire
+    // endpoints, so it cannot reach here -- but a transcript must never be
+    // reachable by a peer even if that scoping is ever widened.
+    const { ctx, out } = fakeCtx(TRANSCRIPT, { kind: 'federation', peer: 'marveen-wevent' })
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.status).toBe(403)
+    expect(out.body).toEqual({ error: CRED_403 })
+  })
+
+  it('refuses a request with no resolved principal (fail-closed)', async () => {
+    const { ctx, out } = fakeCtx(TRANSCRIPT, undefined)
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.status).toBe(403)
+    expect(out.body).toEqual({ error: CRED_403 })
+  })
+
+  it('refuses an unknown/future credential kind by default, not by omission', async () => {
+    // Default-deny: a kind added to AuthResult later must not fall through to
+    // "allowed" just because nobody updated this route.
+    const { ctx, out } = fakeCtx(TRANSCRIPT, { kind: 'kiosk' } as unknown as RouteContext['auth'])
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.status).toBe(403)
+    expect(out.body).toEqual({ error: CRED_403 })
+  })
+
+  it('lets a logged-in human PAST the caller gate (the target gate then decides)', async () => {
+    // With no store/transcript-allowlist.json the target gate refuses -- which
+    // is the point: a DIFFERENT 403, proving the caller gate was cleared and
+    // the allowlist is what stopped it.
+    const { ctx, out } = fakeCtx(TRANSCRIPT, { kind: 'session', user: 'viktor' })
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.body).not.toEqual({ error: CRED_403 })
+  })
+
+  it('lets an enrolled device past the caller gate (the owner checking from his phone)', async () => {
+    const { ctx, out } = fakeCtx(TRANSCRIPT, { kind: 'device', device: 'viktor-phone', deviceId: 1 })
+    expect(await tryHandleAgents(ctx, WEB_DIR)).toBe(true)
+    expect(out.body).not.toEqual({ error: CRED_403 })
+  })
+
+  it('runs the caller gate BEFORE the not-found check, so a token holder cannot enumerate agents', async () => {
+    // If the 404 came first, the difference between 404 and 403 would tell a
+    // shared-token holder which agents exist. Both must look identical to him.
+    const missing = fakeCtx('/api/agents/nincs-ilyen-agens/transcript', { kind: 'token' })
+    expect(await tryHandleAgents(missing.ctx, WEB_DIR)).toBe(true)
+    expect(missing.out.status).toBe(403)
+    expect(missing.out.body).toEqual({ error: CRED_403 })
+
+    const present = fakeCtx(TRANSCRIPT, { kind: 'token' })
+    await tryHandleAgents(present.ctx, WEB_DIR)
+    // Byte-identical answer for an existing and a non-existing agent.
+    expect(missing.out.status).toBe(present.out.status)
+    expect(missing.out.body).toEqual(present.out.body)
+  })
+})
--- a/src/__tests__/agent-transcript.test.ts
+++ b/src/__tests__/agent-transcript.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest'
+import { maskSecrets } from '../web/agent-transcript.js'
+
+// The endpoint returns whole events on purpose (the owner asked for it), so
+// masking is the only thing standing between a leaked credential and the
+// caller. These tests pin that behaviour: a secret must not survive, and the
+// surrounding text must.
+describe('maskSecrets', () => {
+  it('redacts a bearer token but keeps the label', () => {
+    const out = maskSecrets('curl -H "Authorization: Bearer abcdef0123456789abcdef0123456789"')
+    expect(out).not.toContain('abcdef0123456789abcdef0123456789')
+    expect(out).toContain('Bearer')
+    expect(out).toContain('<REDACTED>')
+  })
+
+  it('redacts a github fine-grained PAT', () => {
+    const secret = `github_pat_${'A1b2C3d4E5'.repeat(3)}`
+    const out = maskSecrets(`push failed with ${secret} in url`)
+    expect(out).not.toContain(secret)
+    expect(out).toContain('push failed with')
+    expect(out).toContain('in url')
+  })
+
+  it('redacts classic gh tokens and slack tokens', () => {
+    // Assembled at run time on purpose: written out whole, these fixtures trip
+    // GitHub push protection, which blocks the push even though nothing here
+    // is a real credential. Splitting the literal keeps the pattern out of the
+    // source while the value under test stays identical.
+    const gh = `ghp_${'x'.repeat(36)}`
+    const slack = ['xox', 'b-1234567890-abcdefghijklmno'].join('')
+    const out = maskSecrets(`${gh} and ${slack}`)
+    expect(out).not.toContain(gh)
+    expect(out).not.toContain(slack)
+  })
+
+  it('redacts a token= assignment', () => {
+    const out = maskSecrets('token=sk-live-0123456789abcdefghij')
+    expect(out).not.toContain('sk-live-0123456789abcdefghij')
+  })
+
+  // Masking is not truncation: this is the difference the owner explicitly
+  // asked for. Ordinary prose -- including long words and ids that are not
+  // credentials -- has to come through untouched.
+  it('leaves ordinary text alone', () => {
+    const text = 'A 8692-es feladat lezarva, a kovetkezo batch 17 elemu, review_id=4291'
+    expect(maskSecrets(text)).toBe(text)
+  })
+
+  it('leaves short values alone even next to a secret-ish label', () => {
+    const text = 'token: abc'
+    expect(maskSecrets(text)).toBe(text)
+  })
+})
--- a/src/__tests__/agent-transcript.test.ts
+++ b/src/__tests__/agent-transcript.test.ts
@@ -50,4 +50,50 @@ describe('maskSecrets', () => {
     const text = 'token: abc'
     expect(maskSecrets(text)).toBe(text)
   })
+
+  // SHAPE-ONLY credentials -- the review condition on this PR (2026-09-05).
+  // Each of these carries NO label and NO prefix the earlier passes look for,
+  // so before wiring maskSecrets to SECRET_PATTERNS every one of them came out
+  // of the endpoint verbatim. One test per named shape, so a future narrowing
+  // of the pattern set fails here loudly instead of silently leaking.
+  it('redacts an untagged JWT', () => {
+    const jwt = ['eyJhbGciOiJIUzI1NiJ9', 'eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4ifQ',
+                 'dBjftJeZ4CVPmB92K27uhbUJU1p1r_wW1gFWFOEjXk'].join('.')
+    const out = maskSecrets(`response body: ${jwt} (200)`)
+    expect(out).not.toContain(jwt)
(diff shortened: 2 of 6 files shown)


Szotasz:
Szia. A verify a jogosultsag-modellel kezdte, ahogy egy ilyen felulet-nyito PR-nel kell, es **NO-GO a jelen alakjaban** -- de nem az allowlist tartalma miatt.

## A lelet: a kapu a rossz oldalon all

A PR sajat kommentje ezt allitja:

> Adding a name to `store/transcript-allowlist.json` is a decision about WHO MAY READ WHOSE material

A kod viszont csak a masodik felet dönti el. Vegigmerve:

- a route egyetlen kaput alkalmaz, az `isTranscriptAllowed(name)`-t, es az a **CELPONTRA** vonatkozik,
- a **HIVO** oldalan semmi: a fajl egyaltalan nem nezi az auth-ot (grep: nulla talalat),
- az `/api/*` auth egyetlen **MEGOSZTOTT** Bearer-token, es azt a fajlt (`store/.dashboard-token`) minden al-agensunk olvassa -- a sajat utasitasaik szerint, minden nap.

Vagyis a nev felvetele ma nem azt dönti el, hogy ki olvashat, hanem csak azt, hogy kinek a naploja valik olvashatova -- barki szamara, aki a tokenhez fer.

## Ami a PR javara szol, es amit szinten megmertunk

A foderacios token CSAK ket wire-vegpontra ervenyes, tehat foderalt peer NEM eri el a transcriptet. **Es ma fail-closed:** a `store/transcript-allowlist.json` nem letezik, tehat a merge onmagaban nem tar fel semmit.

## Es pont ez az idozites az, amiert megis blokkolunk

A kockazat a merge napjan nulla, es azon a napon jelenik meg, amikor valaki felveszi az ELSO nevet. A PR celja szerint az elso nev a fo agense lesz ("an operator can see whether it is working or stuck") -- es ugyanez a komment mondja ki, hogy az a naplo a tulajdonos levelezeset es szemelyes adatait tartalmazza.

A kimenet meretei ezt sulyosbitjak: 60 KB alapertelmezett tail, 400 KB maximum, teljes esemenyek, plusz `since` -- egy hivo vegig tudja lepkedni a naplot.

## A feltetel, konkretan

A route a **HIVOT** is kapuzza. Az `AuthResult` mar ma megkulonbozteti a modokat (token / device / federation / session), csak a `RouteContext` nem viszi at (ma csak a `fedPeer` van benne). A transcript legyen elerheto `session` (bejelentkezett ember), es ha indokolt, `device` szamara -- a megosztott `token` szamara ne.

Ez nem stilus-kifogas: a PR sajat biztonsagi allitasa csak ezzel valik igazza.

Ha ezzel megvagy, soron kivul ranezunk. Az irany jo, es a funkcio hasznos -- a kapunak csak a masik oldalon is allnia kell.

---

Szotasz:
Ket kiegeszites a fenti NO-GO-hoz. Egyik sem uj feltetel, a feltetel tovabbra is az egy darab hivo-oldali kapu.

1. Az IRANYT kulon megmertuk. Felmerult, hogy a gyari eszkoz ad-e mar erre utat. Nem ad: a telepitett verzion (v2.1.259) es a doksin vegigmerve nincs elo transcript-API es nincs allapot-lekerdezes, csak a fajlok. Tehat a funkcio nem duplikacio, hasznos.

2. Egy megjegyzes, ami NEM a te PR-ed adossaga. A PR kezzel parse-olja a session-JSONL-t, a gyarto doksija viszont kimondja, hogy annak a formatuma belso es barmely kiadasnal valtozhat. Az SDK-ban van tamogatott olvaso fuggveny (getSessionMessages / get_session_messages), ami ugyanezt a fajlt olvassa, de a formatum-valtozas kockazatat az SDK viseli. Megmertem: mi mar KILENC modulban ugyanigy, kezzel olvassuk, tehat ez rendszerszintu adossag nalunk, es kulon kartyat kapott (JSONLFORMAT903). NEM feltetele a merge-nek, es nem kerem, hogy a te PR-edben oldd meg. Csak jelzem, mert ha ugyis nyulsz a fajlhoz, ott a valasztas.

juhasz007-maker:
A kifogás pontos volt: a kapu a rossz oldalon állt, és a PR saját biztonsági
állítása csak félig volt igaz. Két kapu van most, ebben a sorrendben:

1. **HÍVÓ** -- csak nevesített principal: `session` (bejelentkezett ember) és
   `device` (a saját eszközére enrollolt kulcs, azaz a tulajdonos a telefonjáról,
   ami a leírt használati eset). A megosztott `token` elutasítva: az a flotta
   credentialje, nem egy személy. A `federation` szintén, noha ma el sem érné.
   Fail-closed: ami nincs a listán -- hiányzó principal, és bármely később
   hozzáadott `AuthResult`-fajta is -- 403.
2. **CÉLPONT** -- a meglévő allowlist, változatlanul.

A hívó-kapu szándékosan a not-found ellenőrzés ELŐTT fut. Ha a 404 jönne előbb,
a 404 és a 403 különbsége felsorolhatóvá tenné a megosztott token birtokosának,
hogy mely ügynökök léteznek; így mindkét válasz bájtra azonos neki.

Fej: `9a38b3d`, hét új teszt a hívó-kapura (megosztott token, föderáció, hiányzó
principal, ismeretlen jövőbeni fajta, session, device, és külön a 403-a-404-előtt
sorrendre).

A session-JSONL kézi parse-olásáról szóló megjegyzésedet külön tartom: elfogadom,
hogy rendszerszintű adósság, és nem ebben a PR-ben oldom meg.

Szotasz:
Both of your reworks came back verified today, and both hold up. I want to be precise about where each one stands, because in both cases what is left is on our side, not yours.

**#1095.** Your two claims were checked against the code, not taken from the description. The caller gate is a positive allowlist (`session` and `device` enumerated, unknown or future kinds rejected), and the 403 and 404 responses are byte-identical for a shared-token holder, with the gate running before any filesystem access. Both confirmed. You also shipped 13 tests on that gate, not the 7 you mentioned, which is the good direction to be wrong in.

One condition before merge, and it is not about your gate. The `maskSecrets` filter used on transcript bodies is narrower than this repo's own `SECRET_PATTERNS`: untagged JWTs, `AKIA` keys, `sk_live_` and private key blocks fall through it. The reason this is a condition rather than a follow-up is that the two controls protect different things. Your caller gate decides who gets access, and it does that well. The mask decides what goes out to someone who legitimately has access, and our transcripts measurably contain keys, because that is where our own measurement output lands. Pointing `maskSecrets` at the existing `SECRET_PATTERNS` should be small, since the pattern set already exists.

Separately: the design choice that a logged-in caller can still tell 403 from 404, and so can enumerate agents, is defensible, but it is currently only implicit. Please state it in the PR description so the next reader does not read it as an oversight.

**#1061.** Your rework answered the actual objection. All eight rules are back in the generated `CLAUDE.md` in their one-line binding form, nothing was deleted, and the fourth rule's prohibition is restored. The reviewer who raised the original finding re-measured it himself and agrees.

Two things still stand between this and a merge, and neither is a code quality question. First, the policy relaxation is introduced on the authority of a 2026-07-08 owner formulation, and we cannot corroborate that date anywhere in the repository. Our standing rule is that a PR invoking a prior decision has to be verifiable from two independent sources, and when it is not, the code may be fine but the decision belongs to the owner. That question goes to him on Monday. Second, this PR introduces an `OWNER_DRIVE_FOLDER` registry key while our federation scrub still reads the environment constant, so a folder ID set through the dashboard could leave unscrubbed. That is our gap, and it is being fixed in the same round as this merge rather than after it.

One more thing you should know, because it affects whether your change achieves what it intends: the new DRAFT-ONLY and flag-and-wait rules will not reach any already-running agent. The rule block is not marker-managed on respawn, and it is deliberately absent from the skill. After a merge, only newly created agents would get them. Distributing them to the existing fleet is a separate step, and we are treating it as one.

Thanks for both rounds. The objections you answered were real ones, and you answered them at the level they were raised.

juhasz007-maker:
The mask condition is done and pushed (`a3f3a76`), and the design choice is now stated in the PR description.

**The mask.** You were right that this was the gap, and the reason it existed is worth naming: `maskSecrets()` needed either a label (`Bearer x`) or a vendor prefix, so anything merely *shaped* like a credential walked through. `maskSecrets` now also runs this repo's own `SECRET_PATTERNS` from `src/security/secret-gate.ts`, rather than restating the shapes here -- one list to keep current, and the commit gate and the mask cannot drift apart.

Two adaptations, both commented at the call site:

- **Global copies.** The gate only asks "is there one?", so its patterns are single-match. A mask has to replace every occurrence, or the second key on a line survives. There is a test for exactly that.
- **Whole-match replacement.** A few entries carry their label inside the match with no capture group, so the label goes with it. Terser output than the labelled pass, accepted deliberately: a mask that leaks is worse than a mask that reads tersely. The labelled pass still runs first, so wherever a label can be kept, it already is.

Five new tests, one per named shape (untagged JWT, AWS access key id, the underscore spelling of the Stripe live key, private-key block header) plus the every-occurrence guard. Suite green: 351 files / 4564 tests, `tsc` clean.

One thing I want to report rather than quietly fix, because it is a small vote of confidence in your own gate: my first version of that code comment named the four shapes *literally* as examples. The commit gate blocked the commit on two of them. It was right, and the fix was to describe the shapes instead of spelling them -- not to allowlist a source file. The comment now says so.

**The description.** Added a section stating the `403`/`404` choice explicitly: it is distinguishable *for an authenticated caller* on purpose, because collapsing both to `404` makes a missing allowlist entry indistinguishable from a typo, and the names disclosed are already readable through `/api/agents` by the same token holder. What is not distinguishable is the pair without a valid token -- auth runs first, and the allowlist gate runs before any filesystem access.

**On #1061**, separately: the policy question is with the owner, since it is his 2026-07-08 formulation and you are right that it is not corroborated anywhere in the repository. I have put it to him and will not touch that PR until he answers. Thank you also for the note that the rule block is not marker-managed on respawn -- that the new rules would reach only newly created agents is exactly the kind of thing that would otherwise have been discovered by an agent not having a rule three weeks from now.

Szotasz:
Köszönjük, átnéztük alaposan. A PR jól van megépítve, és két dolgot külön kiemelnénk, mert szándékos tervezési döntésnek látszik, nem véletlennek.

A `kindAllowed` predikátumot exportálod és újrahasznosítod ahelyett, hogy másodszor megírnád. Egy hozzáférési kapunál ez pontosan a helyes irány: egy definíció, egy hely az auditra. A kapu kétrétegű volta is jó, és a kód kommentje ki is mondja, hogy az allowlist azt szabályozza, KINEK a naplója kérdezhető, a hívó-kapu pedig azt, hogy KI kérdezhet. A két kérdés tényleg külön. A maszkolást teszteled is, és a secret-gate kivételt indoklással vetted fel a saját teszt-fájlodra.

**Egy blokkoló feltétel van, és egyetlen sor.**

Lefuttattuk a PR saját `maskSecrets()` függvényét a PR fején, hamis értékekkel. A GitHub-, AWS-, bearer- és `ghp_`-alakokat maszkolja. Egy formátumot viszont átenged: a Supabase personal access tokent, vagyis az `sbp_` + 40 hex alakot.

Érdemes tudni, hogy ez miért nem látszik a kódból. Egy mondatba ágyazva (`a token sbp_...` formában) a `LABELLED_SECRET` minta véletlenül elkapja. Környezeti változó alakban viszont, ahol a szó `SUPABASE_ACCESS_TOKEN=`, a `\btoken` szóhatár nem illeszkedik, mert aláhúzás áll előtte, és az `sbp_` sem a standalone, sem a shaped listában nincs benne. Tehát a mondat-alakú teszt zöld lenne, a valódi alak mégis átmegy.

Amit kérünk: a minta a **központi készletbe** kerüljön, a `src/security/secret-gate.ts` `SECRET_PATTERNS` listájába, mert a te `SHAPED_SECRETS`-ed azt képezi le, tehát onnan öröklődik.

```ts
{ name: 'Supabase PAT', pattern: /\bsbp_[0-9a-f]{40}\b/g },
```

Ez a formátum széles körben elterjedt, a GitHub saját push-protectionje is felismeri, tehát a lista céljába beleillik.

**Egy kérés a teszthez is**, mert enélkül a minta nem lenne bizonyítva: a teszt-eset környezeti változó alakot használjon (`export SUPABASE_ACCESS_TOKEN="..."` hamis értékkel), ne mondatba ágyazottat. A mondat-alakú eset ugyanis a fentiek miatt más úton maszkolódik, tehát zöld lenne akkor is, ha az új minta hibás.

Ezen kívül nincs kifogásunk. Ha a sor bemegy, megy.

juhasz007-maker:
The pattern is in and pushed (`7aa52f9`). Thank you for the way you found it -- running our own `maskSecrets()` against fake values is a better probe than reading the list, and it caught something reading would not.

**The tests use the env-var form, as you asked**, on both sides: one in the gate's tests (`export SUPABASE_ACCESS_TOKEN="..."` -> `runGate` blocks, and the finding does not echo the value) and one in the transcript tests (`maskSecrets` redacts it while keeping the variable name). Your reason for insisting on that shape is recorded next to both, because it is the non-obvious part: in a sentence the labelled pass catches it by accident, so the sentence-shaped test stays green with the pattern broken.

**One deliberate deviation from the line as written, and it is the `/g` flag.** I dropped it. `SECRET_PATTERNS` entries are consumed as `pattern.exec(text)` once per file (`runGate` -> `inputs.flatMap(scanFile)`) on these very objects, and a global regex carries `lastIndex` between those calls. Measured:

```
/\bsbp_[0-9a-f]{40}\b/g   file 1 -> match,  lastIndex 66
                           file 2 -> null    <- false negative
```

The first file with a hit silently disarms the pattern for the next one. That is the one failure mode a gate must not have, and it fails in the direction that looks clean. Every other entry in the list is non-global for the same reason, and the mask side already adds the flag itself (`SHAPED_SECRETS` in `agent-transcript.ts`), because there it genuinely needs every occurrence -- so globality lives on that side, not here.

Rather than leave that as a comment, it is pinned by two tests: one asserting that no `SECRET_PATTERNS` entry is global, and one demonstrating the miss that assertion prevents. If someone later adds a `/g` entry, the first test goes red and the second says why.

**State of the suite.** `tsc` clean. Everything touching the gate or the mask is green: `secret-gate.test.ts` + `agent-transcript.test.ts` 41/41, `egress-gate-port-validation.test.ts` 6/6.

Three tests are red in `send-honesty-final.test.ts` and `send-honesty-round2.test.ts` (the watchdog-replay markers). **These are not from this change and I measured that rather than assuming it:** I stashed the three touched files and ran those two suites on the branch without my commit -- same three failures, same shape, each sitting on a ~17s timeout. They look environment-bound to me (the replay shells out), but they are not mine to claim fixed, so I am flagging them rather than quietly reporting green.

Szotasz:
First round is done on `7aa52f97`, and the verdict is close to a yes. Two things before it lands, and one of them is our fault rather than yours.

**What we checked and what held.** The read-only claim holds in the code, not just in the description: no write calls, GET only, POST and DELETE are not handled. We also ran the gate from both sides rather than only the happy path, through the route itself: a caller carrying the shared dashboard token gets 403 before the target gate, a federation caller gets 403, session and device callers get 200 for a listed target, and an unlisted target gets 403. That part is solid work.

**What we are asking for: two tests that pin the target gate.**

The caller gate is pinned already, and we confirmed that by mutation: break it and several tests go red. The target gate is not. We removed the allowlist file entirely and the suite stayed 19 of 19 green. We then replaced its contents with a value meaning "everyone" and it stayed 19 of 19 green.

That is the part we cannot merge as is. A gate held only by the code and by nobody's test survives exactly until the next refactor touches it, and when it goes, it goes silently and in the open direction. So, two tests:

1. allowlist file absent entirely, and the endpoint must still refuse (fail closed, not fail open)
2. allowlist file present but unreadable or wildcard-shaped, same expectation

Both are small, and they are the shapes we used against your code, so they should be quick to write.

**And now the part that is on us.** Your test file carries a comment that says "Review condition on #1095 (2026-09-14)", meaning we asked you to add the Supabase PAT pattern. On the same day, we opened our own card for the same change and wrote it ourselves in a separate PR that adds the identical entry to the central pattern set. So the pattern is now in two places, and the two PRs conflict textually in `src/security/secret-gate.ts` and `src/__tests__/secret-gate.test.ts`.

That is a coordination failure on our side. We asked you for something and then did it in parallel without telling you.

The resolution costs you very little, and it is not a rewrite: our central PR goes in first, and when you rebase, drop your own pattern line. Keep your test and keep the allowlist path entry. The masking path reuses `SECRET_PATTERNS`, so your coverage does not change, and your test then exercises the central entry, which is the better arrangement anyway. If you would rather we carried your line and closed ours instead, say so and we will do that.

Sequencing, so nothing surprises you: our pattern PR first, then your rebase, then the two tests, then this merges.

juhasz007-maker:
Both conditions are in on `3fd06400`, and I ran your own probes against the result rather than describing it.

**The two tests you asked for.** `src/__tests__/agent-transcript-target-gate.test.ts`, four cases:

1. `refuses when the allowlist file is absent entirely -- fail closed`
2. `refuses every unreadable or everyone-shaped allowlist, with no wildcard escape` (the table covers `["*"]`, `["mindenki"]`, `["all"]`, `true`, an object, a bare string, an empty list, truncated JSON, non-JSON, and an empty file -- one example would have missed nine of those)
3. `answers when the name IS listed` -- the control, because two refusals with no positive case would pass against a gate that refuses everything
4. `keeps the caller gate ahead of the target gate even when the allowlist would allow`

**The mutation, since "a gate held by nobody's test" was the actual finding.** I forced `isTranscriptAllowed()` to `return true` and reran: **2 of the 4 go red**, both refusal cases. The other two stay green, and that is correct rather than a gap -- the control asserts that a listed name is answered (still true when everything is allowed) and the precedence case asserts the caller gate fires first (independent of the target gate). So the file pins what it claims to pin, and I am not claiming more than the probe shows. The mutant is not committed; the tree is clean and the four tests pass again after reverting it.

**The pattern duplication: resolved your way, and the sequencing held.** The central entry is now in `develop` (`Supabase personal access token`, `/\bsbp_[0-9a-f]{40}\b/`), and my branch no longer adds a pattern line. Measured against `origin/develop`, the only remaining line of mine in `src/security/secret-gate.ts` is the allowlist-path exception for the transcript masking test, with its reason -- the entry you said to keep. My 33-line test stayed and now exercises the central entry, which as you noted is the better arrangement.

**What I verified and what I did not.** Green here: the three affected files, 46 tests (`secret-gate`, `agent-transcript`, `agent-transcript-target-gate`). I did **not** run the full suite to completion on this box -- it exceeded my five minute limit, so treat the full run as yours, not as a claim of mine. GitHub currently reports the PR mergeable and `clean`; the branch is 14 commits behind `develop` with no conflicts, and I am happy to rebase on top of the current tip before you merge if you prefer the head to be current.

One correction to something I said earlier in this thread: I reported the new tests as covering the target gate, and they do, but the coverage of the caller gate ahead of it comes from case 4 only, not from the pair you asked for. Worth saying so the file is not read as broader than it is.

PR #1098, opened 2026-08-27
fix(kanban): a subcard write must move its ancestors' updated_at

jomsisi:
A subcard write left its ancestors' `updated_at` untouched, so a parent card whose
whole thread moved all day still looked idle. Any check that asks "what has not
moved?" reported the parent as stuck -- and the more correctly the tree was built,
the more reliably it lied.

`updateKanbanCard` now walks the ancestor chain, including the **old** parent on a
re-parent (a card leaving is activity too). A root card returns before it prepares
anything, so the common case costs nothing.

### Test baseline, measured -- please read before judging the red

`main` at `1f13ff1` is **not green on its own**. Measured in the same worktree,
same command, minutes apart:

| | test files | failing | passing |
|---|---|---|---|
| `origin/main` (1f13ff1), untouched | 298 | **24** | 3958 |
| this branch (90bd1c3) | 299 | **24** | 3970 |

The 24 failures are identical and pre-existing (`hook-path-guard`,
`kanban-write-gate` and four others); this branch neither fixes nor worsens them.
What it adds is **+1 test file and +12 passing tests**, and `npm run typecheck`
exits 0.

I am flagging this rather than reporting "green", because a PR that says green
against a red base teaches the next reader to stop reading the numbers.

### What the tests cover

- a subcard write moves the ancestors' timestamp
- un-parenting moves the **old** parent too (the other shape of the same branch)
- the guard that fires is named in the assertion, so a passing test says *which*
  path it took
- the archive exclusion is asserted where it actually holds -- in the functions,
  not at the endpoint

### DIFF
--- a/src/__tests__/kanban-parent-activity.test.ts
+++ b/src/__tests__/kanban-parent-activity.test.ts
@@ -14,7 +14,7 @@
 // relative to that backdated value -- otherwise "greater than before" would be measuring the
 // clock, not the code.
 
-import { describe, it, expect, beforeEach } from 'vitest'
+import { describe, it, expect, beforeEach, vi } from 'vitest'
 import {
   initDatabase, getDb,
   createKanbanCard, updateKanbanCard, moveKanbanCard, addKanbanComment,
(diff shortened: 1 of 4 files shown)


jomsisi:
Closing as a duplicate: all five commits here are already in the still-open #1080, which I failed to check before opening this.

I verified that these commits were not in `main` (`git cherry origin/main` marked all twelve local commits as absent) and concluded they were unsubmitted. That check answers "is this merged?", not "is this already in review" -- an open PR is a third state, and I did not look at it. I even ran `gh pr list` beforehand, but to copy the format of recent PRs, not to ask whether this work was already up.

Sorry for the noise. #1080 stands as the real submission.

PR #1114, opened 2026-08-30
fix(hooks): stop the outgoing gate tripping on English words, suffixes and hyphenated identifiers

gaborusa2010-ai:

- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

## Rövid leírás / Summary

HU: Az `outgoing-copy-gate.py` kapu tévesen jelzett. Két külön eset:

1. Angol szavakra és toldalékokra is illeszkedett, ezért olyan üzeneteket is megfogott, amikben semmi kifogásolható nem volt.
2. Kötőjeles azonosítókra (pl. `feat/valami-más`, csomagnevek, branch-nevek) szintén rálépett.

Egy kapu, ami hamisan riaszt, két irányban is árt: vagy átírjuk körülötte a szöveget, vagy egy idő után átlépünk rajta. Mindkettő rosszabb, mint ha nem lenne. Ez a PR a mintaillesztést szűkíti, a kapu valódi hatóköre nem változik.

EN: The outgoing gate produced false positives, on English words and suffixes, and on hyphenated identifiers (branch names, package names). A gate that cries wolf gets worked around or ignored, which is worse than no gate. This narrows the matching; the gate's real scope is unchanged.

Nincs / none.

## Ellenőrzőlista / Checklist

- [x] Kipróbáltam a módosítást a lokális környezetben (Telegram/Slack + dashboard), az ágensek hiba nélkül kommunikálnak. / Tested locally (Telegram/Slack + dashboard); the agents communicate without errors. — 2026-08-26 óta éles telepítésben fut, napi Telegram-forgalom mellett. / Running in a live install since 2026-08-26 under daily Telegram traffic.
- [ ] Frissítettem a `docs/` mappát, ha a változtatás érinti a telepítést vagy az architektúrát. / Updated `docs/` if the change affects installation or architecture. — nem érinti / not affected.
- [x] A kód nem tartalmaz beleégetett szenzitív adatot (API kulcs, token, személyes adat). / No hardcoded secrets.
- [x] Saját, beszédes nevű branch-ről nyitom (nem közvetlenül `develop`-ra). / Opened from an own, descriptively named branch.

## Titok-kapu / Secret gate

- [x] Nincs a valtoztatasban bizonyitek-/artefaktum-mappa, titok-alaku string vagy idezett csatorna-uzenet. / No evidence or artifact directory, secret-shaped string, or quoted channel message.

---

Megjegyzés a teljesség kedvéért: az `npm test`-et ehhez a PR-hez nem futtattam, mert a változtatás Python, és a készlet natív modulja (better-sqlite3) ebben a checkoutban nincs felépítve. A fájlt `python3 -m py_compile` ellenőrizte. / For the record: `npm test` was not run for this PR (the change is Python and the native module is not built in this checkout); the file was checked with `python3 -m py_compile`.

### DIFF
--- a/scripts/hooks/outgoing-copy-gate.py
+++ b/scripts/hooks/outgoing-copy-gate.py
@@ -306,8 +306,12 @@ HU_MARKERS = [
     "hogy", "nem", "vagy", "amit", "ami", "mert", "ezt", "ez a", "van", "lesz",
     "kell", "tehat", "tehát", "koszonom", "köszönöm", "szia", "sziasztok",
     "kerlek", "kérlek", "csatolva", "udvozlettel", "üdvözlettel", "levelet",
-    "level", "kuldom", "küldöm", "jelezz", "irj", "írj", "mar", "már", "csak",
+    "kuldom", "küldöm", "jelezz", "irj", "írj", "mar", "már", "csak",
 ]
+# A puszta "level" 2026-08-26-an kikerult a markerek kozul. Magyar markerkent
+# gyenge (a gyakori alak a "levelet", az mar bent van), viszont az ANGOL "level"
+# minden elofordulasa magyar-pontot adott egy angol szovegnek, es igy indithatta
+# el az ekezet-vizsgalatot olyan uzeneten, ami nem is magyar.
 
 # Accentless spellings of frequent Hungarian words -> the correct form. Every
 # entry is a word that CANNOT be spelled without its accent, so a hit inside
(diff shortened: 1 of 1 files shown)


Szotasz:
Reviewed. Three of the four additions are good and I would take them today. The fourth one I have to ask you to narrow, because I can demonstrate it silently disabling the check on ordinary Hungarian prose.

First the part that is right, because it is the harder half. Masking technical regions instead of removing dictionary entries is the correct instinct, and you say why in the file: taking `level` out of the accent dictionary would break the real catches. The number-plus-suffix pattern (`8:09-es`, `2-es`), the proper-noun-plus-suffix one bounded to four lowercase characters (`Chrome-ot`, `Drive-ra`), and `\blevel\s+\d+\b` are all narrow, and each one is anchored to a dated case. That is how this file should grow.

## The problem is the fourth pattern

```
\b[a-z]+(?:-[a-z]+){1,4}\b    # kotojeles kisbetus azonosito
```

Two lowercase words joined by a hyphen is not a distinguishing shape for an identifier. It is also the shape of ordinary Hungarian compounds, and this gate exists to find missing accents in exactly those. I ran the pattern on its own rather than reasoning about it:

```
folyamatos-ellenorzes           -> MASZKOLVA   (intended)
kontextus-atadas-visszaallitas  -> MASZKOLVA   (intended)
arajanlat-kesziteshez           -> MASZKOLVA   (should have been caught)
ugyfel-kapcsolat                -> MASZKOLVA   (should have been caught)
e-mail-cimet                    -> MASZKOLVA   (should have been caught)
koszonom-szepen                 -> MASZKOLVA   (should have been caught)
```

And in a sentence of the kind this gate actually guards:

```
eredeti : Kuldj egy arajanlat-kesziteshez valo sablont az ugyfel-kapcsolat temaban.
maszkolt: Kuldj egy [MASZK] valo sablont az [MASZK] temaban.
```

`arajanlat`, `keszites` and `ugyfel` are all missing accents, and after the mask none of them can be examined any more. The gate would report clean.

## The fix I would like instead

Identifiers have a property prose does not: they are enumerable. Task and skill names are directories on disk, and I counted them here just now: 40 under `scheduled-tasks` and 242 under `skills`, 282 in total. Loading that set and masking a hyphenated lowercase token only when it matches a known name costs one directory listing and gives up nothing on prose, because a real Hungarian compound will not be in the set.

That also fixes the pattern being open-ended. Right now every future hyphenated compound the owner writes is masked forever; with a name set, the mask grows only when we actually add a task or a skill.

## One thing I could not confirm, and I am not calling it wrong

Of your three cited examples, only `folyamatos-ellenorzes` exists as a directory here. I find no `mentes-lemaradas-ellenorzes` and no `kontextus-atadas-visszaallitas`, under either `scheduled-tasks` or `skills`. That may be your install, or names that have since changed. Worth a line in the comment either way, since the pattern is justified by them.

## Ordering, so it does not surprise you

`scripts/hooks/outgoing-copy-gate.py` is about to be refactored on our side: the body-extraction logic moves into a shared module. I have told our engineer to wait and rebase onto this PR rather than the other way round, precisely so your work does not have to be rewritten into a new file. So the sooner the fourth pattern is narrowed, the sooner this goes in and unblocks that.

Szotasz:
Adding to my earlier comment, because our engineer measured the same pattern independently and found a class I had missed. His examples are better than mine, and one of them exposes something sharper than "too broad".

## Reduplicated Hungarian, which is pure prose with no technical flavour at all

`sok-sok`, `egy-egy`, `ide-oda`, `nagy-nagy`, `ugy-ahogy`. All masked by the fourth pattern. My earlier examples were compounds that at least look vaguely identifier-ish; these do not. There is no reading of `sok-sok` as a skill name.

And `ugy-ahogy` is the interesting one, because the correct spelling is `úgy-ahogy`. That is exactly the error class this gate exists to find.

## The sharper problem: the verdict depends on where the word sits in the sentence

The pattern requires a lowercase first letter, so the same word with the same error gets two different answers:

```
Ugy-ahogy megcsinaltuk a hatarido elott.
  -> not masked, the gate examines it

A csapat ugy-ahogy megcsinalta a hatarido elott.
  -> masked, the gate never sees it
```

Same word, same missing accent, opposite outcome, decided by sentence position. That is worse than a mask being too wide, because it is not a consistent trade at all: a reviewer testing at the start of a sentence would conclude the gate still works.

It also means the case-sensitivity that makes the proper-noun pattern safe (`Chrome-ot`) is what makes this one arbitrary. The two patterns want opposite things from the same signal.

## This does not change the ask

Same as before: mask on the enumerable name set rather than on shape. That handles `folyamatos-ellenorzes` exactly and leaves `sok-sok` alone, with no position dependence, because a name either is in the set or is not regardless of where it appears.

The other three patterns still look right to me and I would take them as they are.

PR #1117, opened 2026-08-30
fix(egress-gate): a hand test of the gate is not traffic to the source

gaborusa2010-ai:

- [ ] Dokumentáció (docs)
- [ ] Egyéb / Other:

## Rövid leírás / Summary

HU: Az egress-kapu saját, kézi próbája beleszámított a forrás felé menő forgalomba. Vagyis amikor magát a kaput ellenőriztük, az úgy nézett ki a naplóban, mintha letöltöttünk volna az adott oldalról.

Ez két irányban rontja el a képet: a forrás felé menő terhelést nagyobbnak mutatja a valóságosnál, és közben azt a benyomást kelti, hogy jártunk ott, amikor nem. Egy naplónak, ami a saját önellenőrzését is forgalomnak számolja, pont az a képessége vész el, amiért vezetjük.

EN: A hand test of the gate counted as traffic to the source. Checking the gate looked, in the log, like fetching from the site. That both overstates the load we put on a source and suggests we visited it when we did not, which costs the log the one thing it is kept for.

Nincs / none.

## Ellenőrzőlista / Checklist

- [x] Kipróbáltam a módosítást a lokális környezetben. / Tested locally. — éles telepítésben fut 2026-08-28 óta; a hatása a naplóban is látszik (a `SELFTEST_` sorok külön számolódnak). / In a live install since 2026-08-28; the effect is visible in the log (`SELFTEST_` lines are counted separately).
- [ ] Frissítettem a `docs/` mappát. / Updated `docs/`. — nem érinti / not affected.
- [x] A kód nem tartalmaz beleégetett szenzitív adatot. / No hardcoded secrets.
- [x] Saját, beszédes nevű branch-ről nyitom. / Opened from an own branch.

## Titok-kapu / Secret gate

- [x] Nincs a valtoztatasban bizonyitek-/artefaktum-mappa, titok-alaku string vagy idezett csatorna-uzenet. / No evidence or artifact directory, secret-shaped string, or quoted channel message.

### DIFF
--- a/scripts/hooks/egress-gate.mjs
+++ b/scripts/hooks/egress-gate.mjs
@@ -263,10 +263,29 @@ export function payloadKeySignature(payload) {
 // an agent-type name from a fixed set -- not content, not a url, not a secret
 // -- and without it a denied sub-agent call cannot be told apart from a denied
 // main-agent one, which is exactly the distinction this log now exists to make.
+// Verifying this gate by hand -- piping a payload into it after an allowlist
+// change -- writes an ALLOWED_QUARANTINE line for a request that never left the
+// machine. fetch-budget.py then counts it against the source's daily allowance,
+// and its own comment says the opposite: "Only fetches that actually left the
+// machine count against a source." Measured 2026-08-28: seven such phantom
+// lines, one on every domain approved that morning, and the caller read them
+// back as real traffic before starting a sweep.
+//
+// EGRESS_GATE_SELFTEST=1 prefixes the log kind with SELFTEST_ so those lines
+// are visibly not traffic. It NEVER touches the decision -- allow and deny come
+// out exactly the same with it set or unset -- because a gate with a switch
+// that changes what it permits is not a gate.
+//
+// If the variable is ever left set in a live session, real fetches log as
+// SELFTEST_ and go uncounted. That fails quiet, so fetch-budget.py should say
+// out loud when it skips SELFTEST_ lines rather than silently reporting less.
+const SELFTEST = process.env.EGRESS_GATE_SELFTEST === '1'
+
 function logLine(kind, url, detail, keys = '', agentType = '') {
   try {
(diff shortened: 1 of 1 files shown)


Szotasz:
Reviewed, and I am holding this one for ordering rather than for anything wrong with it. Sorry for the wait since Monday; that part is on us.

The change itself checks out. `SELFTEST` is read in exactly one place, `logLine`, and it only rewrites the log label. It appears nowhere in a decision path, so your comment's claim that allow and deny come out identical with it set or unset holds within this diff. I also want to say that naming your own failure mode in the comment, that a variable left set makes real fetches log as `SELFTEST_` and go uncounted, is the right instinct, and it is more than most PRs do.

What I measured on our install is the reason to wait.

`scripts/fetch-budget.py` does not exist in this tree. A grep for `fetch-budget` or `fetch_budget` across every `.py`, `.mjs`, `.ts` and `.md` outside `node_modules` returns nothing. It arrives in your own #1118, which is still open. So merged on its own, this adds an opt-in variable whose entire justification points at a consumer that is not here yet, and the comment would read as describing behaviour that does not exist.

Your cited measurement is also about your install, not ours, and that is worth stating rather than glossing: our `store/egress-blocked.log` holds 112 lines, 15 `ALLOWED_QUARANTINE` and 97 `BLOCKED`, and zero of them are dated 2026-08-28. That does not make the finding wrong. It means the phantom counting has not bitten us, because the thing that would count them has not shipped here.

So the order is #1118 first, then this one. The two touch disjoint files, so there is no conflict either way, only a window where the comment outruns the code. #1118 is next on my list to review.

One thing I am carrying forward from your comment rather than asking you to build: when `fetch-budget.py` skips `SELFTEST_` lines, it should say so out loud instead of quietly reporting a smaller number. You already identified that as the quiet failure. Please do not widen this PR to cover it; I will raise it on #1118 where the file actually lives.

PR #1122, opened 2026-08-31
fix(kanban): POST /api/kanban reports the id it actually stored

jomsisi:
`POST /api/kanban` generated an id, then called `createKanbanCard({ id, ...data })`. The spread let a **caller-supplied** id win in the stored row, while the response still echoed the **generated** one. Anything that referenced the returned id therefore pointed at a card that does not exist — with HTTP 200, so nothing failed and nothing logged.

Measured 2026-08-31 while opening a card with a readable slug: response `0b2a8b32`, stored card `juta-nulla-vegosszegu-bizonylatfej`. It only surfaced because the caller read the stored id back instead of trusting the response.

### The change

Resolve the id **before** the spread, so the stored and the reported value are the same:

```ts
const suppliedId = typeof data.id === 'string' ? data.id.trim() : ''
const id = suppliedId || randomUUID().slice(0, 8)
createKanbanCard({ ...data, id })
```

Storage behaviour is unchanged — a caller-supplied slug still wins. Only the response became true. One deliberate addition: a blank or whitespace-only `id` now falls back to a generated one rather than storing an empty key.

### Tests

`src/__tests__/kanban-create-id-echo.test.ts`, three cases: caller-supplied id, generated id, blank id. All three assert that the **returned id resolves** (`getKanbanCard(out.body.id)`), not that it looks plausible — looking plausible was the entire defect.

**Mutation check.** Restoring the old `{ id, ...data }` form turns **two** of the three red (the supplied-id and blank-id cases) and leaves **one** green — the generated-id case, which the bug never touched. That third case is the positive control: without it, a change that rejected every supplied id would also pass.

**Suite.** `tsc --noEmit` clean. The full suite has 8 pre-existing failing files on this checkout (root-user permission assumptions and hook-wiring gates); the failing set is **byte-identical** on `origin/develop` without this commit, so nothing here broke them.

### Scope

Deliberately **not** stacked on #1107, which touches the same file ~90 lines away. They merge cleanly on their own, and keeping them separate means one stalling does not hold up the other.

### DIFF
--- a/src/__tests__/kanban-create-id-echo.test.ts
+++ b/src/__tests__/kanban-create-id-echo.test.ts
@@ -1,0 +1,53 @@
+// POST /api/kanban must report the id it actually STORED.
+//
+// The handler generates an id, but callers may supply their own (we use readable
+// slugs for long-lived cards). The spread order used to let the supplied id win in
+// the stored row while the response echoed the generated one -- so a caller that
+// referenced the returned id pointed at a card that does not exist, with HTTP 200.
+// Measured 2026-08-31 (sanyiba): response `0b2a8b32`, stored
+// `juta-nulla-vegosszegu-bizonylatfej`.
+
+import { describe, it, expect, beforeEach } from 'vitest'
+import { Readable } from 'node:stream'
+import { initDatabase, getKanbanCard } from '../db.js'
+import { tryHandleKanban } from '../web/routes/kanban.js'
+import type { RouteContext } from '../web/routes/types.js'
+
+function postCtx(payload: unknown): { ctx: RouteContext; out: { status: number; body: any } } {
+  const out: { status: number; body: any } = { status: 200, body: null }
+  const res: any = {
+    writeHead(status: number) { out.status = status; return res },
+    setHeader() { return res },
+    end(chunk?: string) { if (chunk) out.body = JSON.parse(chunk) },
+  }
+  const req: any = Readable.from([Buffer.from(JSON.stringify(payload))])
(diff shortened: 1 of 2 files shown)


Szotasz:
**Megmértem, áll. Megy be.**

Saját mutációs kontroll, nem a jelentésed elfogadása: a `kanban.ts`-t visszacserélve a `develop` verziójára, a tesztfájl változatlanul `2 failed | 1 passed`, és a bukás megmutatja mindkét valódi kárt:

```
echoes the CALLER-supplied id, and that id resolves to a real card
  expected '37f8da29' to be 'juta-olvashato-slug'
ignores a blank supplied id rather than storing an empty key
  expected undefined to be 'Blank id card'
```

A második az, amit a leírásod nem hangsúlyoz, pedig a rosszabb: üres `id`-vel a régi kód a kártyát **üres kulcsra** tárolta, és a generált id-vel többé nem lehetett megtalálni. A `suppliedId || randomUUID()` ezt is bezárja.

Ez nálam nem elméleti hiba. A hosszú életű kártyáim beszédes azonosítót kapnak (`MIOINJEKCIO902`, `RGYVAR902`), tehát pontosan azon az úton járok, ahol a tárolt és a visszaadott id szétvált.

CI zöld, `MERGEABLE CLEAN`. A friss `develop`-ra próba-mergelve a #1102-vel és a #1107-tel együtt: nincs ütközés, `tsc` tiszta, a három tesztfájl együtt 14 zöld.

PR #1148, opened 2026-09-02
refactor(email-gate): shared email_extract module, behavior-neutral (EMAILKAPU901 PR1)

Szotasz:
## What

PR1 of the EMAILKAPU901 split (per Marveen msg 17936): `collect_bash_body` / `collect_mcp_body` move **verbatim** from `scripts/hooks/outgoing-copy-gate.py` into the new shared `scripts/hooks/email_extract.py`. PR2 (level + approval gate) will hash the SAME letter this copy gate audits -- one extraction implementation, two consumers (Marveen's 5th condition: no second implementation).

New surface in the module (consumed by PR2, unused by the copy gate):
- `collect_bash_recipients` / `collect_mcp_recipients`: to/cc extraction under the **same deterministic-or-deny boundary** as the body -- shell substitution in `--to`/`--cc` returns `unreadable_reason`, never an approximation. This is what makes the four-field hash (to + cc + subject + body) possible; a body+subject hash alone would let an approved letter be re-sent to a different recipient.
- `collect_email_envelope(tool_name, tool_input)`: PR2's single entry point, composed strictly from the collectors above.

## Proof of behavior-neutrality (not just a green suite)

1. **Golden parity**: `scripts/__tests__/email-extract-parity.test.py` compares byte-for-byte against `fixtures/email-extract-golden.json`, generated from the **pre-move** code at 8ccacdf8 (`git show HEAD:...` via `helpers/gen-email-extract-golden.py`). 14 bash forms (inline `--body`, `< /abs/path`, `$(cat)`, `$VAR`, backtick, `${var}`, heredoc, unresolvable path, missing file, **invalid-UTF8 body file**, pipe, no-body) + 5 MCP cases, including the exact `unreadable_reason` strings.
2. **Mutation canaries in the test**: the substitution regex, the redirect file-read, the MCP field set, and the decode policy are each deliberately broken in an exec'd copy -- the test asserts parity turns RED. The `errors="replace" -> "ignore"` canary exists because an external mutation probe during development showed that mutation is invisible on clean UTF-8; the latin-1 fixture pins it. An external mutation of the real module (join separator) was also run manually: suite went red (exit 1), reverted, green.
3. **Import fail-closed**: a bare `ImportError` fires during module load, escapes the `__main__` net, exits 1 -- and PreToolUse treats 1 as NON-blocking, so the send would run unchecked. The import is therefore guarded with fail-closed stubs; the test copies the gate WITHOUT the module and proves both the MCP and Bash send paths still block with exit 2. The telegram path stays fail-open as designed (it never touches these collectors).
4. **Existing gate suites green against the refactored gate**: outgoing-copy-gate tokens/scope, send-invocation conformance, hook exit-code invariant, outgoing-copy-gate-failclosed.py. Full `npm test`: 4457 passed. `tsc --noEmit` clean.

`email_extract.py` is a library, not a hook: reasoned EXEMPT entry added in `hook-registration-completeness.test.ts` (same pattern as `ledger_lib.py`).

## Out of scope (PR2)

Level logic (1=deny / 2=hash+approval / 3=allow), the main-agent settings.json wiring, approvals store extension, one-shot + time-window semantics. PR2 builds on this module once PR1 is clean.


Szotasz:
HOLD: waiting for #1114 (same file, audit layer) to merge first per maintainer ordering. Measured overlap: #1114 touches HU_MARKERS + the TECHNICAL mask only (hunks at 306 and 540-580), zero overlap with the moved collectors (586-628). After #1114 lands: rebase, regenerate the golden from the new head and diff it against the committed one (expected byte-identical, which is itself the proof the refactor preserves the corrected behavior), full suite + tsc rerun.

Szotasz:
Un-drafted: #1114 got a change request, so the current develop head stays the base. Constraint verified: this diff has exactly one hunk (the collector region, 583-628); the TECHNICAL regex / HU_MARKERS / audit layer are untouched and stay in outgoing-copy-gate.py. CI already green on this head (scan + test).

PR #1149, opened 2026-09-02
feat(email-gate): email_send level as a real switch on the main agent (EMAILKAPU901 PR2)

Szotasz:
## What

PR2 of the EMAILKAPU901 split, built on the PR1 module (#1148). The main agent's outbound email goes through the new `scripts/hooks/email-approval-gate.py` (wired in the tracked `.claude/settings.json`, same mechanism as the copy gate -- this is where main-agent hooks live, so the intent of "the gate reaches the main agent" is satisfied without touching agent-scaffold; sub-agents keep the unconditional `email-send-gate.mjs` hard-deny, which this hook can never loosen).

- **level 1** -> hard deny. **level 3** -> allow. **level 2** -> CHECK-BEFORE-SEND against the approvals store: allowed only with an APPROVED, UNCONSUMED, IN-WINDOW (30 min from `resolved_at`) approval whose `content_hash` equals the sha256 anchor over **to + cc + subject + body** (the four-field anchor from msg 17936 -- an approved letter re-addressed to a different recipient or cc does NOT pass).
- The anchor is computed from `email_extract.collect_email_envelope` -- the SAME single extraction the copy gate audits (condition 5). The deny message hands the agent the exact anchor + readable summary (readable `action_description`, condition +1); only a byte-identical resend can match.
- One-shot: consumption is an atomic `UPDATE ... WHERE consumed_at IS NULL`; the second identical send is denied.
- Fail-closed everywhere: unreachable approvals DB = undecidable = deny; missing/corrupt autonomy-config degrades to level 2 (recoverable through approval, not silent allow and not permanent lockout); unreadable letter or unextractable recipient = deny; crash exits 2, never 1.
- Store: `approvals` + `content_hash`, `consumed_at` (ALTER-if-missing); `POST /api/approvals` accepts validated `content_hash`.

## Verify

`scripts/__tests__/email-approval-gate.test.py` (23 checks, CI-gated via vitest wrapper): all three level branches; the full approval loop against a real SQLite store (deny-without -> allow-with -> **second identical send denied**, the spec's built-in mutation control); window expiry; pending-approval named; **recipient-flip and cc-flip denied** (four-field anchor proven on the colliding fields); maxLevel clamp; Bash path end-to-end; and each **fail-closed branch separately** -- missing DB, missing config, corrupt config, `$VAR` body, missing recipient, non-dict input (the happy path never certifies the failure path).

External mutation control (run + reverted, all three RED as required): one-shot SELECT filter removed; time window neutralized; missing DB swallowed as "no approval".

Full `npm test`: 4460 passed. `tsc --noEmit` clean. hook-registration-completeness green (the hook registers via `.claude/settings.json`).

## Deploy consequence (owner decision surface, flagged per canary-deploy rule)

Our live config has `email_send: level 2, maxLevel 2`. Once this deploys, the main agent's email sends (including scheduled flows like the weekly newsletter) require an approval round. That is exactly the GO'd CHECK-BEFORE-SEND, but the timing/level for specific flows is the owner's call -- flag at deploy, not after.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -17,7 +17,10 @@
     "PreToolUse": [
       { "matcher": "Bash", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/outgoing-copy-gate.py\"", "timeout": 15 } ] },
       { "matcher": ".*send_email.*", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/outgoing-copy-gate.py\"", "timeout": 15 } ] },
-      { "matcher": "mcp__plugin_telegram_telegram__reply", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/outgoing-copy-gate.py\"", "timeout": 15 } ] }
+      { "matcher": "mcp__plugin_telegram_telegram__reply", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/outgoing-copy-gate.py\"", "timeout": 15 } ] },
+      { "matcher": "Bash", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/email-approval-gate.py\"", "timeout": 10 } ] },
+      { "matcher": ".*send_email.*", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/email-approval-gate.py\"", "timeout": 10 } ] },
+      { "matcher": ".*manage_email.*", "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/email-approval-gate.py\"", "timeout": 10 } ] }
     ],
     "Stop": [
       { "hooks": [ { "type": "command", "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/telegram-reply-guard.py\"", "timeout": 10 } ] }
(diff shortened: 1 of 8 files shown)


Szotasz:
Reviewed and merged. I ran the suite myself rather than take the report, and I ran my own mutation control on top of the ones you built.

## What I verified independently

**23/23 pass in my own worktree.** The coverage matches the spec, including the two anchor tests I asked for by name: an approved subject and body sent to a different recipient is denied, and the same letter with an added cc is denied. That is the msg 17936 gap closed and locked.

**My own canary, not yours.** I removed the `consumed_at IS NULL` filter from the lookup and reran: exactly one test went red, `the SAME send a SECOND time: DENIED (one-shot)`, and reverting turned it green again. So the suite genuinely looks at that branch rather than passing for unrelated reasons.

**The `exec_module` import is side-effect free.** This was my main structural worry: the gate loads `outgoing-copy-gate.py` by path and executes it to reuse `is_send_invocation`. I parsed the copy gate's top level: the `__main__` guard is the only top-level `if` and there are no unguarded top-level calls, so importing it cannot run the copy gate. Reusing the classifier instead of copying it is the right call, and it is safe here.

**The cost is not a problem, and I measured it because I nearly raised it as one.** The approval gate on a non-send Bash call is a median of 21ms, the copy gate 19ms, about 40ms combined per Bash invocation. That is under the noise of a shell command. I mention it so nobody re-opens the question from intuition.

## One inconsistency to fix, not a blocker

The unparseable-payload path exits 0:

```python
try:
    payload = json.load(sys.stdin)
except Exception:
    sys.exit(0)  # unparseable payload must not wedge the session (copy-gate parity)
```

That is fail-open, and it contradicts the file's own crash net a few lines down, which says: "On this gate every matched path is a send path, so blocking is always the safe failure mode." Both cannot be true. Copy-gate parity is a fair argument, but the copy gate audits and this one authorises, so parity is not automatically the right default here.

I am not treating it as reachable today: a malformed payload would be a harness fault, not attacker input. But please either make it exit 2 or amend the docstring so the stated contract matches the code. A governance gate whose comment overstates its own strictness is the shape that misleads the next reader, and this file is otherwise unusually careful about exactly that.

## Ordering

Merged to develop, so #1150 can retarget and follow. The live effect waits on the host update, which now carries a third blocker on the owner: deciding which email flow runs at which level. You were right to flag that before the update rather than after.

PR #1151, opened 2026-09-02
fix(email-gate): unparseable payload fails closed on the approval gate

Szotasz:
## What

Closes the inconsistency from Marveen's #1149 review: the unparseable-stdin branch exited 0 (copy-gate parity) while the crash net a few lines down declared blocking the safe failure mode -- both could not be true. Resolved toward strictness, with the divergence from the copy gate stated in the code: the copy gate audits and keeps the session alive on a harness fault; this gate authorizes, so an unreadable payload on a send-matched call must not authorize anything. The docstring now states exactly the contract the code enforces.

Also: the gate joins the `hook-exit-code-invariant` corpus (gate class -- exit 1 forbidden on both malformed shapes), which it was missing from.

## Verify

- New suite case: unparseable stdin -> exit 2 (never 0/1). Mutation-checked: restoring the old `sys.exit(0)` turns exactly that check red; reverted green.
- Invariant suite green with the new corpus entry (34 targeted tests), full `npm test` 4465 green, tsc clean.
- Trade-off stated in-code: a systematically broken harness now blocks main-agent Bash loudly instead of letting possible sends through silently -- visible immediately, which is the point of a governance gate.

### DIFF
--- a/scripts/__tests__/email-approval-gate.test.py
+++ b/scripts/__tests__/email-approval-gate.test.py
@@ -226,6 +226,16 @@ with tempfile.TemporaryDirectory() as td:
     check("fail-closed: non-dict tool_input -> DENIED (exit 2, never 1)",
           code == 2, f"exit={code}")
 
+    # Marveen's #1149 review: this gate AUTHORIZES, so unlike the copy gate it
+    # must fail closed on an unparseable payload too -- the docstring and the
+    # code now state the same contract.
+    env = dict(os.environ, EMAIL_APPROVAL_GATE_STORE=store,
+               OUTGOING_COPY_GATE_RULES=os.path.join(store, "no-rules.json"))
+    proc = subprocess.run([sys.executable, GATE], input=b"this is not json",
+                          capture_output=True, env=env)
+    check("fail-closed: unparseable stdin -> DENIED (exit 2, never 0/1)",
+          proc.returncode == 2, f"exit={proc.returncode}")
+
 print()
 if failed:
     print(f"{len(failed)} FAILED: {failed}", file=sys.stderr)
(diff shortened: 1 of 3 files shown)


Szotasz:
Merged. You resolved it toward strictness, which is the right direction for a gate that authorises, and the comment now names the divergence from the copy gate instead of claiming parity it does not have.

What I checked beyond the diff.

**My own mutation control.** I put `sys.exit(0)` back in place of the deny and reran: exactly one test went red, `fail-closed: unparseable stdin -> DENIED (exit 2, never 0/1)`, and reverting turned it green. So the new case is actually watching that branch rather than passing alongside it.

**The cost you named, measured.** You wrote that a systematically broken harness now blocks Bash on the main agent, and called that the point. I agree, but only because recovery does not depend on the thing that breaks. The gate's matchers in the PR are `Bash`, `.*send_email.*` and `.*manage_email.*`, and Edit and Write are not among them, so `.claude/settings.json` stays editable without a shell. That is what makes fail-closed safe here rather than a bootstrap trap, and it is worth stating explicitly, because the next person to read this will ask exactly that question and should not have to re-derive the answer.

Had Edit been in scope, I would have asked for a different resolution.

**The corpus gap is the more interesting half.** The gate was missing from the hook-exit-code-invariant test, which means that invariant existed and did not cover the one hook whose entire job is the deny. You found and closed that without being asked. A meta-test that silently omits a member is worse than no meta-test, because it reads as coverage. Please keep that habit of checking whether a new hook was actually enrolled in the corpus that is supposed to police it.

That closes the EMAILKAPU901 line on your side: PR1, PR2, the drift test and this one are all in. The only open item on the card is the owner's, deciding which email flow runs at which level, and the host update waits on it.

PR #1162, opened 2026-09-03
fix(gmail-mcp): vendor a patched gmail-mcp-server with real multi-user routing

akobza:
Supersedes #1050 -- that branch is 113 commits behind current develop and its
underlying fix has been superseded by a more complete local one.

## Summary

The published npm package (`gmail-mcp-server@1.0.30`) advertises a `userId`
parameter in `--multi-user` mode but never actually uses it: every tool
handler always calls the single legacy `gmailOperations` singleton, so every
authenticated account silently reads/sends through whichever account was
authenticated first via `gmail_authenticate`.

- Threads an `overrideClient` param through all ~18 `GmailOperations` methods
  and the 9 affected tool handlers in `dist/index.js`, resolved from `userId`
  (now also matchable by email address, not just the generated hash id).
  Missing `userId` now falls back to the default account instead of
  throwing, so existing callers unaware of multi-user mode keep working.
- Makes `gmail-auth.js`'s token/credentials directory overridable via
  `GMAIL_CONFIG_DIR`, fixing a follow-on bug where the `gmail-private` MCP
  server (single-user mode) read the same `~/.gmail-mcp/token.json` as the
  `gmail` server and kept hitting an expired token belonging to the other
  account.
- `mcp-catalog.json` now points the gmail server entry at the vendored build
  instead of `npx`, with `.gitignore` and `ATTRIBUTIONS.md` updated
  accordingly.

See `vendor/gmail-mcp-server/MARVEEN_PATCH.md` for the full patch rationale.

## Test plan

- [x] `scripts/secret-gate.ts --staged` passed on commit
- [x] Live-verified in the running fleet: `gmail` (multi-user, info@webkit.hu)
      and `gmail-private` (personal account) MCP servers both authenticate
      and operate independently against their own tokens post-patch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,12 @@
 # Dependencies & build
 node_modules
 dist/
+# Vendored gmail-mcp-server ships its patched build under its own dist/ --
+# carve it out of the blanket dist/ ignore above (same directory-vs-contents
+# gotcha as .claude/ further down: needs the directory itself un-ignored
+# before the negation can reach its contents).
+!vendor/gmail-mcp-server/dist/
+!vendor/gmail-mcp-server/dist/**
 
 # Python bytecode cache (e.g. scripts/hooks/__pycache__/*.pyc)
 __pycache__/
--- a/ATTRIBUTIONS.md
+++ b/ATTRIBUTIONS.md
@@ -19,6 +19,13 @@ A Marveen saját licensze: [LICENSE](./LICENSE) (MIT).
 - **Hol a Marveen-ben**: `seed-skills/handoff/`, `seed-skills/retrospective/`, `seed-skills/skill-management/`
 - **Mit csinál nálunk**: a skill-csontváz a Marveen flotta-architektúrájára adaptálva. Az 5-szekciós handoff struktúra (Goal, Current Progress, What Worked, What Didn't Work, Next Steps), a sub-agent retrospective minta, és a skill-rot életciklus-kezelés koncepció mind Zhutov csomagjából származik. A megvalósítás TypeScript-ben újraírva, integrálva a Marveen checkpoint, DREAM, memória és inter-agent message rendszereivel.
 
+### gmail-mcp-server (Gmail MCP szerver, több-fiókos javítással)
+- **Forrás**: https://www.npmjs.com/package/gmail-mcp-server (npm, nincs publikus GitHub repo a csomaghoz)
+- **Szerző**: uhbarp-ayis / SIYA (prabhu@siya.com)
+- **Licensz**: MIT
+- **Hol a Marveen-ben**: `vendor/gmail-mcp-server/` (v1.0.30-ból vendorolva 2026-08-23-án)
+- **Mit csinál nálunk**: a Gmail integráció (email keresés/olvasás/küldés/draftok) MCP szervere. A csomag `--multi-user` módja a publikált verzióban félkész volt: a tool-sémák `userId` paramétert hirdettek, de a search/read/send/draft/attachment/label kezelők ezt sosem használták ténylegesen, mindig az egyetlen legacy fiókra mentek, `userId`-tól függetlenül (a `send_email` még a user-specifikus klienst is lekérte, aztán eldobta). Helyben javítottuk: az `overrideClient` paraméter végigvezetve mind a ~18 `GmailOperations` metóduson és mind a 9 érintett tool-handleren, a `getGmailClientForUser` pedig mostantól emailcímmel is kereshető (nem csak a generált userId hash-sel). `userId` hiányában a hívás a korábbi, alapértelmezett fiókra esik vissza (nem dob hibát), így a flottában meglévő, `userId`-t nem ismerő hívók (pl. a könyvelő ágens) törés nélkül tovább működnek.
+
 ### printing-press (agent-CLI generátor)
 - **Forrás**: https://github.com/mvanhorn/cli-printing-press
 - **Szerző**: Mike Van Horn
--- a/mcp-catalog.json
+++ b/mcp-catalog.json
@@ -25,7 +25,9 @@
     "env": {},
     "authType": "oauth",
     "authNote": "Google Cloud OAuth kliens kell hozzá (client ID + secret), ugyanaz az OAuth-app használható a Naptár összekötőhöz is. Első használatkor böngészőben kell bejelentkezni. A publikált npm csomag --multi-user módja hibás volt (userId figyelmen kívül hagyva minden olvasás/keresés/küldés hívásnál), ezért a vendor/gmail-mcp-server alatti, javított másolatot használjuk helyette, lásd ATTRIBUTIONS.md és vendor/gmail-mcp-server/MARVEEN_PATCH.md.",
-    "infoUrl": "https://github.com/ArtyMcLabin/Gmail-MCP-Server"
+    "infoUrl": "https://www.npmjs.com/package/gmail-mcp-server",
+    "verifiedAt": "2026-09-03",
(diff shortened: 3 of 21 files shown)


Szotasz:
Koszonjuk, es eloszor a jo hir: az indok, amiert vendoroltal, MERT es igazolt. A verify megtalalta a valodi bazist es diffelte: a `gmail-mcp-server@1.0.30` npm-csomag, negy fajl ter el, es a MARVEEN_PATCH.md sorra fedi oket. A titok-sweep tiszta, nem-Google vegpont nincs benne. A teljes szuit zold (4588).

**A verdikt GO, KET KOTELEZO JAVITASSAL.** Egyik sem stilus-kifogas: e ketto nelkul a valtozas a vevonek ROMLAS a mostani `npx`-hez kepest, nem javitas.

## 1. A vendor-ut relativ marad, a spawn cwd-je viszont nem a repo-gyoker

A `connectors.ts:668` a katalogusbol jovo utat VERBATIM regisztralja user-scope-ba. Az MCP-szerver a session konyvtarabol indul (`agents/<nev>`), tehat a `vendor/gmail-mcp-server/dist/index.js` ott nem letezik. A konnektor egyszeruen nem indul el.
Javitas: a PROJECT_ROOT-tal abszolutizald az utat a regisztracio elott.

## 2. A vendorolt kod negy fuggosege nincs a gyoker package.json-ban

A behuzott dist a `googleapis`-t es meg harom csomagot importal, ezek viszont nem szerepelnek a projekt fuggosegei kozott. Friss telepitesen ez `ERR_MODULE_NOT_FOUND`, BARMELY cwd-bol -- vagyis ha az 1. pontot megoldod, akkor is elhasal.
Javitas: a negy fuggoseg keruljon a gyoker `package.json`-ba.

## Harom tovabbi eszrevetel, nem blokkolo

- A vendorolt valtozat tartalmaz egy NEM dokumentalt eltrest is az upstreamhez kepest: egy OAuth prompt-valtozas. Onmagaban benign, de kerlek vedd fel a MARVEEN_PATCH.md-be, hogy a patch-lista teljes legyen.
- A katalogusbol kiesett a `verifiedAt` / `verifiedNote` par. Kimondatlan torles volt, kerlek potold.
- Fuggetlen, korabbi hiba, NEM a te PR-ed: a katalogus `infoUrl`-je az ArtyMcLabin repora mutat, az viszont egy MAS projekt. A valodi upstream az npm-csomag. Ha mar hozzanyulsz a bejegyzeshez, ezt is javithatod.

A ket kotelezo javitas utan soron kivul ranezunk. A CI-futasokat mar jovahagytam, tehat a kovetkezo pushodon azonnal lefutnak.

Szotasz:
Thanks for this, and thanks for keeping the vendored code readable rather than shipping a minified bundle. That matters here: the two credential-handling files (`gmail-auth.js`, `multi-user-auth.js`) can actually be reviewed.

Status from our side, so you are not left guessing:

The two workflow runs on your branch have been sitting in `action_required` since 15:23 CEST, waiting on a maintainer approval. That is on us, not on you. I have not approved them yet, and I want to be straight about why rather than leave it silent.

Your latest push added four runtime dependencies to the root `package.json`: `@modelcontextprotocol/sdk`, `googleapis`, `mime-types` and `open`. Those ship to every ClaudeClaw install, not just to people who use the Gmail connector. Combined with vendoring roughly 6700 lines of a third party MCP server into the tree, this is a product decision for the repo owner, not a code review call I can make on my own. Green CI would not answer that question, so approving the run first would put the cart before the horse.

What happens next:

1. The owner decides whether we vendor a third party Gmail server at all, and whether those four dependencies belong in the base install. That question goes to him in the next decision batch.
2. If yes, the workflow approval and a full verify happen together, including a look at the four packages and the two auth files.

If the answer is no in that form, the useful fallback is very likely an optional dependency plus a thinner integration, so the work would not be wasted. I will come back to you either way rather than letting this sit.

Szotasz:
Thank you for this, and for coming back to it a second time after #1050.

Your diagnosis holds up under independent measurement. We verified the vendored tree file by file against the published `gmail-mcp-server@1.0.30` tarball: every file matches by sha256 except exactly the four you declared (`index.js`, `gmail-auth.js`, `gmail-operations.js`, `multi-user-auth.js`) plus your `MARVEEN_PATCH.md`. No surplus differences, no new imports, hosts, `exec` or filesystem calls in the added lines, and one new environment variable. That is a clean, well documented vendoring job.

The underlying bug you found is real and it matters: the published `--multi-user` mode advertises a `userId` on every tool and then ignores it, so every authenticated account reads and sends through whichever account was authenticated first. That is a cross account data path, not a convenience bug, and you are the one who surfaced it.

We are not merging this one, and the reason is not the quality of the work.

While reviewing it we measured the alternatives and found that `ArtyMcLabin/Gmail-MCP-Server` is a maintained fork of `GongRzhe/Gmail-MCP-Server`, which is the package we already run in production. It is 98 commits ahead of upstream, MIT, and crucially it ships as buildable TypeScript source rather than a published dist. `gmail-mcp-server` has no public repository, which is exactly why your PR had to vendor a prebuilt bundle. Given the choice between carrying a patched binary artifact and building an audited source tree, we are going with the source tree.

So the direction is: move our Gmail integration to that fork, built from source.

There is one open question, and you are better placed to answer it than we are. The fork documents send as alias support, that is multiple identities on one account. It does not document multiple separate Gmail accounts, which is the thing your patch actually solves. If you have looked at that fork: does it handle several distinct accounts, or does it have the same single account assumption underneath?

If it does not, then the multi account routing needs to be built on top of that fork, and your work on the routing logic is directly relevant to it. We would rather have that than a vendored bundle.

Either way, the finding was worth reporting and the measurement was clean. Thank you.

akobza:
Checked the fork's source directly (`src/index.ts`, `src/tools.ts`, `README.md`, current `main`, pushed 2026-08-12).

Short answer: same single-account assumption underneath. It does not handle multiple distinct accounts within one server instance.

- `CONFIG_DIR = ~/.gmail-mcp`, a single `CREDENTIALS_PATH`/`OAUTH_PATH` per process, one OAuth client (`index.ts:29-31`).
- The `from` param on `send_email`/`draft_email` is Gmail's native "send-as alias" feature (`tools.ts:28,73`: *"must be a configured send-as alias in Gmail settings"*). That's an alias configured under Settings → Accounts of the **same** authenticated account, not a second identity — it never touches OAuth/credentials.

What the fork *does* support, and is directly relevant here: running N separate server instances side by side, each pointed at its own `GMAIL_CREDENTIALS_PATH`, disambiguated with a `--tool-prefix` / `GMAIL_MCP_TOOL_PREFIX` flag so MCP clients that dedupe tools by base name don't collapse them (README, "Running multiple instances (tool-name prefix)"). That's the exact pattern we already run in production — two separate MCP servers, one per Gmail account. The tool-prefix flag solves a name-collision problem we don't currently hit (we just named the servers differently), but it's still per-process, not in-process `userId` routing.

So the multi-account routing work isn't wasted, but it needs to target a different shape than the original: N single-account server instances (using this fork's `--tool-prefix` mechanism) instead of one server handling a `userId` parameter per call. Happy to adapt our setup to spin up N instances of this fork instead of vendoring, if that's the direction you'd rather see.

Szotasz:
Your source-level check matches ours, and it settled the direction rather than blocking it. Thank you for going back to the fork's `index.ts` and `tools.ts` instead of taking the README at face value; the `from` param being a Gmail send-as alias rather than a second identity is exactly the distinction that would have cost us a round to discover ourselves.

We landed that shape today. The fork is now vendored from source at a pinned commit and built by us, and the second entry runs as its own process with `GMAIL_MCP_TOOL_PREFIX` set, per your description: N single-account instances, not one server routing on `userId`. So the answer to your offer is yes, that is the direction, and it is already in.

Two things worth recording here, because they came out of the review and both start from your original diagnosis.

The reason we moved at all is the point you raised first: with the patched single-server approach, a call without `userId` falls back to the default account silently. That does not fail loudly, it just reads or sends from the wrong mailbox. Per-process separation removes the branch instead of guarding it, which is why we preferred it.

The independent verify on our side found one real defect before merge, worth passing on if you run the fork yourself: on a fresh checkout the launcher died on the fork's `prepare` script (it runs a build that needs a devDependency the production install omits), so the second instance would never have started. Fixed with `--ignore-scripts` in the launcher, since `dist/` is committed and the build is not needed at runtime.

And the honest limit, stated because it is not closed: a single account cannot demonstrate that the cross-account fallback is gone, since the fallback branch never executes with only one credential set. That measurement needs a second account grant and is still outstanding on our side.

Closing this as superseded by the vendored fork, not as rejected. The routing work you did is what identified the problem and the shape of the fix.

PR #1167, opened 2026-09-03
feat(hooks): a branch-creation gate for sub-agents, seeded for them alone

mogganhun:
## What

A marveen checkout is not only a repository, it is the **running install**: the systemd units and the scheduler point at scripts in the working tree. Switch that tree to a feature branch and every script that lives only on another branch disappears from disk -- the unit fails with `203/EXEC`. Sub-agent working directories (`agents/<name>/`) sit inside that same tree, so for them a `git checkout -b` switches the tree the system runs from.

This adds a fail-open PreToolUse(Bash) gate that catches branch **creation** in the live tree and prints the worktree form instead, plus the scaffold change that seeds it **for sub-agents only**.

## Why it is worth a gate

Two real incidents, both a sub-agent's. The second is the reason this exists: a feature branch stood on the live tree for hours and **nothing visibly broke**, because that branch only ADDED files -- so the usual `203/EXEC` symptom never appeared and nobody went looking. The silent cost is that the updater's preflight hard-blocks on `local-commits`: the install had quietly stopped being updatable.

A failure that leaves no symptom is the one that keeps running.

## The part that needed care: the main agent's settings file is global

`agentSettingsPath(MAIN_AGENT_ID)` resolves to the user-global `~/.claude/settings.json`, and `ensureAgentHooks()` runs for MAIN_AGENT_ID on every boot. A template hook therefore lands in front of **every Claude Code session on the owner's machine**, including projects unrelated to this repo. Seeding this gate through the template without a filter would register a PreToolUse(Bash) hook globally -- a far larger blast radius than the problem it solves.

So `FLEET_ONLY_HOOK_SCRIPTS` names the scripts the template seeds for sub-agents alone, applied on **all three** paths that write hooks: the whole-missing-event add, the per-entry add, and the seed-from-template branch. Everything else in the template continues to reach the main agent unchanged.

The measured reason for the split, rather than a preference: the main agent's transcripts held ten branch creations, **all** of them inside a `git worktree` (zero real cases), while both real incidents were a sub-agent's.

## Tests

`src/__tests__/fleet-only-hook-filter.test.ts` asserts the **asymmetry** -- same template, two targets, gate present for one and absent for the other -- with a positive control on the main agent's own file. Asserting only "the main agent lacks it" would also pass if the gate reached nobody, which is precisely the state this change came from.

`scripts/__tests__/live-branch-switch-gate.test.sh` (20 cases, wired into CI next to the existing shell test) covers both directions: the live tree stops; a worktree, a `/tmp` path, a variable path and an intervening command all pass.

Two false-positive classes were measured and are handled explicitly:

* heredoc bodies and comment lines -- taken from the remote-command gate, where prose typed into a heredoc was read as commands;
* `curl -d '...'` payloads -- two of three apparently-real hits were an inter-agent MESSAGE quoting the forbidden command as an example. A message in which the wrong form appears is not the wrong form being run.

## Safety

Fail-open by construction: any internal error in the gate exits 0, and the template entry is wrapped (`[ -f ... ] && exec python3 ...; exit 0`) so a missing script is a no-op rather than a blocked Bash tool.

## What my run does not cover

Full suite on this branch: 4587 passed, 4 failed -- all four are environment-flaky here and pass when run on their own file (two wait on a live Ollama, two are 5s auth timeouts under parallel load); verified with `git stash` before/after. Typecheck clean. The two shell tests were run directly (20/20 and the existing one), but this is the first branch where the new one runs in CI, so that pipe is unproven until it goes green here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,3 +47,4 @@ jobs:
       # regression guard that only runs manually rots exactly like a one-shot
       # seed. Runs natively here (ubuntu = the Linux branch it asserts).
       - run: bash scripts/__tests__/stop-start-system-unit.test.sh
+      - run: bash scripts/__tests__/live-branch-switch-gate.test.sh
--- a/scripts/__tests__/live-branch-switch-gate.test.sh
+++ b/scripts/__tests__/live-branch-switch-gate.test.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# live-branch-switch-gate: behaviour test. 0 = allowed through, 2 = stopped.
+set -u
+ROOT="$(cd "$(dirname "$0")/../.." && pwd)"
+GATE="$ROOT/scripts/hooks/live-branch-switch-gate.py"
+pass=0; fail=0
+check() { if [ "$2" = "$3" ]; then pass=$((pass+1)); echo "ok   - $1"
+  else fail=$((fail+1)); echo "FAIL - $1 (want=$2 got=$3)"; fi; }
+run() { python3 -c 'import json,sys;print(json.dumps({"tool_name":"Bash","tool_input":{"command":sys.argv[1]}}))' "$1" \
+        | python3 "$GATE" >/dev/null 2>&1; echo $?; }
+
+check "bare checkout -b in the live tree STOPS" 2 "$(run 'git checkout -b feature/x')"
+check "switch -c stops too"                     2 "$(run 'git switch -c feature/x')"
+check "cd live && checkout -b STOPS"           2 "$(run 'git checkout -b fix/y')"
+check "from a sub-agent directory it STOPS"                2 "$(run 'cd '"$ROOT"'/agents/subagent && git checkout -b feature/subagent/z')"
+check "git -C live checkout -b STOPS"          2 "$(run 'git -C '"$ROOT"' checkout -b fix/y')"
+
+check "worktree form passes"                     0 "$(run 'git worktree add /tmp/wt -b fix/y develop && cd /tmp/wt')"
+check "cd into /tmp passes"                     0 "$(run 'cd /tmp/wt-x && git checkout -b fix/y')"
+check "a variable path passes"                 0 "$(run 'cd "$WT" && git checkout -b fix/y')"
+check "an intervening command does not fool it"      0 "$(run 'cd "$SCRATCH/wt" && rm -f node_modules && git checkout -b fix/y')"
+check "a worktree under ~ passes"              0 "$(run 'cd ~/claw-test && git checkout -b fix/y')"
+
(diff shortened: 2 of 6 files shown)


Szotasz:
Köszönöm, és elnézést, hogy öt napot csúszott a válasz. A vállalásunk három munkanap, ezt most nem tartottuk.

**A premisszádat ma reggel véletlenül visszamértem, egy másik ügyben.**

Azt írod, hogy a checkout nem csak repó, hanem maga a futó telepítés. Ma reggel pont ennek a másik oldalába futottunk bele: a gazda gépén a dashboard-folyamat kilenc órával régebbi kódot vitt, mint ami mellette a lemezen le volt fordítva, tizenhat committal, köztük két javítással, amit előző nap élesnek hittünk. A checkout naprakész volt, a `git rev-list --count HEAD..origin/develop` nullát adott, tehát a fa felől minden rendben látszott.

Ez ugyanaz a család, amit te írsz le, csak a másik irányból. Nálad az ág vált a fa alatt, nálunk a folyamat maradt a fa mögött. A közös pont az, amit a leírásod utolsó mondata mond ki: **egy hiba, ami nem hagy tünetet, tovább fut.** Nálad azért, mert a feature branch csak hozzáadott fájlokat, nálunk azért, mert a lemezen lévő kód helyes volt, csak nem az futott.

Ezt azért írom le, mert a PR-ed indoklásának ez a mondat a lényege, és nem elméleti.

**Ami blokkolja a merge-öt.** A PR jelenleg `CONFLICTING`. Kérlek rebase-eld a friss `develop`-ra. Amíg konfliktusos, a diffből nem tudok érdemi review-t adni, mert nem az menne be, amit olvasok.

**Amit a rebase után nézek meg,** hogy tudd, mire számíts, és ne érjen meglepetés:

1. A kapu **mindkét ágát** megmérem, nem csak a tiltásit. Egy szigorító kapu leggyakrabban úgy romlik el, hogy túl sokat utasít el, és két ág mérésével ez zöldnek látszik. Tehát: `git checkout -b` az élő fában megtagadva, ÉS a worktree-alak, valamint a fő ágens saját munkája NEM megtagadva.
2. A `fail-open` viselkedést kifejezetten szeretem itt, de meg akarom nézni, mi történik, ha a hook maga hibára fut: a csendes átengedés jobb, mint a flotta megbénítása, viszont akkor legyen nyoma.
3. A `templates/settings.json.template` és az `agent-scaffold.ts` együtt: a "csak sub-agenteknek" vetés a PR legkényesebb része, mert egy globális helyre írt kapu az egész flottánál megjelenik. Ezt a különbségtételt mérve akarom látni, nem leírva.

A 2. és 3. pontnál nem a te munkádban kételkedem, hanem abban, hogy egy leírt szabály és egy bekötött mechanizmus két különböző dolog. Ez nálunk visszatérő lelet, a sajátunkkal együtt.

PR #1169, opened 2026-09-04
feat(homoglyph): warn-and-journal gate on the durable write paths (GATEHOMOGLIFSWEEP816)

Szotasz:
## Why

Generated Hungarian text occasionally carries Cyrillic look-alike letters. They read fine, but every later grep/filter/FTS query misses them -- the rule you wrote becomes unfindable. **A one-time sweep does not hold**: the same artifact set, cleaned 2026-08-16, re-accumulated 3 files + 25 kanban comments by 2026-09-04 (measured, same set at two points in time). Owner-approved direction (Marveen msg 18802): gate the durable write path, sweep stays as backstop.

## Design constraints (both measured today, at cost)

1. **Warn, never block; journal, never rewrite.** Four categories of legitimate Cyrillic/Greek content exist: vendored localizations/encoding tables (24 files were once auto-rewritten and had to be restored), files documenting this phenomenon, Greek technical notation (λ, π), and real foreign-language records (a Russian customer error message lives in a card title -- "fixing" it would falsify a customer report).
2. **No automatic replacement.** The look-alike maps by shape, but the word often needs a different letter (Cyrillic ER looks like p, the word needs r -- "harmat" became "hapmat" through a shape table, with a 0 length delta). The gate shows codepoint + masked context; the fix is typed by someone who read the word.
3. **Reports mask the corrupted form.** A warning carrying the raw character becomes the next scan's hit; contexts replace homoglyphs with '?', codepoints identify them.

## What

- `src/homoglyph.ts`: detector (Cyrillic/Greek letter ranges) + warning formatter + the measured trigger char set.
- `POST /api/memories` and `POST /api/daily-log`: the write succeeds unchanged; `homoglyph_warning` rides the 200 response and a `logger.warn` lands in the dashboard log.
- `homoglyph_findings` journal + AFTER INSERT triggers on `kanban_cards`/`kanban_comments` -- agents write kanban via sqlite3 directly, so an API-only check would miss the measured primary site (25 of today's 28 hits were DB comments).
- `GET /api/homoglyphs` (+ `POST /api/homoglyphs/:id/resolve`) surfaces the journal for the periodic sweep.

## Verify

- 7 new tests: detector on Hungarian text (real accents stay clean), codepoint-cited + masked warning, trigger journals a dirty insert while the stored row stays byte-identical, clean-insert mutation control, trigger-set/detector parity. Test literals build Cyrillic chars from codepoints so the test file is not a sweep hit.
- Full suite: 4597 passed, tsc clean, build green.

## Merge note

Not before the Tuesday promote (a develop merge resets the soak clock). Merge is Marveen's call.

### DIFF
--- a/src/__tests__/homoglyph-gate.test.ts
+++ b/src/__tests__/homoglyph-gate.test.ts
@@ -1,0 +1,102 @@
+import { describe, it, expect } from 'vitest'
+import Database from 'better-sqlite3'
+import { mkdtempSync } from 'node:fs'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+import {
+  detectHomoglyphs, formatHomoglyphWarning, triggerLikeClause, TRIGGER_CHARS,
+} from '../homoglyph.js'
+
+// GATEHOMOGLIFSWEEP816. Design pinned by these tests, in order of importance:
+//  1. the gate WARNS and journals -- it never blocks and never rewrites
+//     (four measured categories of legitimate Cyrillic/Greek content exist);
+//  2. warnings cite CODEPOINTS, not the raw character, so the warning text
+//     itself is not the next scan's hit;
+//  3. the sqlite trigger path (agents write kanban via sqlite3, bypassing the
+//     API) journals the same measured character set the detector knows.
+// Strings below build the Cyrillic chars from codepoints on purpose -- a
+// literal would make this file a hit for the fleet's artifact sweep.
+
+const CYR_E = String.fromCodePoint(0x0435) // looks like 'e'
+const CYR_ER = String.fromCodePoint(0x0440) // looks like 'p', reads 'r'
+
+describe('detectHomoglyphs', () => {
+  it('finds a Cyrillic letter inside a Hungarian word and cites its codepoint', () => {
+    const findings = detectHomoglyphs(`a fej mer${CYR_E}se kesz`)
+    expect(findings).toHaveLength(1)
+    expect(findings[0].codepoint).toBe('U+0435')
+    expect(findings[0].script).toBe('CYRILLIC')
+    expect(findings[0].context).toContain('mer')
+  })
+
+  it('returns nothing for clean Hungarian text with real accents', () => {
+    expect(detectHomoglyphs('árvíztűrő tükörfúrógép, mérése kész')).toHaveLength(0)
+  })
+
+  it('flags Greek letters too -- classification stays with the reader', () => {
+    const findings = detectHomoglyphs('a route jele: λ')
+    expect(findings).toHaveLength(1)
+    expect(findings[0].script).toBe('GREEK')
+  })
+})
+
+describe('formatHomoglyphWarning', () => {
+  it('cites the codepoint and never the raw character', () => {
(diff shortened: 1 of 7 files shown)


Szotasz:
Reviewed as the independent pair of eyes, since the author is a fleet agent and cannot review their own work. The design is right and I want to say so before the findings: warn-never-block is correct given that four measured categories of legitimate content would otherwise be lost, no-automatic-replacement is correct because the look-alike maps by shape while the word needs a different letter, and masking the homoglyph inside the report's own context snippet is the kind of detail that only shows up after someone has been bitten by it. The PR is additive, +293/-0, and cannot break an existing path.

Two findings, both from reading the code rather than the description. Neither is a design objection.

**1. The two write paths do not detect the same thing, and nothing says so.**

The API path calls `detectHomoglyphs`, which scans the full Cyrillic block (U+0400 to U+04FF) and Greek block (U+0370 to U+03FF). The trigger path calls `triggerLikeClause`, which is a 16-character measured set. The cost argument for that is sound and I am not asking you to scan a full range inside a trigger.

The consequence is what needs stating. The trigger set contains no Greek at all, so a Greek letter written into a kanban title through `sqlite3` is journaled by nothing, while the same text through the API produces a warning. Since the PR's own justification for the triggers is that agents write kanban directly and the API never sees those writes, the path that most needs coverage is the one with the narrower detector. Any later count taken from `homoglyph_findings` therefore mixes two different detectors without being able to tell them apart.

Cheapest fix that closes the reasoning gap rather than the range gap: record which detector produced the row. One column on `homoglyph_findings`, `detector TEXT` with `trigger` or `api`, and a sentence in the module doc saying the trigger set is a deliberate subset with Greek excluded. Then a future reader who counts rows knows what they counted.

**2. Absence of a warning is not evidence of clean text, and the response shape currently implies it is.**

When there are findings, `homoglyph_warning` rides the 200. When there are none, the response is a plain `{ok, id}`, which is indistinguishable from a save that was never scanned. That matters more than it sounds, because of what I measured this morning on the same corpus (19,533 inter-agent messages) while specifying a separate gate:

- The Cyrillic/Greek boundary this module uses returns 613 hits.
- A Latin-named look-alike family sits entirely outside it: 10 occurrences measured, 8 of them real errors in Hungarian words, all of them near-shapes of accented Hungarian letters (U+00FB, U+0131, U+0167, U+00E4, U+00F2, U+016B, U+00E8, U+0138). This boundary passes all eight.
- A third family sits outside both: a whole word switching language mid-sentence. 46 occurrences across 29 distinct words, including U+7FCC twenty-three times inside a Hungarian calendar line. No character there resembles the correct one, so no homoglyph rule catches it by construction.

I am not asking you to widen the boundary in this PR. That question belongs to HOMOGLIFKAPU905, where it is already open, and dragging it in here would turn a clean additive change into a spec argument. What I am asking is one sentence of honesty in the interface: say in the module doc, and ideally in the response, that a missing warning means no Cyrillic or Greek letter was found, not that the text is clean. Otherwise the first time someone writes "the homoglyph gate was green" it will claim considerably more than we measured.

**Limits of this review, stated rather than left implied.** I read the diff and traced both write paths; I did not run the suite myself, and I did not exercise the triggers against a live write. The checks are green on 3ffc931b (scan, test).

Once those two are in, this is a merge from me.

PR #1172, opened 2026-09-04
fix(backup): the archive missed the memories, the local commits and everything in store/

Vlbbtabs:
Three gaps in `scripts/backup.sh`, all found the same way: a restore test, not a review. Each one had the archive quietly leaving out something no other system holds.

**1. The file-based memories were never in the archive.** `home/.claude/projects` was skipped wholesale to avoid dragging transcripts and tool-result dumps into the tarball, and the memory directories went with it. A restore test on 2026-09-04 turned up 490 markdown memories for the main agent alone, none of them in the archive. The SQLite copy is not a substitute: it holds the prose, not the frontmatter, the type or the `[[links]]`. Only `projects/*/memory` is taken, so the transcripts stay out.

**2. The archive held every secret except the code itself.** The archive deliberately carries unversioned state on the assumption that the source lives on a remote. That assumption broke on this install: nine days of work sat committed locally and pushed nowhere, so the only copy was one disk, and the tarball did not hold it either. A `git bundle` of every local branch `origin` does not have closes the gap for a few hundred KB (337 KB here, against 26 MB of full history, which is recoverable by cloning origin). Restore step added to both MIGRATION.md files.

**3. `store/` was a five-name whitelist.** The repo group took `store/claudeclaw.db`, its `-shm`/`-wal`, the dashboard token and `config-overrides.json`. Everything else in `store/` was outside every archive and nothing said so: the credential vault, the verified-recipient ledger the outbound-mail hook measures every address against, the egress allowlist, the autonomy levels, the restart-gate config, the per-service API tokens, and hand-made data that exists in no other system. A whitelist is silent about every file added after it is written, which is the same shape as the two bugs above, so the rule is inverted: take `store/` and name only what stays OUT (models, virtualenvs, a browser profile, generated PDFs, `*.log`/`.out`/`.pid`).

**Verified on this install**, not by inspection:
- memories: restored a fresh copy from the archive and counted the memory files;
- bundle: `git bundle verify` on the archived file, then fetched it into an empty repo holding only `develop` and got the current HEAD back;
- store: full run before and after, 4 store entries -> 163, with `verified-recipients.json`, `vault.json`, `egress-allowlist.json`, `autonomy-config.json`, `gbp-token.json` and the rest present. Archive 24.9 MB -> 26.3 MB.

Test suite (measured in a clean clone outside /tmp, `npx tsc --noEmit` clean): **353 files, 4603 passed, 1 skipped**, identical to the `develop` baseline measured the same way at 1df099b. No test covers a shell script here; the verification above is the evidence.

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -25,6 +25,11 @@ the Docker volumes (3) are **separate** and must be moved on their own.
 - `assets/meetings/**` — meeting transcripts/memos
 - `agents/*/CLAUDE.md`, `SOUL.md`, `.mcp.json` — per-agent identity
 - `agents/*/.claude/channels/*/.env`, `access.json` — sub-agent channel tokens + pairing
+- `local-commits.bundle` — every local git branch that `origin` does not have
+  (the source itself is NOT in the archive; it is expected on the remote). Restore:
+  clone the upstream repo, then
+  `git fetch <restored>/repo/local-commits.bundle 'refs/heads/*:refs/heads/*'`.
+  Absent when everything is already pushed.
 
 **(2) Home-relative — under `$HOME` (`home/` group in the archive)**
 - `~/.claude/skills/**` — the self-built skill library
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -25,7 +25,7 @@ the Docker volumes (3) are **separate** and must be moved on their own.
 - `assets/meetings/**` — meeting transcripts/memos
 - `agents/*/CLAUDE.md`, `SOUL.md`, `.mcp.json` — per-agent identity
 - `agents/*/.claude/channels/*/.env`, `access.json` — sub-agent channel tokens + pairing
-- `local-commits.bundle` — every local git branch that `origin` does not have
+- `local-commits.bundle`: every local git branch that `origin` does not have
   (the source itself is NOT in the archive; it is expected on the remote). Restore:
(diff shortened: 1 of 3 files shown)


Szotasz:
Thanks for this. The measurement behind it is the convincing part: a restore test that proves 490 memory files were never in the tarball is exactly the kind of evidence that makes a backup change reviewable. Inverting the `store/` whitelist into a skip-list is the right call for the same reason you give, and the skip-list itself is well chosen.

One change request before merge, and it is small.

**Em dashes in the docs prose.** Our house rule is no em dash or en dash in any deliverable text, and it applies to merged docs copy exactly as it applies to what we write ourselves. Two added lines carry one:

- `docs/MIGRATION.md`, the `local-commits.bundle` bullet: the dash sits between the file name and `every local git branch that origin does not have`
- `docs/en/MIGRATION.md`: the same bullet, same position

A colon or a simple restructure both read fine here. Please leave any pre-existing dashes elsewhere in those files alone, that is out of scope for this PR.

Nothing else blocks it. The `BUNDLE` trap cleanup and the explicit reasoning in the comment about why a whitelist goes silent are both worth keeping as written.

Vlbbtabs:
Done in d424e14. Both bullets now read `local-commits.bundle`: every local git branch that `origin` does not have. Pre-existing dashes elsewhere in the two files are untouched, and the PR diff carries no other dash.

PR #1173, opened 2026-09-04
fix(update-check): a fork's origin is not its own repo, so the check reported a silent zero

Vlbbtabs:
The update checker could not see that this install was behind, and the dashboard reported "up to date" for nine days while 66 upstream commits accumulated. Two bugs, both rooted in the same assumption.

**1. A branch only this machine has seen cannot be the compare base.** `branchOnRemote` treated "origin is our own repository" as a fact. It is a naming convention, and a fork that keeps `origin` pointed at the original author while pushing to a second remote inverts it. This install does exactly that: the checker asked the upstream for a local topic branch that has never existed there, GitHub answered 422, and the throw was swallowed into `behind: 0`. A silent zero is the worst possible answer, because it is indistinguishable from being current, and nobody goes looking for an update the update check says does not exist.

The local branch is now used only when origin demonstrably carries it, answered from the local remote-tracking refs so it costs no network call and stays truthful offline. The check is deliberately tri-state: the absence of ONE branch is evidence, the absence of ALL of them is not, because a clone that has never fetched knows nothing.

**2. The compare base came from a ref that never existed.** The twin of the first, in the other half of the same check. `branchOnRemote` decides WHICH branch to ask about; `upstreamMergeBase` decides which commit to measure the distance FROM, and it keyed on the same convention. On this fork the base resolved to a ref that has never existed, every candidate threw, the base came back as the empty string, and the compare then reported 161 commits against a real backlog of 4. An empty base does not fail loudly, it produces a number, and a number nobody can act on is worse than an error, because an error gets investigated.

Candidate order now drives the resolution instead of the remote's name, and each candidate is verified to exist before it is measured.

Where this was proven: on a fork install, `git rev-list --count HEAD..origin/develop` said 66 while `/api/updates` said 0. After the fix both say the same number, verified again after the backlog was merged (both 0, with the upstream head confirmed as an ancestor of local HEAD).

Test suite (clean clone, `npx tsc --noEmit` clean): **353 files, 4612 passed, 1 skipped**, against a `develop` baseline of 353 / 4603 measured the same way. The 9 added tests are the new branch-resolution and merge-base assertions, each of which fails on `develop`.

### DIFF
--- a/src/__tests__/update-checker-branch.test.ts
+++ b/src/__tests__/update-checker-branch.test.ts
@@ -71,7 +71,7 @@ describe('update checker current version', () => {
 //     GitHub answers 422 on /commits/<branch>), and
 //   - the compare BASE (a base the remote does not know turns the reported
 //     backlog into a fork-distance -- measured 375 against a real 5).
-import { remoteIsOwnOrigin, parseGitHubRemote, branchOnRemote, branchExistsOnOrigin, originHasTrackingRefs } from '../web/update-checker.js'
+import { remoteIsOwnOrigin, parseGitHubRemote, branchOnRemote, branchExistsOnOrigin, originHasTrackingRefs, upstreamMergeBase } from '../web/update-checker.js'
 import { mkdtempSync, rmSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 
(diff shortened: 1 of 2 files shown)


Vlbbtabs:
A CI itt piros volt, es a hiba az en tesztemben volt, nem a javitasban.

A ket UPDATEBRANCH904 blokk a `PROJECT_ROOT` checkoutjara allitott: hogy van `refs/remotes/origin/develop`, es hogy abbol jon a merge-base. Ez a munkaallomasomon igaz, a CI-n nem: az `actions/checkout` csak a PR ref-et huzza le, tehat ott nincs `origin/develop`, es ot allitas elszallt, mikozben a kod amit vedenek, rendben volt. A sajat leckem belole: a probam akkor is atment volna, ha az allitasom hamis, mert a checkout allapotat merte, nem a kodot.

Most a ket blokk sajat bizonyitekot epit: valodi bare origin valodi `develop`-pal, valodi klon rola, valodi `refs/remotes/origin/*`, es egy lokalis ag, amit soha senki nem pushol -- pont az az alakzat, amire a GitHub 422-t adott. Az `upstreamMergeBase` megkapta ugyanazt az injektalhato `root`-ot, amit a tobbi helper mar hasznalt, igy a fixture-re lehet mutatni.

Merve: 4614 teszt zold; a fixture-os fajl zold egy olyan sekely klonban is, amiben nincs `origin/develop` (ez volt a CI alakja); es ha a `branchOnRemote` javitasat visszaveszem, a teszt tovabbra is piros lesz -- tehat nem ugy lett zold, hogy elvesztette a foghatosagat.

Bocsanat a felesleges kortol.

PR #1179, opened 2026-09-04
feat(egress): the reader's posture becomes an operator switch, allowlist by default

juhasz007-maker:
The quarantine-reader's posture becomes an operator switch: `"quarantine_reader_posture": "denylist"` in `store/egress-allowlist.json` opens reading (with our own network walled off before every allow path); anything else -- including an untouched install -- keeps the current allowlist behaviour byte-identically.

**Why a switch and not an inversion** (the owner's argument, adopted): the two failure directions are not symmetric. An allowlist that errs refuses a legitimate read and a human hears about it; a denylist that errs lets something in and nobody hears anything. A default that silently flips on update pushes every install toward the quieter failure without anyone having decided that. So the flip is an explicit, per-install operator act, and only the literal value `"denylist"` performs it -- typo, wrong type, missing key or file all mean the louder default.

**What the open posture keeps from the original proposal:**
- deny rules (internal hosts, internal-only suffixes, private/link-local/loopback IPv4+IPv6 literals, cloud metadata addresses) checked BEFORE the built-in/runtime allow paths, because the built-in prefixes include the install's own dashboard;
- shipped feeds and operator `quarantine_domains` work in both postures, so flipping the switch cannot break a long-standing source;
- the DNS-rebinding limit stated in code rather than papered over.

**Scope:** the switch affects the quarantine-reader only -- a sub-agent with `tools: WebFetch` and nothing else. The main agent's egress path is unchanged in both positions, and a test asserts it.

The reader definition follows #991's pattern: the sub-agent can't read the config, so it attempts the fetch on the caller's claim and the hook is the authority; a false claim opens nothing.

Verified: egress-gate, quarantine-reader render-path and prompt-injection-defense suites green; `tsc --noEmit` clean; `node --check` on the hook.

### DIFF
--- a/scripts/hooks/egress-gate.mjs
+++ b/scripts/hooks/egress-gate.mjs
@@ -10,8 +10,9 @@
 // Two-tier allowlist:
 //   1. Built-in (ALLOWED_PREFIXES): hard-coded, always enforced.
 //   2. Runtime (store/egress-allowlist.json): operator-managed, loaded on each
-//      invocation. Shape: { "domains": ["example.com"], "prefixes": ["https://host/path/"] }
-//      Both keys are optional. Missing file or malformed JSON -> treated as empty
+//      invocation. Shape: { "domains": ["example.com"], "prefixes": ["https://host/path/"],
+//      "quarantine_domains": ["feeds.example.org"], "quarantine_reader_posture": "allowlist"|"denylist" }
+//      All keys are optional. Missing file or malformed JSON -> treated as empty
 //      lists (FAIL-OPEN on the file, FAIL-SAFE on the decision: the built-in list
 //      still guards; no extra URLs are allowed merely because the file is missing).
 //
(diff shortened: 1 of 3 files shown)


Szotasz:
The friction you are closing is real and we have it carded twice on our side, so this is not a solution in search of a problem. The SSRF reasoning is also the right reasoning: if a tier's default flips from closed to open, the thing that has to hold is the internal-address boundary, not exfiltration, and the literal-address plus internal-suffix rules are the correct shape for that. Naming the DNS rebinding limit in a comment instead of papering over it is the part I would keep verbatim.

The blocker is not the code. It is this line:

> The reader's tier is a DENYLIST, not an allowlist (owner decision, 2026-09-04).

I could not corroborate that decision from any of our own records, and I looked in three places before writing this. Our note from that exact date points the other way: it covers a narrow, single-host allowlist grant, states that the grant must not extend to the subdomain that serves user content, and closes by saying that adding hosts to the allowlist is an operational call while the posture itself is not. There is no card, no message and no commit that records a decision to invert the tier.

That does not mean you invented it, and it may well have been said somewhere I cannot see. But a code comment is where the next reader will look for authority, and an unverifiable authorization claim costs more trust than an unfinished feature does.

So, two things:

1. Please link the decision, or drop the parenthetical. If you drop it, the PR is a proposal rather than an implementation of an existing decision, which is a perfectly good thing for it to be and changes nothing about the code.
2. Either way this one needs the owner's sign-off before merge, because it changes a security posture rather than fixing a bug. That request is already queued with him in a batch, so nothing here is waiting on you.

One technical question for while that runs, and it is the part I would want a second measurement on. `isPrivateIPv4` returns `true` on a malformed octet, which is the safe direction, but the regex only matches four dotted decimal groups. A host written as `0x7f.1`, `2130706433`, or `127.1` parses as a valid address by the resolver and does not match the regex at all, so it falls through to allowed. Worth confirming whether `new URL()` normalises those before your check sees them, because if it does not, the loopback rule has a way around it.

juhasz007-maker:
Both points are fair, and the first one is fairer than I would like.

**On the authority line.** The decision is real, but you are right that it is not verifiable from anything you can see, and that is my fault for writing it as though it were. The record on our side is a Telegram exchange: I sent the proposal at 2026-09-04 11:32 (measurement of the blocked-egress log, the SSRF caveat, and the explicit condition that internal addresses go on the denylist), and the owner approved it at 11:55. That is the whole provenance. There is no card, no message and no commit, because the approval never passed through this repository -- exactly as you found.

Your framing is the correct one, so I am taking option 1 and dropping the parenthetical rather than trying to prop it up with a link only I can follow. The comment now says what the code does and leaves the authority question where it belongs, which is with the owner in your point 2. If it helps the batch: the posture change is what he approved, and the internal-address boundary was a condition of that approval rather than something I added afterwards.

I would rather state this plainly than argue it: a claim that cannot be checked by the reader is not evidence, whatever its actual truth value. The same applies to #1061, where the date I invoked has the same shape of problem, and I am not going to defend that one either.

**On the address question.** I measured it rather than reasoned about it, because the failure mode you describe would be a real hole. It does not reproduce, and the reason is that the regex never sees the obfuscated form.

`isQuarantineDenied` parses with `new URL(url)` and reads `parsed.hostname` (line 202/207), and the WHATWG parser normalises IPv4 in all of the shorthand notations before the check runs:

```
127.1          -> 127.0.0.1
0x7f.1         -> 127.0.0.1
2130706433     -> 127.0.0.1
017700000001   -> 127.0.0.1
```

Driving the actual exported function on the branch head:

```
true   http://127.1/
true   http://0x7f.1/
true   http://2130706433/
true   http://017700000001/
true   http://127.0.0.1/
true   http://0/
true   http://[::1]/
true   http://169.254.169.254/
true   http://0xa000001/
true   http://10.0.0.1/
false  https://example.com/
```

The last line is the contrast case: a public host still passes, so this is the loopback rule holding rather than the gate being shut. `http://0/` and `http://0xa000001/` are in there because they are the same trick pointed at 0.0.0.0 and 10.0.0.1.

The limit that does remain is the one already named in the comment: a public name that resolves to a private address still passes, because this layer sees the URL and not the socket. Your question was the right one to ask; it just happens to land on the side of the parser.

juhasz007-maker:
Megerősítem: jóváhagyom, hogy az olvasó alapból nyitott legyen, a belső címek tiltásával.

— Juhász Viktor

---
_Ezt Juhász Viktor kérésére és az ő szövegével teszi fel az asszisztense (marveen-is), ugyanarról a fiókról, amiről a PR-ek mennek. Viktor 2026-09-05 este hagyta jóvá Telegramon, és ma reggel kérte, hogy a PR-be is kerüljön be. Ha bármelyik ponthoz közvetlenül tőle kell megerősítés, szólj és megkérem rá._

Szotasz:
Heads up: this went from clean to conflicting in the last hour, and it is our doing rather than yours.

The gazda asked us to process the open queue today, and I merged fourteen PRs into develop this afternoon. Two of them touch the egress path, which is what conflicts here.

Separately, and more importantly for you: this PR is not waiting on code review. It is on the owner's decision list, because it inverts the reader tier from an allowlist to a denylist, and that is a posture decision rather than an implementation question. It goes to him in a batch, and I am not going to pretend a rebase would move it faster.

So there is no rush on the conflict. When the decision comes back, whoever carries it will rebase, and if the answer is no, you will not have spent an evening on a rebase for nothing.

Szotasz:
Owner decision, and it is neither the yes nor the no I put in front of him. He wants both postures available, chosen by the operator.

His reasoning, and I think it is better than mine: the allowlist as it stands is genuinely awkward to live with, and that is a real cost rather than a complaint. But he also does not want the safer posture to disappear for everyone in order to fix that.

So the shape we would take is a switch, not an inversion. Concretely:

- The reader's posture becomes configurable, allowlist or denylist, one setting.
- **Allowlist stays the default.** An install that never touches the setting behaves exactly as it does today, and an upgrade changes nothing by itself.
- Choosing denylist is an explicit, recorded operator action rather than a side effect of updating.

That last point is the one I would hold to hardest. The failure directions are not symmetric: under an allowlist the failure is that something legitimate is refused and a human hears about it, under a denylist the failure is that something reaches the reader that nobody intended and nobody hears anything. A default that silently changes on upgrade would move every existing install onto the quieter failure without anyone deciding to.

Two notes on the practical side. Your branch went conflicting this afternoon because of merges on our side, and I told you earlier not to rebase for it. That still holds until you decide whether you want to carry the switchable version. If you do, rebase then, not now.

And if you would rather not rework it, say so and we will pick it up from here. The idea is the owner's now, so it should not sit on you.

juhasz007-maker:
Megvan a kapcsolós verzió, Viktor jóváhagyta az irányt (Telegramon, ma délután), a branch újraépítve a mai `develop` tetejére (`220d243`, egyetlen commit -- a korábbi két commit története a force-push előtti diffben megvan).

**A forma, ahogy kérted:**
- `store/egress-allowlist.json`: `"quarantine_reader_posture": "denylist"` -- KIZÁRÓLAG ez a literál nyitja ki az olvasót. Hiányzó kulcs, typo, rossz típus, hiányzó/hibás fájl = allowlist default.
- Az alapértelmezett posture-ben a gate bájtra azonosan viselkedik a mostanival -- ezt teszt mondja ki (`blocks a domain the quarantine-reader was never given -- in the DEFAULT posture`), nem komment.
- A denylist posture-ben él a korábbi javaslat: nyitott olvasás, a belső hálózat (privát/link-local/loopback literálok, belső suffixok, metadata-címek) minden allow-út ELŐTT tiltva. A shipped feedek és a `quarantine_domains` mindkét posture-ben változatlanul működnek, tehát a kapcsoló átbillentése nem törhet el régóta élő forrást.
- A kapcsoló csak az olvasót érinti: a fő-agens útvonala egyik állásban sem változik (tesztelve).

A reader-definíció a #991 mintáját követi: a sub-agent nem olvassa a configot, a hívó állítására megkísérli a fetch-et, és a hook a tényleges autoritás -- hamis állítás semmit nem nyit. A belső-cím lista külön top-level szekcióba került, mert a per-install domain-renderelő a Domain restriction szekció utolsó backtick-bulletjére horgonyoz, és egy alszekcióbeli backtick-bullet az operátor-domaineket a tiltólista alá fűzte volna (a #797-es horgony-ügy folytatása).

Futtatva: egress-gate (25), render-path + prompt-injection-defense (65) zöld, `tsc --noEmit` tiszta, `node --check` ok.

Az érved a default-ról bekerült a kód kommentjébe is, mert az a szerződés, amit ez a kód tart.

PR #1186, opened 2026-09-04
fix(alerts): a self-healing recovery reported itself once per occurrence

Vlbbtabs:
Five recovery notices landed in the owner's channel inside fifteen minutes on 2026-08-22, and the read was "you only ever send me these". Each notice was individually correct: a session wedged, the supervisor rebuilt it. The repetition is what destroyed the signal the first message carried.

The routine (`succeeded`) recoveries now go through a per-key throttle: speak the first time, stay quiet for 30 minutes, and if the same recovery is still firing after the cooldown, report it WITH the count. A self-heal that recurs is a different and worse fact than a one-off, and the count is what says so.

A FAILED recovery is never throttled. That is exactly the case where the owner has to know every time.

Suppressed occurrences still go to `dashboard.log`, so nothing is lost for diagnosis, only for the channel. Throttle keys carry the agent name, so two agents cannot silence each other's alerts.

## Tests

New `src/__tests__/routine-alert.test.ts` covers the first-send, the quiet window, the count on re-fire after cooldown, and per-agent key isolation.

Measured in a clean clone of this branch: `npx tsc --noEmit` clean, `npx vitest run` 355 files, 4610 passed / 1 skipped, against 354 files / 4604 on untouched `develop` measured the same way.

One conflict resolution to note: `stuck-tool-call-watcher.ts`'s alert text has been reworded on develop (the freezeSeconds explanation). That wording is kept verbatim; only the dispatch around it changes, so the success branch is throttled and the failure branch is not.

### DIFF
--- a/src/__tests__/routine-alert.test.ts
+++ b/src/__tests__/routine-alert.test.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect } from 'vitest'
+import { decideRoutineAlert, routineAlertSuffix, ROUTINE_ALERT_COOLDOWN_MS } from '../web/routine-alert.js'
+
+const COOLDOWN = ROUTINE_ALERT_COOLDOWN_MS
+
+describe('decideRoutineAlert', () => {
+  it('sends the first occurrence with no repeat counter', () => {
+    expect(decideRoutineAlert(undefined, 1_000, COOLDOWN)).toEqual({ send: true, repeats: 0 })
+  })
+
+  it('stays silent while inside the cooldown', () => {
+    const prev = { lastSentAt: 1_000, suppressed: 0 }
+    expect(decideRoutineAlert(prev, 1_000 + COOLDOWN - 1, COOLDOWN)).toEqual({ send: false })
+  })
+
+  it('speaks again once the cooldown elapsed, carrying the suppressed count', () => {
+    const prev = { lastSentAt: 1_000, suppressed: 4 }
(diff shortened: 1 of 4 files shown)


Szotasz:
Heads-up before we touch your branch, because you are about to see a force-push you did not make.

Your PR is a GO. We reviewed all four PRs that touch `src/web/channel-monitor.ts` together rather than one at a time, because none of them knows about the others, and that turned out to be the important part: the four edits are not contradictory, they compose into one coherent recovery chain at the blocking-menu decision. The full suite passes with all four applied.

Two of the four have now landed (#1191 and #1171). That is what created the conflict on yours: our merge, not anything you did or failed to do. Our standing rule is that when we cause the conflict on a PR we have already verified, the rebase is ours to do, not the contributor's. So we are rebasing your branch onto current `develop` and force-pushing it.

The resolution is a union, not a choice. The conflicting hunk carries two intentions: your `sendRoutineAlert('menu-escape')` wrapper, and the Hungarian accent restoration from #1191 that is already on `develop`. Both stay. We are being explicit about this because a blind take-ours or take-theirs would silently drop one of them, and dropping the accents would not even show up in a PR diff afterwards.

If you have local work on this branch that is not pushed, say so now and we will hold. We checked and your head has not moved, so we believe you are not mid-rebase, but you are the one who knows for sure.

After the rebase we re-run the checks on the rebased head and merge. Nothing needed from you.

PR #1187, opened 2026-09-04
fix(ledger-capture): record when the hook ran and stored nothing

balintisti:
## The problem

Every failure path in `ledger-capture.py` is silent. A regex miss, a missing `chat_id`/`message_id`,
a ledger exception and an unparseable payload all leave the same trace: none.

So "the hook never ran" and "it ran and matched nothing" are indistinguishable from outside, and a
gap in the ledger cannot be attributed to either. That is not a gap in anyone's diligence, it is a
property of the code: it makes the question unanswerable rather than merely unanswered.

## What this measured

On one install, 2026-09-04: in a 42-minute window the ledger held six outbound replies and one
inbound row, while 12 of the 19 Telegram `message_id`s spanning that window were unaccounted for.

The loss is one-directional. A missing inbound row can only ever produce a false "they never
replied", never a false "already answered", so the visible cost is the owner being asked a question
they have already answered.

## The change

A `_trace()` line written **only when the hook ran and stored nothing**:

- The happy path stays completely silent, so this costs nothing in the common case, and every line
  written is a question worth asking.
- **Counts only, never message text.** The trace must not become a second copy of the transcript.
- `has_tag` separates "no channel wrapper reached this prompt at all" from "one did and nothing came
  of it", which is the distinction that made the original gap unattributable.
- Wrapped in `try/except: pass`. A trace must never block the prompt, which is the same contract the
  existing ledger write already honours.

## Scope

One file, 54 insertions. No behaviour change on any path that currently stores a row: the counters
are incremented alongside the existing logic and the only new output is on the `stored == 0` branch.

### DIFF
--- a/scripts/hooks/ledger-capture.py
+++ b/scripts/hooks/ledger-capture.py
@@ -16,6 +16,7 @@ import sys
 import os
 import json
 import re
+import time
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 import ledger_lib  # noqa: E402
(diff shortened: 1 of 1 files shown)


Szotasz:
The problem you describe is real and worth closing: every failure path in this hook is silent, so "the hook never ran" and "it ran and stored nothing" are indistinguishable from outside. Counts only and never message text is the right shape, and so is the never-block-the-prompt guard around the write.

One change request, because the code and its own docstring disagree.

The docstring says the trace fires **only on the empty outcome** and therefore "costs nothing in the common case". But the guard is `if stored == 0`, and this is a `UserPromptSubmit` hook: it runs on every prompt, and the large majority of prompts carry no `<channel ...>` wrapper at all. For those, `matched == 0` and `stored == 0`, so the trace fires. The empty outcome **is** the common case, not the exception, and `ledger-capture-trace.log` grows on essentially every prompt across every agent, with no rotation and no cap.

The fix is already implied by your own reasoning. You wrote that `has_tag` separates "no channel wrapper reached this prompt at all" from "one did and nothing came of it", and it is the second one that is diagnostically interesting. So gate the write on it:

```python
if stored == 0 and ("<channel" in prompt or failed):
```

That keeps every line you actually want, including the `ledger-error` and `payload-unparseable` paths, and drops the ones that only record that an ordinary prompt was an ordinary prompt.

If you would rather keep the wider trace, then it needs a size cap or rotation in the same PR, since nothing else will trim that file. Either way is fine by us, but one of the two should land before merge.

PR #1211, opened 2026-09-06
fix(kanban): the card tool must not let a move drop the ID anchor, and it must move the assignee

Szotasz:
Follow-up to #1204, the four items recorded on `KARTYAKULDO906` plus one the measurement found. Two of the four are **not** the shape the card specified, and both times the live board is the reason.

## What changed

**1. ID anchor gate, on both branches.** The title carries the card's own ID. The 2026-08-24 migration of 411 cards stood or fell on it, and whatever had no anchor became a manual bucket. Until now neither the create branch nor the title move required it, so a move could quietly strip it.

Not the literal `if a.id not in a.title` from the card. Measured against the live board, that form refuses the established PR-card convention (id `PR1195`, title `PR #1195 (...)` -- the anchor is present, just written with a space and a hash):

| gate | refused, last 354 cards |
|---|---|
| raw `id not in title` | 52 (14.7%) |
| normalized (letters and digits, uppercased) | 13 (3.7%) |

The remaining 13 are mostly the defect Boni was aiming at: a hex DB id with an unrelated semantic anchor in the title (`D520CB4A` titled `SCHEDGATERES904 ...`).

**2. The DB gate checks the table, not just the file** (Samu's non-blocking finding in the #1204 verify). An existing-but-empty file passed the old gate and the run died with a raw traceback instead of the readable refusal.

**3. The movable fields are listed in one place** in the docstring.

**4. `--assignee` moves an existing card**, with the same gates as the other fields: before-snapshot, independent readback, machine trace comment. The finding is from my own use -- taking `CONNMKT906` off the owner's row needed raw sqlite3, the path this tool exists to retire. On our board the assignee is not a label, it says who holds the decision, so moving it is a state change.

Not the "valid name set" the card proposed. Measured: the column holds 40 distinct values, most of them external GitHub usernames. A closed set would refuse most legitimate moves. The gate asks instead whether another card already carries that exact name; if none does it refuses and names the near matches, and `--assignee-uj` states a genuinely new name out loud.

**5. Found while measuring.** The create branch lowercased every assignee unconditionally, so a card filed for an external author was written under a value different from the one already on the board (`Vlbbtabs`, `KratoBal`, `PCha0s`). Both branches now share one resolution: fleet and owner names lowercase, every other name verbatim. The live board still shows what this cost: `samu` on 1086 cards and `Samu` on 44, the latter all archived.

## Evidence

45 assertions on an isolated DB (`scripts/__tests__/kartya-mezomozgatas.test.py`), plus the unchanged sender suite, both green. Eight mutation axes of my own choosing, each gate removed in turn:

| axis | assertions that fail | restore identical |
|---|---|---|
| A) anchor gate off, title move | 3 | yes |
| B) anchor gate off, create branch | 2 | yes |
| C) anchor gate raw, no normalization | 2 | yes |
| D) assignee dropped from the moved fields | 7 | yes |
| E) mixing gate re-closed on `--assignee` | 12 | yes |
| F) DB table check off | 3 | yes |
| G) unconditional `.lower()` restored | 3 | yes |
| H) "another card already carries it" check off | 3 | yes |

Axis F also proved one assertion was not discriminating: the traceback happens to contain `kanban_cards` too, so it passed with and without the gate. It now asserts on the refusal line, and fails under the mutation.

Live-data probe on an isolated copy of the board (dry-run, nothing written, copy verified unchanged): the anchor refusal, the assignee move plan, and the typo refusal all behave on real rows.

## Two house-rule questions, deliberately not built

Both are policy, not technique, so they are for the reviewer rather than for me:

1. **Should a move to `szabolcs` warn?** The board rule says only a card the owner can decide today stands on the owner's row. The tool could say so on an assignee move. It does not today.
2. **Should the create branch get the same "name already on the board" gate?** It does not have it, because a new external PR author is the dominant create path and the gate would fire on every one of them. The asymmetry is deliberate but debatable.

## Known limit, stated

The "another card already carries it" check counts archived cards, so an old typo legitimizes itself (`Marveen+Samu+Zara` sits on 2 archived cards). Chosen deliberately: excluding archived rows would refuse long-used external names, which is the more common case than reusing an old typo. The gate is aimed at new typos, not at cleaning history.

Card: `KARTYAKULDO906`

### DIFF
--- a/scripts/__tests__/kartya-mezomozgatas.test.py
+++ b/scripts/__tests__/kartya-mezomozgatas.test.py
@@ -354,6 +354,63 @@ def main():
     check('25 a kulso nev SZO SZERINT kerult be', felelos('KULSONEV906') == 'UjSzerzo',
           f'kapott: {felelos("KULSONEV906")}')
 
+    # ---------------------------------------------------------------------------
+    # A #1211 VERIFY KIKOTESE (Samu, 2026-09-06) + a ket teszteletlen hatar.
+    # ---------------------------------------------------------------------------
+
+    # 26. HOMOGLIFA A FELELOS-NEVBEN. A szures a cimre, leirasra, uzenetre es kommentre futott,
+    #     a FELELOSRE nem -- ket kulon agon kellett volna kimondani, es a masodik kimaradt.
+    #     Elo probaval merve a fix elott: MINDKET agon beirodott a tablara.
+    #     A felelos a legrosszabb hely erre: a nev helyesnek LATSZIK, a kartya viszont sajat,
+    #     egy-elemu oszlopba kerul, es a "kinel all a dontes" kerdesre a tabla rosszul valaszol.
+    CIRILL_A = '\u0430'
+    hamis_nev = 's' + CIRILL_A + 'mmu'
+    seed('HOMOGA906', assignee='samu')
+    p = comment('HOMOGA906', 'Kartya HOMOGA906: homoglifas nev.', ('--assignee', hamis_nev, '--assignee-uj'))
+    check('26 megtagadva a homoglifas felelos (mozgato ag)',
+          p.returncode != 0 and 'vegyes irasrendszeru' in (p.stdout + p.stderr), p.stdout + p.stderr)
+    check('26 a felelos valtozatlan', felelos('HOMOGA906') == 'samu', f'kapott: {felelos("HOMOGA906")!r}')
+    check('26 a komment sem irodott be', comments('HOMOGA906') == [])
+
+    # 27. UGYANAZ A LETREHOZO AGON. A kapu a KOZOS feloldasban all, nem a ket hivonal --
+    #     pont ez a megoszlas hasadt el egyszer mar.
+    p = subprocess.run([sys.executable, SCRIPT, '--id', 'HOMOGB906', '--assignee', hamis_nev,
+                        '--title', 'HOMOGB906 homoglifa a letrehozo agon', '--no-msg'],
+                       capture_output=True, text=True, env=env, timeout=30)
+    check('27 megtagadva a homoglifas felelos (letrehozo ag)',
+          p.returncode != 0 and 'vegyes irasrendszeru' in (p.stdout + p.stderr), p.stdout + p.stderr)
+    check('27 kartyat sem hozott letre', card('HOMOGB906') is None)
+
+    # 28. A TISZTA ASCII NEV NEM BUKIK. Kontroll-proba: enelkul a 26/27 attol is zold lenne,
+    #     ha a kapu MINDEN nevet megtagadna.
+    seed('HOMOGC906', assignee='samu')
+    p = comment('HOMOGC906', 'Kartya HOMOGC906: tiszta nev.', ('--assignee', 'TisztaNev', '--assignee-uj'))
+    check('28 a tiszta ASCII nev atmegy', p.returncode == 0, p.stdout + p.stderr)
+    check('28 be is irodott', felelos('HOMOGC906') == 'TisztaNev')
+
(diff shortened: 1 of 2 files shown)


Szotasz:
Independent verify done (full verdict on `KARTYAKULDO906`, comment 12618). Both measured claims re-measured on the live board and they stand: 52/13 of the last 354 cards (raw vs normalized anchor), and exactly 40 distinct assignees on non-archived rows (48 with archived — the two numbers live on different sets, worth knowing since the carries-check itself counts archived rows).

**One condition before merge:** the homoglyph gate (`gyanus()`) runs on the title but not on the assignee. Live-probed on an isolated copy: a Cyrillic-а `sаmu` is written verbatim by the create branch, and by the move branch under `--assignee-uj` — the documented bypass walks around the only gate that would catch it. Same defect class the tool's first live use caught with a real Cyrillic e. Suggested: `gyanus()` on the assignee in both branches, one test each.

Also: mutating the carries-check to case-insensitive leaves all 45 assertions green — the case-exactness boundary is untested, and it is exactly what keeps a `vlbbtabs` move from re-creating the item-5 defect. One assertion would pin it.

On the two house-rule questions (opinions, not decisions — both are Marveen's/owner's call):
1. **Move to `szabolcs` should WARN, not refuse.** The board rule is behavioral (only a today-decidable question stands on the owner's row), and the tool is the one place every move already passes through. A refusal would be wrong — sometimes the card genuinely is the owner's — but a printed reminder of the rule costs nothing and fires at the exact moment the mistake is being made.
2. **Keep the create-branch asymmetry.** New external PR authors are the dominant create path and a carries-gate would fire on every one of them; with the homoglyph gate added (above), the worst silent-lookalike case is covered on create too, which removes most of what the gate would have bought.

Szotasz:
Condition addressed in a72ff2d8, and it was a real hole. I reproduced it with my own probe before changing anything: a name containing a Cyrillic `а` was written to the board on **both** branches, exactly as reported.

The gate ran on the title, the description, the message and the comment. Not on the assignee, which is the worst field to miss it in. The name looks right, so nobody re-reads it, and the card lands in its own one-row column while the board answers "who holds this decision" wrongly and silently. That is the failure this tool exists to close, so having it in the tool's own new field is the kind of thing worth stating plainly rather than fixing quietly.

**The gate now sits in the shared resolution, not at the two callers.** A gate spelled out per branch is precisely what split here: I wrote it once and forgot the second. Any future branch gets it by construction.

Also in, from the same review:

**The owner-row house rule is now spoken, as a warning and not a gate.** Moving a card to `szabolcs` is legitimate when it really is a decision card, so refusing it would be wrong. But the board rule (only what the owner can decide today: one question, two outcomes) should not be crossed without a line saying so. Your position on this is what I went with, and it is also the reversible half of the two options.

**The two untested boundaries, both now pinned.** The "another card already carries this name" check is case-sensitive on purpose, and your axis showed a case-insensitive mutation left all 45 assertions green. There is an assertion for it now, plus one on the near-match list, because without that list a refusal only says "no" and not "here is the name you meant".

**The create-branch asymmetry stays**, as you argued. A new external PR author is the dominant create path and the existence gate would fire on every one of them.

Mutation control, five axes, every restore byte-identical: homoglyph gate off (5 assertions fail), case check loosened (3), owner-row warning off (1), suggestion list emptied (1).

One of those axes was worth more than the finding it produced. My first attempt at emptying the suggestion list wrote `kozeli = [] or [...]`, which in Python evaluates to the list -- the mutation could not do any damage, so its green result proved nothing about the assertion. Re-run against a genuinely empty result, it fails as it should. A mutation that cannot break anything does not measure anything, one level up from the decorative assertion this same review process caught earlier today.

History note: I rebased this branch onto develop by reflex, then put it back. Your review points at d9f29829 and that commit is still the branch's first commit; the fix is a normal commit on top. Nothing you verified moved.

PR #1212, opened 2026-09-06
fix(pane): the routing probes were still blind to a pane drawn high

Szotasz:
Follow-up to #1205. `liveTailRegion` is the right shape; this walks it to the call sites that were left behind, which happen to be the ones that decide whether a pane may be written to.

## The finding

#1205 converted six probes that answer *is a prompt on screen*. It left four windows untouched: both in `detectPaneState`, both in `shouldRetrySubmit`.

Measured on a real `tmux capture-pane` of a busy fleet agent (2026-09-06), appending an 18-line blank tail:

```
farok NELKUL : state=busy   esc@busyRegion(12)=true  esc@footerRegion(5)=true
ures FAROKKAL: state=idle   esc@busyRegion(12)=false esc@footerRegion(5)=false
```

The direction is the costly one. An idle verdict is a licence to deliver, so the router pushes a message into a live turn. That is exactly the interleaving the pane monitor exists to prevent, and a blank tail is not exotic: `capture-pane` returns one whenever the content is shorter than the pane.

## Both windows in each function, and why

The two windows catch different shapes. `esc to interrupt` lands on the footer line; a spinner lands several rows higher, inside the wider busy window and outside the narrow footer one. Converting one and not the other leaves whichever shape falls in the other window still blind. Each of the four is pinned by its own input class in the new suite.

## Mutation control

234 assertions green. Five axes, each reverted to the pre-fix shape, every restore byte-identical:

| axis | assertions that fail |
|---|---|
| `detectPaneState` footer window reverted | 1 |
| `detectPaneState` busy window reverted | 1 |
| `shouldRetrySubmit` busy window reverted | 1 |
| `shouldRetrySubmit` footer window reverted | 1 |
| `liveTailRegion` stops trimming the tail | 6 (including #1205's own two) |

Two of those axes only became discriminating after the suite was fixed, and that is worth stating plainly. The first version of the retry test asserted `false` on a busy pane -- but `shouldRetrySubmit` returns `false` there for several independent reasons, so reverting the window did not move it. It survived the mutation. The test now drives the retry down its placeholder path, where the busy verdict is the only thing standing between a live turn and a clear-and-resend, and it carries a control probe showing the retry *does* fire once the busy signal is removed. The footer axis needed a second input class for the same reason.

## Found while measuring, filed separately: `PANESPINPERC906`

`BUSY_INDICATORS` does not match a minute-form duration. The real capture read:

```
✽ Metamorphosing… (2m 0s · ↓ 6.4k tokens · still thinking with high effort)
```

Both patterns require `\(\s*\d+s`, and `(2m 0s` puts `2m ` between the paren and the seconds. So **any turn past sixty seconds loses its spinner signal entirely** and the busy verdict rests on `esc to interrupt` in the footer alone -- which is precisely why the blank-tail gap above mattered as much as it did.

That is a separate defect with a wider blast radius (it changes how broadly we read panes as busy, fleet-wide), so it gets its own change and its own mutation control rather than riding along here. The fixture in this PR deliberately uses the sub-minute form `(52s · ↓ 2.6k tokens)`, so this suite cannot pass for the wrong reason.

### DIFF
--- a/src/__tests__/pane-busy-blank-tail.test.ts
+++ b/src/__tests__/pane-busy-blank-tail.test.ts
@@ -1,0 +1,120 @@
+import { describe, it, expect } from 'vitest'
+import { detectPaneState, shouldRetrySubmit } from '../pane-state.js'
+
+// A RUNNING turn read as idle, because the pane had empty rows below it.
+//
+// #1205 fixed this shape for the prompt probes: every footer window counted back
+// from the last LINE of the capture, so a pane drawn high with a blank tail pushed
+// the footer out of the window. It converted the six probes that answer "is a
+// prompt on screen". It did not convert the two windows in detectPaneState, nor
+// the two in shouldRetrySubmit -- the functions that decide whether a pane may be
+// written to.
+//
+// Measured 2026-09-06 on a real `tmux capture-pane` of a busy fleet agent: with an
+// 18-line blank tail appended, detectPaneState flipped busy -> idle. The direction
+// is the costly one. An idle verdict is a licence to deliver, so the router would
+// have pushed a message into a live turn, which is the queue-interleaving this
+// monitor exists to prevent.
+//
+// Both windows have to move together. `esc to interrupt` sits on the footer line,
+// but the spinner ("✽ Metamorphosing… (2m 0s · ...)") is drawn several lines above
+// it, inside the wider busy window and outside the narrow footer one -- so a fix to
+// only one of them still misses whichever shape lands in the other.
+
+/** Faithful to a real capture: spinner high, footer last, nothing after it. */
+const BUSY_PANE = [
+  '  Reading the sprint notes before the merge.',
+  '',
+  '✽ Metamorphosing… (2m 0s · ↓ 6.4k tokens · still thinking with high effort)',
+  '  ⎿  Tip: Use /btw to ask a quick side question without interrupting Claude\'s',
+  '     current work',
+  '',
+  '───────────────────────────────────────────────────────────────── Samu ─',
+  '❯ ',
+  '────────────────────────────────────────────────────────────────────────',
+  '  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt · ← for ag…',
+].join('\n')
+
+const BLANK_TAIL = '\n'.repeat(18)
+
+describe('a busy pane with a blank tail below it', () => {
+  it('is still busy, exactly as it is without the tail', () => {
+    // The control probe is the same pane untailed: without it a green assertion
+    // would not distinguish "the tail is handled" from "this fixture never read busy".
+    expect(detectPaneState(BUSY_PANE)).toBe('busy')
+    expect(detectPaneState(BUSY_PANE + BLANK_TAIL)).toBe('busy')
+  })
+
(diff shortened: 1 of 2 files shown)


Szotasz:
Independent verify done — **GO**. (Merge left to the coordinator: `pane-state.ts` is named in the recovery stack, so I'm not auto-merging after my own verify even as the second pair of eyes.)

What I measured (worktree, union with fresh develop, head c25d85a5):
- **Independent live reproduction, not the PR's capture:** took a `tmux capture-pane` of my own agent pane mid-turn (footer showing `esc to interrupt`) and an idle peer pane. On develop: busy pane + 18-line blank tail flips **busy → idle** (the write-licence direction). On the PR: stays busy. The idle capture stays idle with and without the tail on both versions — the widening buys no false-busy on real captures.
- Full suite green: 4768 passed, 1 skipped; `tsc --noEmit` clean.
- **Own mutation axes** (not the PR's five): window-size mutation on the retry footer (5→12) fails exactly 1 assertion — the size is pinned, good.

Two silent survivors, neither blocking:
1. **Busy-scan scope on `detectPaneState` is unpinned:** reverting `busyRegion` to a whole-pane scan leaves all 4768 tests green. That is the stale-token-counter regression the comment above the line exists for (false busy = delivery starvation, the 94-retry shape). Pre-existing gap, not introduced here — but one fixture with a stale counter line in scrollback would pin it, maybe alongside the `PANESPINPERC906` work since both live in the busy-verdict set.
2. `liveTailRegion`'s `.trim()` is unpinned (an `=== ''` mutation survives). Defensive only — the shipped path (`capture-pane -e -p` + ANSI strip) yields genuinely empty lines, and my four real captures had zero whitespace-only lines — but a one-line assertion would make it non-revertible.

Szotasz:
Addendum to the GO above — the three attack vectors from the coordinator's note, all run:

**1. Fixture authenticity:** my own independent `capture-pane` of a live turn matches the fixtures' shape — footer line byte-identical (`⏵⏵ bypass permissions … · esc to interrupt · ← for ag…`), same structure (tip block, rules, prompt, footer as last content, no natural blank tail at 24-line height). The earlier reproduction already ran on my capture, not the PR's: develop flips busy→idle with the tail, this branch holds busy. No divergent shape found.

**2. The unmeasured direction (dead pane):** a pane frozen mid-turn is content-indistinguishable from a live busy pane, so this branch reads it busy indefinitely — where develop's blank-tail variant read it idle and would have delivered/retried into a dead buffer. The hold is not unbounded on the system level: `message-router`'s session-stuck detection fires specifically on prolonged BUSY with a pending queue (`src/web/message-router.ts:103`) and surfaces to the orchestrator. So the trade is conservative-hold + watchdog alert instead of silent delivery into a dead pane — the right direction, and a different axis covers the remainder (the 06-30 pinning class stays with `PANESPINPERC906`).

**3. A fifth surviving axis:** none found among the windows — reverting each of the four individually fails exactly 1 assertion each (run separately, restores byte-identical). The survivors I did find are outside the four: the busy-scan *scope* on `detectPaneState` (whole-pane mutation, all green) and `liveTailRegion`'s `.trim()` — both already named above, neither blocking.

GO stands.

PR #1217, opened 2026-09-06
fix(secret-gate): read git objects, not the working tree; and except one fixture literal instead of a whole file

iszzu80-dev:
Two changes to the secret gate, in two reviewable commits. Both came out of
looking at why PR #775 has been red for two weeks and why #628 would go red the
moment it is rebased. Neither is an allowlist.

## 1. Scan what the commits contain, not what happens to be on disk

`scripts/secret-gate.ts` resolved every path with `readFileSync`. PR #775 adds
two versioned symlinks that point at directories — `releases/monitor-current`
and `releases/monitor-previous`, both mode `120000`. `readFileSync` follows a
link into a directory and throws `EISDIR`, so the gate reported

```
NOT SCANNED, therefore NOT CLEARED (2):
  releases/monitor-current   read failed (EISDIR: ...)
  releases/monitor-previous  read failed (EISDIR: ...)
```

and failed closed. No secret pattern matched. The gate refused correctly, on a
question it had asked wrongly.

The crash is the symptom. Reading the working tree makes a **range** gate answer
a different question from the one it is named for: it scans whatever is on that
disk at that moment, which in CI is a checkout and locally is an arbitrarily
dirty tree. So the source of truth is now git object storage in every mode — the
head tree for `--range`, the index for `--staged` and `--all` — and a symlink is
scanned as the blob git actually stores for it, which is the target path as
text. Following it would scan a file the commit does not contain, and for a
directory target there is nothing to follow at all.

`--staged` was not part of the original problem and is changed here deliberately
rather than quietly: it had the identical defect, and the index *is* "what is
about to be committed", so an edit made after `git add` can no longer smuggle a
secret past the pre-commit hook.

Eight runner cases in `src/__tests__/secret-gate-runner.test.ts`, four of which
go RED against the previous implementation: the symlink case, both
working-tree-versus-commit directions, and the `--staged` index case. The other
four are invariants that must hold either way — including a regression test for
a bug introduced while writing this: `git cat-file --batch` answers one record
per line *sent*, and a loop that skipped a record without advancing its cursor
reported one file's content under another file's name. It was caught by running
`--all` (990 files) under both implementations and diffing the verdicts, not by
a test.

Measured on the real #775 range `fde78707..a8577259`: 185 files in scope — the
same count CI reports — PASS.

## 2. Except ONE fixture literal, not a whole file

`ALLOWLISTED_PATHS` blinds an entire file. That is the right trade when the
file's whole subject is synthetic secrets (the gate's own tests) and the wrong
one for an ordinary test file that needs a single key-shaped literal: from then
on nothing in it is checked, including the parts nobody was thinking about when
the entry was added.

PR #628 is the case. Two costops collector tests carry an `sk-` shaped fixture,
and both literals are load-bearing rather than decorative: one drives *"rejects
a raw secret in secret_ref (must be vault:)"*, the other drives *"the secret
never reaches the report, the audit row or the DB dump"*. Replace them with a
neutral string and both tests still pass while measuring something weaker than
their names claim. Blind both whole files to keep two lines, and the gate stops
watching 43 files' worth of new code.

So an exception names **one literal in one file**, keyed on the **sha256 of the
matched text**:

* not a line number, which goes stale the moment anyone edits above it and can
  then cover a line nobody approved;
* not the literal itself, because this file is scanned too, and writing a secret
  shape into the scanner to excuse a secret shape elsewhere is how the exception
  becomes the leak;
* a hash is exact and fails closed — edit the fixture and the exception simply
  stops applying, the gate goes red, and a human looks again.

Constrained on every axis: **test-only** (a non-test path can never carry one,
checked at validation and again at lookup), exact literal scope, a written
reason that must be substantial, and the rest of the file still scanned. A
malformed entry is not inert, it is a **finding** — the gate refuses to run with
an exception it cannot justify. Applied exceptions are printed in the report
with file, line, detector and reason, because an exception nobody sees is an
exception nobody re-reads.

Eight cases prove the narrowness in both directions, including a second secret
in the same excepted file still blocking, the same literal in a different file
not being covered, and an edited fixture losing its exception. They run against
**synthetic** entries on purpose: testing through the real ones would mean
writing their literals into the test file, which would put the strings this
exists to contain into one more place — and that file would then need its own
exception.

Measured on the real #628 range: 43 files in scope, 2 exceptions applied, PASS.

## Why this comes first

Rebasing #628 onto current `develop` produces a clean tree and a green suite,
but its `secret-gate` job fails on those two fixtures under today's gate
(measured: 2 BLOCKED, exit 1). With this change it passes (42 files, 2
exceptions, exit 0). #775's `scan` job goes green here too. So this is the
prerequisite for both, and it carries no product change of its own.

## What this PR does not claim

The local suite on this branch is green except for one pre-existing failure that
has nothing to do with it: `scripts/__tests__/intel-db.test.sh` shells out to the
`sqlite3` CLI, which is not a declared dependency of this repository — it is
absent from `package.json`, absent from the CI workflow, and `scripts/backup.sh`
explicitly tolerates its absence. The test can use Python's stdlib `sqlite3`
module instead, which every environment already has. That is a separate,
one-file change and is deliberately not bundled here.

---

https://claude.ai/code/session_015HDX2xPs9u6B9nyG4ekiRm

### DIFF
--- a/scripts/secret-gate.ts
+++ b/scripts/secret-gate.ts
@@ -228,6 +228,12 @@ function report(result: GateResult, mode: string, entries: Entry[]): void {
     for (const l of links) line(`  - ${l.path}`);
   }
 
+  if (result.exceptions.length) {
+    line();
+    line(`Fixture exceptions applied (${result.exceptions.length}) -- ONE literal each, the rest of the file still scanned:`);
+    for (const e of result.exceptions) line(`  - ${e.file}${e.line ? `:${e.line}` : ''}  [${e.detector}]  <- ${e.reason}`);
+  }
+
   if (result.allowlisted.length) {
     line();
     line('Allowlisted by path (NOT scanned, on purpose):');
--- a/scripts/secret-gate.ts
+++ b/scripts/secret-gate.ts
@@ -12,53 +12,221 @@
  *
  * Everything this runner cannot scan is REPORTED and FAILS. Size limits, binary
  * files, read errors: each is named. A silent skip would read as "clean".
+ *
+ * ── THE SOURCE OF TRUTH IS GIT OBJECT STORAGE, NEVER THE WORKING TREE ────────
+ *
+ * Found 2026-09-06 on PR #775, which was red for two weeks with NO secret in it.
+ * The runner resolved each path with `readFileSync`, and two of the paths were
+ * versioned symlinks pointing at directories (`releases/monitor-current`,
+ * `releases/monitor-previous`, both mode 120000). `readFileSync` on those
+ * follows the link into a directory and throws EISDIR, so the gate reported
+ * "NOT SCANNED, therefore NOT CLEARED" and failed closed -- correctly, by its
+ * own rule, on a question it had asked wrongly.
+ *
+ * The deeper problem is not the crash. Reading the working tree makes the gate
+ * answer a question nobody asked: it scans whatever happens to be on this disk
+ * right now, which in CI is a checkout and locally is an arbitrarily dirty tree.
+ * What a range gate must scan is what the COMMITS CONTAIN. So every mode now
+ * resolves content from git:
+ *
+ *   --range   the head tree of the range
+ *   --staged  the index -- which is precisely "what is about to be committed",
+ *             and is not the same thing as the working tree once a file has been
+ *             edited after `git add`
+ *   --all     the index
+ *
+ * A symlink is then scanned as what git actually stores for it: a small blob
+ * holding the TARGET PATH as text. That is the honest object to scan. Following
+ * it would scan a file the commit does not contain, and for a directory target
+ * there is nothing to follow at all.
  */
 import { execFileSync } from 'node:child_process';
-import { readFileSync, statSync } from 'node:fs';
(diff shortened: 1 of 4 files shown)


iszzu80-dev:
Context on why this one is worth taking first: it unblocks two other open PRs.

- **#775** (`memory-pressure-monitor`) — its `scan` job fails on `releases/monitor-current` and `releases/monitor-previous`, two versioned symlinks pointing at directories. `readFileSync` throws `EISDIR`, so the gate reports "NOT SCANNED, therefore NOT CLEARED" and fails closed. No secret pattern matches. Measured on the exact range `fde78707..a8577259`: 185 files in scope, PASS with this change.
- **#628** (`costops` collectors) — rebased onto current develop, `test` passes 387/387, and `scan` fails on two `sk-` shaped fixtures whose shape is load-bearing for the assertions they drive. Measured on its range: 43 files, 2 exceptions applied, PASS with this change.

Neither needs anything of its own once this lands; both should go green on a re-run.

Both commits here are independently reviewable and the second does not depend on the first beyond sharing the file.

PR #1228, opened 2026-09-07
feat(ledger): record which message an inbound Telegram reply quoted

palvolgyidigital:
When a user replies to a specific earlier message on Telegram, the agent cannot
see which message it was, so a one-word answer attaches to the wrong thread.

**It caused a real error here, 2026-09-04.** A colleague quote-replied
"lezárható" ("you can close it") to one concrete message. The agent had several
open threads with him, could not see the quote, guessed the wrong one, and
closed the wrong card -- a business decision was recorded incorrectly. His
actual answer arrived 42 seconds later and said something different. His words
were: *"nem értem, hogy miért nem látod, azt hogy ha egy üzenetre válaszolok
konkrétan, hogy lezárhatod akkor miért más témához írod?"*

**Root cause.** The Telegram Bot API update carries `message.reply_to_message`,
and the channel plugin already uses it for another purpose, but the inbound
`<channel>` notification never passed it to the model. The outbound direction
could already quote (the reply tool takes `reply_to`); only the inbound half was
missing.

**What this commit does** is the persistence half: a nullable
`reply_to_message_id` column on `conversation_log`, and the write path that
fills it. That makes the association recoverable afterwards, not just visible in
the moment.

Deliberately NOT stored: the quoted text excerpt. It is a copy of someone else's
message, and it is retrievable from the message id.

Verified on this branch: `tsc` clean, the schema test 5/5, and the ledger shell
suite 52/52. Two real quoted replies have been captured live at our install
since the change went in.

### DIFF
--- a/scripts/__tests__/conversation-ledger.test.sh
+++ b/scripts/__tests__/conversation-ledger.test.sh
@@ -97,6 +97,21 @@ print(json.dumps(payload))
 PYEOF
 }
 
+# Emit an inbound payload that quotes an earlier message (reply_to_message_id).
+emit_inbound_replyto() { # chat_id message_id text reply_to_message_id [cwd]
+    python3 - "$@" <<'PYEOF'
+import json, sys
+chat_id, message_id, text, reply_to = sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4]
+block = (f'<channel source="plugin:telegram:telegram" chat_id="{chat_id}" '
+         f'message_id="{message_id}" user="x" reply_to_message_id="{reply_to}" '
+         f'ts="2026-06-02T14:20:25.000Z">\n{text}\n</channel>')
+payload = {"hook_event_name": "UserPromptSubmit", "prompt": block}
+if len(sys.argv) > 5:
+    payload["cwd"] = sys.argv[5]
+print(json.dumps(payload))
+PYEOF
+}
+
 # Emit a Telegram reply PostToolUse payload.
 emit_reply() { # chat_id text [cwd]
     python3 - "$@" <<'PYEOF'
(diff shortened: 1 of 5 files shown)


Szotasz:
We verified this and we are taking it. The schema parity between the TypeScript and Python sides is exact, the drift test was updated and passes, the migration is additive and idempotent on both sides, the shell suite is green on macOS as well as Linux, and storing only the quoted id rather than an excerpt is the right call, since it leaves nothing to drift out of sync.

One thing we found that you should know about, because it changes what this PR delivers today rather than whether we want it.

**Nothing currently emits the attribute you read.**

The hook reads `reply_to_message_id` off the channel tag. We checked three places: the installed Telegram plugin builds its inbound meta with chat id, message id, user, user id, timestamp and attachment fields, and no reply field; there is no other tag producer in this repository; and the attribute does not occur in our live transcripts. So after this merges, every inbound row keeps a null in that column, and the promise in the title, recording which message an inbound reply quoted, does not yet hold on the installed chain.

Your test is green because its fixture supplies the attribute itself. That is a reasonable way to test a parser, and the parser is correct. It just cannot tell you whether anything upstream produces the input.

**What we are doing with it.** Merging it as plumbing, with the limitation written down in the places our own rounds actually read, and opening a separate item for the producer side, which lives in the plugin and is not yours to fix. We would rather have the column and the parsing ready than hold your work behind a dependency you do not control. The reason we are spelling this out is that a merged feature which never fires reads, six months later, as a feature that works.

**A question, and it may change how we prioritise the producer side.** Did you observe `reply_to_message_id` arriving somewhere, on a different plugin version or a different channel, or was this written in anticipation of it? Either answer is fine. If you have seen it in the wild, we are measuring the wrong plugin build and we would like to know.

The merge itself waits until tomorrow, for a release reason on our side that has nothing to do with this PR.

palvolgyidigital:
Yes — observed in the wild, not written in anticipation. Two independent measurements on our host.

**1. The producer exists in the installed plugin.** `~/.claude/plugins/cache/claude-plugins-official/telegram/0.0.7/reply-meta.ts` exports `replyMeta(replyTo)`, returning `{ reply_to_message_id, reply_to_excerpt }`, and `server.ts` imports it at line 26. It is fail-safe by design: a non-reply message yields `{}`, so the tag looks unchanged for the common case.

**2. Real inbound tags carry it, with values.** Not fixtures — two live `<channel>` tags from our own transcripts, both 2026-09-07:

- `message_id="5142"` … `reply_to_message_id="5115"`
- `message_id="5160"` … `reply_to_message_id="5100"`

Both also carried a populated `reply_to_excerpt`.

**A likely explanation for the difference.** There are two copies of the plugin on this machine, and only one has the producer. The marketplace copy at `plugins/marketplaces/claude-plugins-official/external_plugins/telegram/server.ts` has zero occurrences of `replyMeta` or `reply_to_message_id`; the cached, installed copy has both. If you checked the marketplace copy, you would have seen exactly what you described.

**One caveat on identifying the build.** The cache directory is named `0.0.7`, but the `version` field in its own `package.json` says `0.0.1`. The path is the reliable identifier here, not the version string.

Thanks for spelling out the limitation in the merge — you were right that a merged feature which never fires reads later as a feature that works. In our case it does fire, so the column should populate on this build.

PR #1229, opened 2026-09-07
feat(overview): show the subscription quota, with the age of the reading

naki003:
## Problem

A fleet draws its 5-hour and weekly quota from one pool, and until now the only place those numbers appeared was a terminal status line. Whoever was not attached to a pane could not see how much was left, which is exactly the person most likely to start a long run.

## Fix

`scripts/statusline-ratelimit.sh` already writes the quota block to `store/.claude-rate-limits.json` on every render, at no token cost, and `scripts/limit-monitor.sh` already alerts on it. This adds the missing reader: a narrow strip above the overview stats with one bar per window, the percentage, and the countdown to the reset.

## The age of the reading is part of the strip, not a tooltip

A stale quota number is more dangerous than a missing one: a six-hour-old number looks exactly like a fresh one and reassures just as much. Past `QUOTA_MAX_AGE_SEC` (the same knob and the same 21600 s default the monitor uses) the bars keep their numbers but lose their colour, and the strip says the reading is out of date. A window whose `resets_at` has already passed is treated the same way.

When there is no reading at all, the strip stays visible and says why, with a different message for each cause: the file is absent, the file is unreadable, or it carries no `rate_limits` block. An absent strip and a healthy one look identical from across a room, which is the one thing this change exists to avoid.

## Tests

`src/__tests__/quota-snapshot.test.ts`, 10 cases: missing file, malformed JSON, fresh reading, stale reading, elapsed reset, both windows, a caller-supplied threshold, and the max-age boundary itself. The boundary case pins `age === maxAge` as fresh, matching `scripts/lib/quota-check.py`; both sides compare with a strict `>`, and that agreement is what stops the strip and the monitor's alert from contradicting each other. 10/10 green.

## Note for reviewers

The warning colour at the 60 and 80 percent thresholds comes from the existing palette, which has no yellow, so the middle band renders orange. That is a one-line change in the stylesheet if you would rather it looked different.


Szotasz:
Thank you for this, and sorry it sat for three days. Reviewed properly rather than skimmed: every codebase claim in the description was measured against the head and against the base, because a description is not a source.

**Verdict: change request, two small items.** Nothing in the behaviour is blocking, and the reasoning in the description holds up where I could check it.

## What I measured, and it holds

- `scripts/statusline-ratelimit.sh` does write `store/.claude-rate-limits.json` with `written_at` and `rate_limits` on every render (lines 26, 50, 53).
- `scripts/limit-monitor.sh` really does default to `QUOTA_MAX_AGE_SEC=21600` (line 99), and the route reads the same environment knob, so "the same knob" is accurate rather than aspirational.
- `scripts/lib/quota-check.py` does skip expired windows, so treating a passed `resets_at` the same way keeps the two consistent.
- **The staleness boundary agrees on both sides**: `ageSec > maxAgeSec` here, `age > max_age` in `quota-check.py`. Both strict, so a reading exactly at the threshold is fresh in both places. This is the part that makes the strip and the alert unable to disagree, and it is correct.
- The percentage is clamped to 0..100 before it reaches the bar width, so a malformed payload cannot paint outside the track.
- A missing `written_at` becomes age 0, which lands in `stale` rather than `ok`. Failing to stale rather than to fresh is the right direction and the suite pins it.
- i18n: all 13 keys the strip uses exist in both `hu.js` and `en.js`. A missing key here would have rendered as a raw dotted string in production.
- Your test suite: 9 passed on your head.

**The one thing the green checks did not prove.** Your branch forks from a `develop` that is now 48 commits behind, and `MERGEABLE/CLEAN` only means there is no textual conflict, not that the merged state works. I merged current `develop` into your head locally and ran it there: `tsc --noEmit` clean, full suite 404 files, 5158 passed, 1 skipped, 0 failed. So the gap is harmless, but it was worth measuring rather than assuming.

## Change requests

**1. Pin the exact boundary in the suite.** The tests cover `maxAge + 1` (stale) and 100 seconds against thresholds of 60 and 600, but not `age === maxAge`. That equality is the whole basis of the claim that the strip and the alert cannot disagree, and today it holds only by both sides happening to use `>`. One assertion makes it durable:

```ts
it('treats a reading exactly at the threshold as fresh, like the monitor does', () => {
  write(healthy({ written_at: NOW - DEFAULT_MAX_AGE_SEC }))
  expect(readQuotaSnapshot(file, NOW).status).toBe('ok')
})
```

**2. One line of the description is wrong, and the code is better than it says.** "If the file does not exist at all, the strip renders nothing rather than showing zeroes." What `renderQuotaStrip` actually does is keep the strip visible and put a specific line in it saying why there is no reading, with a different message for `no-file`, `unreadable` and `no-rate-limits`. That is the better behaviour, since an absent strip and a healthy one look identical from across a room. Worth correcting so the next reader does not trust the weaker version.

## Two findings that are ours, not yours

Raised here for the record; neither is a change request against this PR and both are being carded on our side.

- **`QUOTA_MAX_AGE_SEC` has no upper bound anywhere**, not in your route and not in the existing `limit-monitor.sh`. A typo with one extra zero makes a reading from weeks ago render as fresh and coloured, which is precisely the failure this PR exists to prevent. Your code inherits the gap rather than creating it, but it inherits it into the place where it does the most damage.
- **"A fleet draws its 5-hour and weekly quota from one pool"** is true on our install (no `claude-plans.json`, all 14 agents on the default), but the product supports per-agent Claude plans with their own config directories. On such an install the file holds whatever session rendered last, and the strip would present it as the fleet's. Not something you could have known from outside.

## On the orange

The palette note is fine as it is. If the middle band ever needs its own token, that is a separate change and not worth holding this one for.

Once the boundary test is in, this is ready from my side.

naki003:
Both change requests are in, as `b6f9871`.

**1. The boundary is pinned.** Added essentially your assertion, plus the age check so a future reader can see what the number is meant to be:

```ts
it('treats a reading exactly at the threshold as fresh, like the monitor does', () => {
  write(healthy({ written_at: NOW - DEFAULT_MAX_AGE_SEC }))
  const snap = readQuotaSnapshot(file, NOW)
  expect(snap.status).toBe('ok')
  expect(snap.ageSec).toBe(DEFAULT_MAX_AGE_SEC)
})
```

I checked that it bites rather than merely passing: flipping `quota.ts` to `ageSec >= maxAgeSec` fails this case and nothing else in the file, which is the gap you described. 10/10 green now.

**2. The description is corrected.** You were right that the code is better than the sentence claimed. The strip stays visible and names the cause, with a separate message for `no-file`, `unreadable` and `no-rate-limits`; the body now says that, and gives your reason for why it matters, since it is a better reason than the one I had.

**On the two findings that are yours.** Both carded on our side rather than folded in here, since neither is this PR's to fix:

- The missing upper bound on `QUOTA_MAX_AGE_SEC` is the one I would take seriously. Your framing is the part I had not seen: the gap is inherited from `limit-monitor.sh`, but this change moves it to where it does the most damage, because a wrong bound now paints a confident coloured bar instead of failing an alert quietly. If you would rather it were fixed here, say so and I will add the clamp to both sides in this PR; my instinct is that it belongs in its own change, touching the shell script and the route together.
- The per-agent plans point is fair and I could not have measured it from outside. On our install the pooled reading is correct, and the file has no field that would let the route tell the two topologies apart, so I would not want to guess in the strip. Worth a card at your end for whoever adds `claude-plans.json` support.

Thank you for measuring the merged state rather than trusting the green check. The 48-commit gap was invisible from my side, and "MERGEABLE means no textual conflict" is the sort of thing that is obvious only after someone says it.

Szotasz:
Merged as e30161a. Thank you -- the reading's age next to the number is exactly the part that makes it
honest, and the new boundary case earns its place: we re-measured it by loosening the strict
comparison in `quota.ts`, and 1 of 10 went red, precisely the new one. The other nine stayed green,
so it is not a duplicate of the existing case (that one measures BEYOND the threshold, this one AT it).

One finding from the review that is NOT a problem with this PR, and that we are taking on ourselves
as the next item. The test comment states a guarantee: that the strip and the monitor agreeing is
"the whole reason they cannot contradict each other, and without this assertion it only holds by
accident". That is true for the TypeScript side. It is not yet true for the monitor side: loosening
the same comparison in `quota-check.py` leaves the shell test ALL PASS and all 36 TS quota cases
green. So on the monitor side the threshold is still unguarded.

This matters a little more since this morning, because `develop` gained its own
`DEFAULT_MAX_AGE_SEC = 21600` in the monitor (#1280) -- the same value as in `quota.ts`, but now in
two places with no cross-check. The fix we are adding is therefore not a repeat of your boundary case
on the python side, but an assertion that the two defaults are equal, which covers both gaps at once.

Nothing for you to do here. We mention it because the guarantee sentence was worth taking literally,
and taking it literally is what found the missing half.

PR #1230, opened 2026-09-07
fix(security): device-identity check for shared bearer token spoofing

B-Aut:
## Summary
- A megosztott bearer token korábban bármely ágensnek lehetővé tette, hogy tetszőleges (akár nem létező) agent_id-vel járjon el approval kérésnél, kanban kommentnél, memória írásnál és üzenetküldésnél.
- Minden ilyen művelet mostantól ellenőrzi, hogy a megadott ágens ténylegesen létező flotta-tag (`isKnownAgent`), plusz egy device-key alapú réteg is bekerült az auth-gate-be (`deviceIdentityMismatch`), ami egy adott hívó eszközt egyetlen agent névhez köt.
- Munka: Sentinel Bon, a réteg a 2026-08-26-i identitás-hamisítási felfedezés (MEMOWNER826) folytatása.

## Test plan
- [x] `npx tsc --noEmit` -- tiszta
- [x] `npx vitest run` a 6 érintett teszt-fájlra (approvals-notify, auth-gate, kanban-write-gate, messages-post-sender-guards, memories-get-agent-id-alias, installer-service-auth-gate) -- 75 teszt, mind zöld
- [ ] Kód-review develop-en (a branch develop-re lett rebase-elve/mergelve, nem main-re, mert a repo main aktuálisan 100 commit-tal a develop mögött áll)

Megjegyzés: a `.claude/settings.json` eredeti (main-en dolgozó) módosítása kimaradt ebből a PR-ből -- az csak formázási/plugin-config drift volt, semmi köze a biztonsági javításhoz.

### DIFF
--- a/src/__tests__/approvals-notify.test.ts
+++ b/src/__tests__/approvals-notify.test.ts
@@ -8,6 +8,16 @@ vi.mock('../config.js', async (importOriginal) => {
   return { ...real, MAIN_AGENT_ID: 'agent-a' }
 })
 
+// REQUESTORGUARD829: approval creation now rejects any agent_id that is not a
+// known fleet agent (see approvals.ts). These fixtures use synthetic ids
+// ('agent-a', 'agent-b') with no real `agents/<id>/` directory on disk, so
+// isKnownAgent is mocked to recognize exactly the two ids this suite exercises
+// -- same shape as the MAIN_AGENT_ID mock above, kept minimal on purpose.
+vi.mock('../web/agent-config.js', async (importOriginal) => {
+  const real = await importOriginal<typeof import('../web/agent-config.js')>()
+  return { ...real, isKnownAgent: (name: string) => name === 'agent-a' || name === 'agent-b' }
+})
+
 import { tryHandleApprovals } from '../web/routes/approvals.js'
 import type { RouteContext } from '../web/routes/types.js'
 
@@ -91,7 +101,12 @@ describe('approvals self-approval guard', () => {
     expect(postOut.status).toBe(201)
     const id = postOut.body.id
 
-    // agent-b attempts to approve its own request
+    // agent-b attempts to approve its own request. REQUESTORGUARD829 means
+    // agent_id at creation must now be a known fleet agent, so resolved_by
+    // being that SAME name is also always a known agent -- AGENTRESOLVE826
+    // ("no fleet agent may resolve any approval") fires before the older,
+    // more specific self-approval guard ever gets a turn. The outcome (403,
+    // blocked) is unchanged; only the wording is now AGENTRESOLVE826's.
     const { ctx: patchCtx, out: patchOut } = fakePatch(id, {
       status: 'approved',
       resolved_by: 'agent-b',
@@ -99,7 +114,7 @@ describe('approvals self-approval guard', () => {
     const handled = await tryHandleApprovals(patchCtx)
     expect(handled).toBe(true)
     expect(patchOut.status).toBe(403)
-    expect(patchOut.body.error).toMatch(/cannot approve its own request/)
+    expect(patchOut.body.error).toMatch(/not a fleet agent/)
   })
 
   it('allows approval when resolved_by differs from the requesting agent_id', async () => {
--- a/src/web/auth-gate.ts
+++ b/src/web/auth-gate.ts
@@ -119,3 +119,29 @@ export function resolveAuth(
 
   return { kind: 'none' }
 }
+
+// DEVICEIDENTITY829: shared bearer token is still the only credential every
+// fleet agent has, so a body-declared identity field (from/agent_id/author/
+// resolved_by) can always be forged by any holder -- the isKnownAgent guards
+// in each route only reject an UNKNOWN name, not a stolen KNOWN one. A device
+// key closes that: it is a hashed, per-caller credential (auth-device-keys.ts),
+// so when a request authenticates as one, the caller cannot legitimately claim
+// any identity but its own. Once every fleet agent holds a device key, this
+// makes body-declared identity redundant for those callers; while migration is
+// in progress (some callers still on the shared token), this only tightens the
+// device-authenticated ones and leaves 'token' callers exactly as before.
+//
+// Route usage: call BEFORE the existing isKnownAgent-family guard, with the
+// identity value the route is about to persist. Returns an error string to
+// send as 403, or null to fall through to the existing guard unchanged.
+// Takes the flattened `RouteContext['auth']` shape (not `AuthResult` directly)
+// since that is what every route handler actually holds.
+// Sentinel Bon + Logra Bon, device-key migration plan, 2026-08-29.
+export function deviceIdentityMismatch(
+  auth: { kind: string; device?: string } | undefined,
+  claimed: string,
+): string | null {
+  if (!auth || auth.kind !== 'device') return null
+  if (auth.device === claimed) return null
+  return `device key '${auth.device}' may not act as '${claimed}'`
+}
--- a/src/web/routes/approvals.ts
+++ b/src/web/routes/approvals.ts
@@ -9,6 +9,8 @@ import {
 } from '../../db.js'
 import { logger } from '../../logger.js'
 import { readBody, json } from '../http-helpers.js'
+import { isKnownAgent } from '../agent-config.js'
+import { deviceIdentityMismatch } from '../auth-gate.js'
 import { resolveOwnerChatId } from '../../owner-chat.js'
 import { sendTelegramMessage } from '../telegram.js'
 import type { RouteContext } from './types.js'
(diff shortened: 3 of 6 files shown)


Szotasz:
Thanks for this. We ran a full verify on a worktree at your head, rebased onto current `develop`, and the direction is right: the self-elevation hole this closes is real, and on live data the resolve guard does not bite any legitimate call we could find. We want this in.

Three things need a fix before we can merge. The first two block, the third would only show up on some installs.

**1. The suite is red: a positive control fails.**

`src/__tests__/kanban-comment-requires-card.test.ts` fails, because the new author guard rejects the fixture author used there. You mocked `isKnownAgent` in `approvals-notify.test.ts`, but that suite was not covered. Mechanical, but we cannot merge on a red suite.

**2. The device check runs before the owner and system exceptions, and that locks out the repository owner.**

This is the one we would ask you to think about rather than patch narrowly. On the kanban and messages routes, `deviceIdentityMismatch` is evaluated ahead of the owner/system exemptions, and a device key is named after the device, not after an agent id.

In our live install exactly one device key exists, and the dashboard injects any `?token=` value, including a device key, as the Bearer token on every `/api` call. The result in such a session:

- posting a kanban comment with the author the UI fills in by default gets a 403 saying the device key may not act as that name
- sending a message with the sender the UI resolves gets a 403
- creating or editing a memory with the agent id the UI always sends gets a 403

So the repository owner's own writes from their own dashboard would start failing. That is a behaviour change no one asked for, and it is the kind that shows up as "the dashboard is broken" rather than as a security message.

Three directions, any of which works for us, and we do not have a preference:

- run the owner and system exemption before the device check
- let the device check apply only when the key's name is itself a known agent id, which we read as the intent of the migration plan anyway
- add an explicit device to agent mapping

**3. `isKnownAgent` resolves through directory existence, so it is case sensitive on some filesystems.**

Identities that carry an uppercase letter, including the bot name that appears in the dashboard's own author dropdown, pass on a case insensitive filesystem such as APFS and fail with a 403 on a Linux install. Our own historical comment authors are mostly in that uppercase form, so this would not be an edge case for them.

A lowercase normalised comparison against the agent list plus the main agent id would settle it.

**Scope, so you do not spend time on our side of the line:** please keep this to the three points above. Anything else the verify touched is ours to handle, and we are not asking you to widen the PR.

Once you push a fix we will re-run the same verify on the new head. The measurement above came from a real run against live data, not from reading the diff, so if you disagree with any of it, say so and we will re-measure rather than argue.

Szotasz:
Seen, and thank you for turning this around. Your `2a7db902` landed yesterday evening and the
headline says it addresses the case-insensitivity and the owner fallback, which were two of the
three findings from our verify.

Two things before the re-verify, and the second one is on you.

First, the state as measured just now: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. That is
not about your commit. `develop` moved underneath the branch while it waited, and the conflict
appeared with that movement. There is nothing for us to re-verify until the branch sits on a base
that matches today's `develop`, because a suite run on the current head would be measuring a tree
that will never exist.

So the next step is a rebase onto current `develop`, then a push. Once the head moves again we pick
it up for the second verify round, and that round reads all three findings, not only the two the
headline names.

Second, the honest timing note. Our `develop` is under a merge freeze today until midday local time
while a release is cut, so nothing lands before then regardless of state. The rebase is still worth
doing now, because the re-verify can start as soon as the freeze lifts and a fresh base is the thing
that takes longest to organise afterwards.

One request for the rebase itself: resolve toward keeping both sides where both are legitimate.
Where your change and an incoming change touch the same guard, we would rather see two guards than
a silent pick of one. If a resolution forces a real choice, say so in a comment rather than folding
it into the rebase, and we will decide it here.

This one is a security path, so it gets a full second round rather than a spot check. That is not
distrust of the fix, it is the class of the file.

Szotasz:
We rebased this branch onto current `develop` ourselves and pushed the result to your branch. The conflict was ours. Your guards are unchanged; what moved is the handler they sit in, and part of your diff has been overtaken by the base.

**What the base already had, so you know what is left.** `messages.ts` acquired the same requestor check while this PR waited: the POST sender path already refuses a `from` that is not a known fleet agent. Device keys in `auth-gate.ts` are older still (#729, 2026-07-27) -- your `deviceIdentityMismatch` helper is new and sits on top of that lane rather than introducing it. What was genuinely unguarded, and is what this PR now delivers, is the other three: `approvals.ts`, `kanban.ts` and `memories.ts`.

**The conflict, and how it was resolved.** `develop`'s `PUT/PATCH /api/memories/:id` handler grew considerably: an `updated_by` write-trace distinct from `agent_id`, category validation against `MEMORY_CATEGORIES`, a partial-update backfill so a category-only change does not have to resend the content, and the suspicious-content filter. Your version of that handler predates all of it. We kept `develop`'s handler in full and grafted your two guards in ahead of the write, rather than restoring the older shape your diff was written against.

**One thing we checked explicitly, because it would have been a live regression.** Both helpers no-op when `agent_id` is absent, and the fleet's documented category-only move (`PATCH` with just `category` and `updated_by`) sends no `agent_id`. So that path is unaffected -- verified in the code, not assumed.

Measured on the rebased tree: `tsc` clean; `approvals-notify` + `kanban-comment-requires-card` + `auth-gate` 27/27; the full suite 498 files, 6214 assertions green, 1 skipped.

**One review finding, and for a security change it is the one worth acting on: none of the four new guards is pinned by a test.**

We removed each in turn and measured:

- `approvals.ts`, the `isKnownAgent(agent_id)` rejection -- full suite green, 6214.
- `memories.ts` POST, `rejectUnknownAgent(data.agent_id)` -- full suite green, 6214.
- `memories.ts` PUT/PATCH, the same guard -- full suite green, 6214.
- `kanban.ts`, the author check -- its own targeted file green, 4/4.

This is not a measurement artefact. As a positive control on the same file and the same instrument, disabling the `agent_id` query alias in the memories GET handler turns three assertions in `memories-get-agent-id-alias.test.ts` red immediately. So tests do reach this route; they just do not reach these four branches.

The two test files in your diff add mocks so the existing suites keep passing with the guard in place, which is necessary, but it means the guard's own behaviour -- a 403 for an unknown `agent_id`, and a device key refused when it claims another name -- is never asserted. Four small tests, one per route, would mean a future refactor cannot quietly remove the thing this PR exists for.

Szotasz (commented):
Independent review of the rebased head `727dab62` (fleet reviewer; the rebase was Geri's, this is a second pair of eyes).

**Measured, on this head**

1. **The four new 403 gates are not asserted by any test.** Reproduced Geri's finding on one gate myself: with the `isKnownAgent(agent_id)` gate in `approvals.ts` replaced by `if (false)`, the two test files this PR touches plus every `approvals*.test.ts` still pass (6 files, 65/65). The PR's test changes add mocks so the existing suite passes *with* the gate; nothing turns red when the gate is removed. A guard nobody measures is one refactor away from silently disappearing.
2. **No live caller is broken by the known-agent gates** (measured against the fleet's own data on the maintainer host): every `memories.agent_id` written in the last 7 days is a registered agent; every `approvals.agent_id` on record is one (`marveen`, `samu`); `resolved_by` values on record are `szabolcs` and `dashboard`, neither a fleet agent, so the resolve gate lets them through. Kanban comments carry non-agent authors (`kartya-es-ertesites`, `cim-kapu (trigger)`), but those are written by a raw-SQL tool and a DB trigger, not through `/api/kanban/*/comments`, so the author gate does not see them. The documented tier-move (`PATCH` with `category` + `updated_by` only, no `agent_id`) is a no-op for both helpers, as Geri measured.
3. `isKnownAgent` going case-insensitive via `listAllAgentNames()` is correct and matches how the fleet actually writes (`Samu`/`samu` both appear in the data).

**What I ask before merge**

For each of the four gates, one test that sends the rejected shape and asserts the 403 (unknown `agent_id` on `POST /api/approvals`; a fleet-agent name in `resolved_by` on the resolve; an unknown `author` on the kanban comment; an unknown `agent_id` on memories `POST` and `PUT/PATCH`), plus one positive case per route so the test proves the route is reached (the existing `memories-get-agent-id-alias` test is the pattern: disabling the alias turns three assertions red, so we know that file exercises the route). With those in, I have no further objection; the code of the gates themselves reads correctly.

Szotasz:
Correction to my previous comment, because it understated your change.

I wrote that "what was genuinely unguarded, and is what this PR now delivers, is the other three: `approvals.ts`, `kanban.ts` and `memories.ts`". That is right about the *requestor* guard -- `messages.ts` already had an `isKnownAgent` check on the sender -- but it reads as if `messages.ts` gets nothing from this PR, and that is wrong. It gets the device-identity check, which is new there as it is everywhere else.

The accurate split, measured on the rebased tree:

- `messages.ts` -- `deviceIdentityMismatch(ctx.auth, from)`: NEW. The pre-existing `isKnownAgent` sender check is a different layer and stays untouched, exactly as your comment in that hunk says.
- `approvals.ts`, `kanban.ts` -- both layers new: the device check and the known-agent rejection.
- `memories.ts` -- both layers new, on POST and on PUT/PATCH.

So the PR adds a device-identity lane to four routes, and a requestor guard to three. My earlier sentence collapsed those two into one number.

The coverage finding extends to the new lane as well, and I measured that too rather than assuming it: removing the `deviceIdentityMismatch` call from `messages.ts` leaves the full suite green at 6214. Same instrument, same validated positive control as before. It does not change the conclusion -- it widens it by one surface.

Szotasz (commented):
Addendum to my review (same head `727dab62`), extending the merge condition by one lane that Geri found and I reproduced.

**The device-identity lane is untested too.** `deviceIdentityMismatch(ctx.auth, ...)` is called at four sites (`approvals.ts` POST, `messages.ts` POST, `memories.ts` POST + PUT/PATCH via the shared helper, `kanban.ts` comment). With all four replaced by a function that always returns `null` ("never a mismatch"), `tsc --noEmit` is clean (so the mutant was really applied) and the route tests around them stay green (7 files, 71/71; Geri measured the full suite the same way: 498 files, 6214 green). That lane is the PR's title, so as it stands the protection the title promises merges with zero assertions.

**Merge condition, complete (six surfaces, two lanes each where applicable):**
- `POST /api/approvals`: unknown `agent_id` -> 403; a device-key caller claiming another agent's `agent_id` -> 403; own name -> passes.
- approval resolve: a fleet-agent name in `resolved_by` -> 403; a human source (`telegram_text`, `dashboard`) -> passes.
- `POST /api/kanban/:id/comments`: unknown `author` -> 403; device-key caller with another name -> 403; own name / owner name -> passes.
- `POST /api/memories`: unknown `agent_id` -> 403; device-key caller with another name -> 403; own name -> passes.
- `PUT/PATCH /api/memories/:id`: same two cases; and the documented tier move (`category` + `updated_by` only, no `agent_id`) must still pass.
- `POST /api/messages`: device-key caller with another `from` -> 403; own name -> passes (the known-sender guard there is pre-existing).

Each test should turn red when its gate is removed; the existing `auth-device-keys` test shows how a device-authenticated request is built. Nothing else in the review changes.

PR #1231, opened 2026-09-07
feat(graph-mail): file attachments, with a drop directory and a size cap

stivi1g-gif:
Split out of #900 at your request, and reworked rather than resubmitted. Reviewing the original diff against your two questions, both already had answers in the code, and both were gaps rather than defences. So this PR carries the fixes instead of an explanation.

**"What stops it reading and sending an arbitrary path?"** Previously: nothing. The `--attach` value went straight into `readFileSync`.

Now `--attach` resolves inside an attachment drop directory (`GRAPH_MAIL_ATTACH_DIR`, default `<repo>/store/mail-attachments`) and anything outside it is refused. Both sides are `realpath`'d before the comparison, so a symlink planted inside the directory cannot reach out of it, and the prefix check includes the separator so a sibling like `<dir>-evil` does not match. Directories and other non-regular files are refused too. Putting a file in the drop directory is the deliberate act that authorises sending it.

**"What happens at the 3MB Graph limit?"** Previously: the limit was a doc comment on the interface field and nothing enforced it, so Graph would have been the thing that failed.

Now the combined base64 payload is capped at `MAX_TOTAL_ATTACHMENT_BASE64_BYTES` (3 MiB) and rejected in `sendMail` before the request is built. The cap sits in the library rather than the CLI so every caller inherits it, and the error names the measured size and the limit.

**"Should an attachment be allowed from a non-interactive caller at all?"** That one is a policy call, not a code question, so I have left it to you. The drop directory narrows it usefully either way: a non-interactive caller can only send what someone already placed there. If you would rather gate it harder, say how and I will add it.

**Verification.** Run, not inspected. A file inside the directory is accepted and proceeds to the credentials step; an absolute path outside, a traversal that resolves to a real file (`/etc/hosts`), a symlink pointing out of the directory, a sibling directory sharing the name prefix, and a directory argument are each refused. The size cap was exercised in both directions: 4 MiB is rejected by the check, 1 MiB passes it and fails later at credentials, so the check distinguishes rather than blocking everything.

#900 now contains only the alias and memory half and is unaffected by this.

### DIFF
--- a/scripts/graph-mail.ts
+++ b/scripts/graph-mail.ts
@@ -4,13 +4,52 @@
 //
 //   tsx scripts/graph-mail.ts verify
 //   tsx scripts/graph-mail.ts list [--unread] [--top N] [--folder inbox|sentitems]
-//   tsx scripts/graph-mail.ts send --to a@b.hu[,c@d.hu] --subject "..." --body "..." [--cc ...] [--html]
+//   tsx scripts/graph-mail.ts send --to a@b.hu[,c@d.hu] --subject "..." --body "..." [--cc ...] [--html] [--attach FILE]
+//
+// --attach does NOT take an arbitrary path. The file must sit inside the
+// attachment drop directory (GRAPH_MAIL_ATTACH_DIR, default
+// <repo>/store/mail-attachments); anything outside it is refused. Putting a
+// file there is the deliberate act that authorises sending it.
 //
 // Credentials come from the gitignored marveen-mail-ugyfelkod file (override
 // with MARVEEN_MAIL_CREDS). Send is intentionally CLI-explicit; the sub-agent
 // email-send-gate hook still applies to any programmatic use elsewhere.
 
 import { listMessages, sendMail, verifyAccess } from '../src/graph-mail.js'
+import { readFileSync, realpathSync, statSync } from 'node:fs'
+import { basename, resolve, sep } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const REPO_ROOT = resolve(fileURLToPath(new URL('..', import.meta.url)))
+
+/**
+ * Resolve --attach against the attachment drop directory and refuse anything
+ * outside it. Symlinks are resolved first (realpath on both sides), so a link
+ * planted inside the directory cannot reach out of it, and the separator on
+ * the prefix check keeps a sibling like `<dir>-evil` from matching.
+ */
+function readAttachment(rawPath: string): { name: string; contentBytes: string } {
+  const dir = realpathSync(
+    resolve(process.env.GRAPH_MAIL_ATTACH_DIR ?? resolve(REPO_ROOT, 'store/mail-attachments')),
+  )
+  let file: string
+  try {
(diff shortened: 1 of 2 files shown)


Szotasz:
Thank you for splitting this out and, more to the point, for reworking it rather than
resubmitting it with an explanation. Both of my questions came back as code, which is the
answer I wanted.

Since the PR ships no tests, I measured the behaviour myself rather than reading the diff
and agreeing with it. I lifted `readAttachment` verbatim into a scratch harness and ran it
against a synthetic tree, with an accepting case included as a positive control so that a
uniformly refusing harness could not pass for a working guard:

```
ACCEPT  file inside the drop directory (positive control)
REFUSE  absolute path outside            -> outside the drop directory
REFUSE  traversal ../secret/key.pem      -> not found
REFUSE  symlink inside pointing out      -> outside the drop directory
REFUSE  sibling dir sharing the prefix   -> outside the drop directory
REFUSE  the directory itself             -> not a regular file
ACCEPT  nested subdirectory inside       -> allowed
REFUSE  /etc/hosts                       -> outside the drop directory
```

Your size claim also holds when measured rather than read: 4 MiB of raw bytes becomes
5592408 base64 bytes and is rejected, 1 MiB becomes 1398104 and passes. The effective
ceiling on a real file is about 2.25 MiB, which sits comfortably under the Graph limit.

Merging this. Three notes, none of which are requests, so please do not widen the PR for
them. Two are ours to fix and I have carded them; the third is a framing point I want on
the record rather than a defect.

1. The drop directory is not in the tree, and `realpathSync(dir)` sits outside the try
   block. On a fresh checkout the first `--attach` therefore fails with a raw
   `ENOENT ... lstat '<repo>/store/mail-attachments'` instead of the message you wrote,
   which is exactly the message that would tell the user where to put the file. Ours to
   fix, either with a tracked `.gitkeep` or by naming that error too.

2. `file.startsWith(dir + sep)` cannot match when `dir` is the filesystem root, because
   the comparison becomes `startsWith('//')`. It fails closed, so nothing unsafe follows
   from it, but with `GRAPH_MAIL_ATTACH_DIR=/` nothing at all can be attached. A trailing
   separator on the configured directory is otherwise handled, since realpath strips it.

3. The framing point. `GRAPH_MAIL_ATTACH_DIR` moves the guard, so a caller that can set
   the environment can point it at any directory and send anything in it. I verified this:
   with the variable aimed at a directory holding a key file, that file is accepted. That
   is not a flaw in your change, since you documented the variable and anyone able to set
   it can already read the file directly. It does mean the guard defends against the wrong
   path being typed and against an agent being talked into attaching something, not
   against a caller that already has the shell. Worth stating plainly so the next reader
   does not credit it with more than it does.

On your open policy question, whether a non-interactive caller should be able to attach at
all: the drop directory is enough for now, and I would rather not gate it harder. What is
missing is not a further restriction but a trace, so that an override of the directory is
not silent. We will add an info log naming the resolved directory when the variable is set,
key name only. That is on our card, not yours.

PR #1233, opened 2026-09-07
fix(gate): pass configDir to msSinceTranscriptWrite so fleet agents are visible

mogganhun:
## The bug

Two calls sit next to each other in `src/web/context-restart-gate-runner.ts`. One passes the agent's Claude config root; the other does not.

```
614: readContextTokensFromProjectDir(workingDir, configDirFor(name))   // passes it
643: msSinceTranscriptWrite(workingDir, nowMs)                          // does not
```

Without the config root, `projectsDirFor(workingDir)` resolves under the host default `~/.claude/projects/`. For any agent launched with its own `CLAUDE_CONFIG_DIR`, no transcript lives there, so `msSinceTranscriptWrite` returns `null` -- and `context-restart-gate.ts:223` turns that into a fail-closed block with no deadline. The gate never opens and never says why beyond `transcript-unreadable (fail-closed)`.

The comment at lines 130-145 warns about exactly this failure, and the call directly beneath it honours the warning. The second call site did not. A warning that holds at one call site is not protection.

## Measured impact

Five of six fleet agents were blind, not one. Measured on this host (re-measured 2026-09-07 22:45; an earlier revision of this description said three, which undercounted zaphod and trillian):

| agent | config root | host-default path | correct path |
|---|---|---|---|
| eddie | `~/.claude-team` | missing | exists, 30 jsonl |
| slarti | `~/.claude-team` | missing | exists, 2 jsonl |
| ford | `~/.claude-team` | missing | exists, 6 jsonl |
| zaphod | `~/.claude-team` | missing | exists, 3 jsonl |
| trillian | `~/.claude-team` | missing | exists, 1 jsonl |
| deept | `agents/deept/.claude-config` | exists | exists (same inode) -- **not affected** |

The symptom that surfaced this: eddie's gate blocked for 120 minutes straight with `transcript-unreadable`, past its 150k-token threshold, with no stuck work behind it.

The main agent is unaffected: `configDirFor` returns `undefined` for it, so it reads the host default legitimately. That is why this went unnoticed.

## Timeline -- when this started, and why it stayed quiet

Two dates, not one, and the gap between them is the reason nobody noticed.

- **2026-08-28 18:25 -- the defect is born.** Commit `160bdbf` (#968, *"fix(context-restart-gate): macOS-blind assumptions kept the gate shut"*) introduces `msSinceTranscriptWrite`, already without the `configDir` argument. The function was never correct. The commit that added it was itself a fix for the gate staying shut.
- **2026-09-04 01:32 -- the defect starts causing harm.** Until then only the main agent was enrolled in the gate, and the main agent is unaffected. On that date the five fleet agents were added to `store/context-restart-gate.json`, and from that moment the gate was watching agents it could not see.
- **2026-09-07 20:04 -- the first visible symptom.** The persistent-block alert fires only after 120 minutes of continuous blocking (`persistentBlockAlertMs`), so the first report was eddie's, and ford's followed at 22:05.

So the harmful window is **3 days 20 hours**, not the ten days since the commit.

What could *not* be established: whether the heavier context-guard restarts increased because of this. eddie's daily rate is 0.33 (08-01..08-28), 0.79 (08-28..09-04) and 0.77 (since 09-04). The step is at 08-28, not at 09-04 -- but the gate was not watching eddie before 09-04, so it cannot be the cause. With 9, 5 and 3 events the series is too short to read a cause from, and it is left unattributed on purpose.

## The fix

`msSinceTranscriptWrite` takes an optional `configDir` and forwards it to `projectsDirFor`; the call site passes `configDirFor(name)`, matching line 614. The function is exported so the routing can be tested directly.

Nothing in the fail-closed logic changes. The gate blocking when it cannot see is correct behaviour; the defect was that it could not see.

## Verification

`npm test src/__tests__/context-restart-gate.test.ts` -- 100 passing.

The new tests were checked against the unfixed code in a separate throwaway worktree, in two stages, because "fails without the fix" can mean two different things:

1. Test file applied to `develop` unchanged: both new tests fail with `TypeError: msSinceTranscriptWrite is not a function`. That only proves the export is missing.
2. Export added, `configDir` forwarding still absent: `returns a non-null age when configDir points to the correct root` fails with `expected null not to be null`.

Stage 2 is the real regression guard -- it fails for the behavioural reason, not a missing symbol. The companion test (`returns null when configDir is not provided`) is a control: it pins the documented no-config-root behaviour and would pass against the broken code by itself.

### DIFF
--- a/scripts/__tests__/stuck-modal-guard.test.sh
+++ b/scripts/__tests__/stuck-modal-guard.test.sh
@@ -113,6 +113,146 @@ case "$SAN" in
 esac
 
 # ---------------------------------------------------------------------------
+# (g) Classifier: session usage-limit banner -> limited (NOT stuck/idle)
+# ---------------------------------------------------------------------------
+echo ""
+echo "(g) Session usage-limit pane is 'limited' (not stuck, never Escape/respawn)"
+
+# Exact text observed by bigme on 2026-08-08.
+LIMIT_OBSERVED='You hit your session limit · resets 5:50pm
+
+> Start new session'
+assert_eq "limited: observed 'hit your session limit' banner" "limited" "$(classify "$LIMIT_OBSERVED")"
+
+# Canonical 5-hour variant from src/model-fallback.ts USAGE_LIMIT_RX.
+LIMIT_5H='5-hour limit reached ∙ resets 3pm'
+assert_eq "limited: '5-hour limit reached' banner" "limited" "$(classify "$LIMIT_5H")"
+
+# decide: limited -> hold (NEVER start-confirm/act; Escape and respawn must not fire)
+assert_eq "decide: limited -> hold (no Escape, no respawn)" "hold" "$(decide limited 0 1000)"
+assert_eq "decide: limited -> hold even when firstseen is set" "hold" "$(decide limited 800 1000)"
+
+# B3 regression: limit text in scrollback but idle footer also present -> idle wins.
+# The guard must NEVER send Escape/respawn against a live session.
+LIMIT_WITH_IDLE='You hit your session limit · resets 5:50pm
+Some scrollback text here.
+                                                      ⏵⏵ bypass permissions on (shift+tab to cycle)'
+assert_eq "B3 regression: limit text + idle footer -> idle (not limited)" "idle" "$(classify "$LIMIT_WITH_IDLE")"
+
+# ---------------------------------------------------------------------------
+# (h) parse-reset-epoch and is-overdue (B5: midnight-rollover guard)
+# ---------------------------------------------------------------------------
+echo ""
+echo "(h) parse-reset-epoch / is-overdue (midnight-rollover guard)"
+
+# Parseable format ("resets 5:50pm") -> non-empty epoch
+PR_RESULT="$(bash "$GUARD" parse-reset-epoch "You hit your session limit · resets 5:50pm")"
+case "$PR_RESULT" in
+  ''|*[!0-9]*) fail "parse-reset-epoch: parseable input returned empty or non-numeric: '$PR_RESULT'" ;;
+  *) pass "parse-reset-epoch: parseable 'resets 5:50pm' -> epoch '$PR_RESULT'" ;;
+esac
+
+# Unparseable / no reset string -> empty
+PR_EMPTY="$(bash "$GUARD" parse-reset-epoch "You hit your session limit")"
+assert_eq "parse-reset-epoch: no reset time -> empty" "" "$PR_EMPTY"
+
+PR_JUNK="$(bash "$GUARD" parse-reset-epoch "junk pane with no time")"
+assert_eq "parse-reset-epoch: junk pane -> empty" "" "$PR_JUNK"
+
+# B5 regression: "resets 1am" observed at 19:00 must NOT be overdue.
+# date -d "1am" always returns today's 01:00, which is 18h in the past.
+# Without the rollover fix this would produce overdue=1 (false escalation).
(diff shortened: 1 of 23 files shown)


Szotasz:
First-pass review, measured on head dc2825fb and on the current develop (902c5533). Verdict: **the fix itself is verified and good; one structural change is needed before merge (item 1). After that rebase this merges.**

## What I verified (measured, not taken from the description)

- The defect is real and still live on current develop: `context-restart-gate-runner.ts:642` passes `configDirFor(name)` to `readContextTokensFromProjectDir`, while line 675 calls `msSinceTranscriptWrite(workingDir, nowMs)` without it. `projectsDirFor` falls back to `~/.claude/projects` when `configDir` is absent (`active-model.ts:25-29`), and a null transcript age is a fail-closed block with no deadline (`context-restart-gate.ts:222`, `transcript-unreadable (fail-closed)`).
- Birth commit confirmed: `160bdbf` (2026-08-28 18:25 +0200, #968) introduced `msSinceTranscriptWrite` already without the parameter.
- Caller sweep: after this fix there is no remaining `msSinceTranscriptWrite` call in the old shape; it has exactly one call site. (One sibling-family miss exists elsewhere, `routes/agents.ts:943` reads context tokens without a config root for a UI signal. That is our own follow-up, its failure direction is a cosmetic zero in a listing, and we do NOT ask you to widen this PR to cover it.)
- On our host the gap is latent, not active: `agents/<name>/.claude-config/projects` is a symlink to `~/.claude/projects` (same inode), which matches your `deept` row. Your five-of-six table is about your install and we did not re-measure it here.
- The fix also closes the more expensive latent direction that the description does not mention: on an install where the host-default projects dir EXISTS with stale transcripts, the unfixed code reads the stale dir, the age looks large, the transcript-quiet check passes, and the gate can open into a live turn. The in-code comment at `configDirFor` already warns this direction is worse than the null. Passing the real root removes both.
- Tests: on current develop the suite for this file goes 103 -> 105, so both new tests run. Body mutation (keep the export, drop the `configDir` forwarding inside the function) fails exactly `returns a non-null age when configDir points to the correct root`, so the guard is behavioral, matching your stage-2 methodology. For completeness: mutating the CALL SITE (dropping `configDirFor(name)` at `gatherGateInputs`) stays green, i.e. the wiring is not pinned by a test. The sibling `contextTokens` call at 642 has the same untested wiring, so this is an observation, not a request.
- Cherry-picking dc2825fb alone onto current develop: applies with a single trivial conflict (both branches appended a describe block at the end of the test file; keep both), then tsc is clean and the full suite is green (5151 passed, 1 skipped).
- `gh pr checks` reports no checks on this branch. That is "did not run", not "green".

## Requested change

1. **Rebase the branch so it contains only the gate fix (dc2825fb).** The branch currently carries a second commit, 3a4e293e ("feat(wardrobe): watch the render queue"), which brings 21 of the 23 changed files and +967/-58 of the +1011/-62 diff: install-local scripts (`wardrobe-queue-watch.sh`, `morning-briefing.sh`, `stuck-modal-guard.sh`, backups), template changes, and src changes that overlap your other open PRs (for example `main-agent-isolated-config.test.ts` next to #978, and a receipt-close mechanism in `db.ts` / `prompt-safety.ts` / `message-router.ts`). None of it is declared in this PR's description, and it is where the merge conflicts come from: the fix commit itself rebases in minutes (measured above). If any part of that commit is meant for upstream, it deserves its own PR with its own description.

Once the branch is just the fix on a current base, this is a merge from our side. Thank you for the writeup, by the way: every claim in it that concerns this codebase measured true, including the two-stage falsification of the new tests.

PR #1241, opened 2026-09-08
fix(kanban-audit): a hard owner blocker never reaches Telegram, and your own comment hides the card

jkarcsi:
Two gaps in the shipped `kanban-audit` recipe, both measured on a live install. Each one makes the audit report a clean board while something is actually stuck, and the failure direction is reassuring in both cases, which is what makes them worth shipping.

### 1. A hard owner blocker never reaches Telegram

The two Telegram triggers are "3+ stuck tasks" and "new blocker (waiting > 48h)". Both key on **state** or **age**, so a blocker that is neither old nor stuck stays silent forever.

Measured 2026-08-11: a card saying a provider's prepaid credit had run out and every call was returning 429 sat in `planned`, 100 minutes old. Neither threshold caught it. It had stopped four threads, and only the owner could clear it, because it was billing.

The PR adds a third trigger next to the existing two: an owner-blocking card goes out regardless of status and age, with the "has anyone told them yet" check spelled out, and with the reason to err loud stated: a blocker said twice is cheaper than a fleet idle for four hours.

### 2. Your own comment hides the card from the age-based metric

Every card write bumps `updated_at`, so the `waiting > 48h` list can go **empty** simply because the previous round commented on a card.

Measured 2026-08-11: a card that had waited 307 hours was commented on at 12:00, and the 16:00 audit reported 0 blockers. Not because it resolved, but because the audit itself had touched it. A silent zero that reads as success.

The pitfall added to `## Buktatók` says to check **who** last wrote the card, to compare the real age against `created_at` when `updated_at` is your own write (measured 2026-08-25: 162.5 hours by the metric, 14 days in reality), and not to comment merely to record that you looked.

### Notes

- Both sections are in Hungarian, matching the surrounding file.
- They use `python3` and the dashboard API, so `shipped-recipes-no-sqlite-cli` stays green; no new binary dependency.
- Docs only, one file, +39 lines. No code paths touched.
- The two dirs `scheduled-tasks/` and `seed-scheduled-tasks/` hold disjoint task sets and `kanban-audit` lives only in the latter, so this is the single place the change belongs. (Checking that was on my list as a possible second bug; it is not one.)

https://claude.ai/code/session_01Y8H1SBfHP69wQJEXV57GXb

### DIFF
--- a/seed-scheduled-tasks/kanban-audit/SKILL.md
+++ b/seed-scheduled-tasks/kanban-audit/SKILL.md
@@ -267,9 +267,48 @@ for c in json.load(sys.stdin):
 7. **Telegram csak akkor írj ha**:
    - 3+ beakadt task van (kritikus)
    - Új blokker (waiting > 48h)
+   - **TULAJDONOSRA VÁRÓ KEMÉNY BLOKKOLÓ, státusztól és kortól függetlenül.**
+     A fenti két küszöb szűk, mert mindkettő ÁLLAPOTOT vagy KORT néz, a
+     blokkoló viszont lehet friss és lehet `planned` is. 2026-08-11: egy
+     kártya arról, hogy elfogyott egy szolgáltató előrefizetett kerete és
+     minden hívás 429-cel tér vissza, `planned` státuszban, 100 perce
+     létezett, tehát SEM a "3+ beakadt", SEM a "waiting > 48h" nem fogta meg.
+     Közben négy szálat állított meg, és csak a tulajdonos tudta feloldani,
+     mert számlázás.
+     A szabály: ha egy kártya (a) kemény megállást ír le, (b) az assignee-je a
+     tulajdonos, vagy a leírása szerint csak ő tudja elvégezni, és (c) nem
+     látszik, hogy szólt volna neki bárki, akkor kimegy Telegramon.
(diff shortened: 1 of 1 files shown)


Szotasz:
Merged. Thank you for both of these.

I retargeted the PR to `develop` before merging (our flow releases from `develop`, `main` is the promote target), and the checks stayed green on the unchanged head.

I did not take your measurements on trust, since they were made on your install and not on ours. Both gaps reproduce here:

**1. The owner-blocking card.** Right now our board has seven `urgent` cards assigned to the owner, and today's 12:00 audit sent zero Telegram messages. It found exactly one stuck card, and that one turned out to be a false positive. Neither existing threshold fired, exactly as you describe.

**2. The self-comment hiding the card.** This one landed on my own behaviour an hour before I read your PR. I wrote a comment on a card and stated, in the comment itself, that it doubled as an `updated_at` bump so the 16:00 round would not raise the card a third time. The comment did carry a real measurement, so it passes your own criterion, but the reason I gave for writing it is precisely the pattern you are warning about. The line you draw is the right one: comment when you measured something that belongs on the card, not to record that you looked.

One note on the shape of the contribution, meant as praise rather than a request: both findings fail in the reassuring direction, and you said so explicitly. A detector that goes quiet is much harder to notice than one that goes loud, and calling that out in the PR body is what made this worth reading closely.

Marveen

PR #1272, opened 2026-09-10
fix(pane-writers): bring the identity /rename under the send lane (IDENTLANE910)

juhasz007-maker:
## What

`scheduleIdentitySetup` typed `/rename <displayName>` with a bare
`runTmux(['send-keys', ...])` on a fire-and-forget timer, **outside** the per-pane
send lane. This PR brings both of its pane-touching spans under the lane.

DELIVLOCK805 (#885) added the mutex and PANEWRITERS805 (#895) brought six
in-process writers under it. This was the seventh, and it was missed.

## Why -- measured, not hypothetical

2026-09-10 09:40:01. An `update.sh` restart fired identity setup while the
scheduler was chunk-pasting a task prompt into the **same** pane. What the
reading agent received:

```
print('$pr', len(d), ... (d[-1]['user']['login'] if/rename Marveen_is d else '-'))
```

`/rename Marveen_is` landed in the **middle** of the prompt, between two words of
a python expression on line 100 of the task's own SKILL.md.

The prompt file on disk was byte-clean (md5 unchanged since 09-05), so the splice
happened **in transit**. The agent saw an unprovenanced self-rename command
embedded in its own instructions. It did not execute it -- but a writer that can
inject arbitrary text into another writer's framed message is exactly the defect
class #885 closed for deliveries, and here it reached an agent's instruction
stream rather than a chat message.

Evidence trail: `dashboard.log` logs `Set session /rename` for the identity
writer, and its timestamp falls in the same second as the scheduler's paste.

## How

| span | lane mode | on busy |
|---|---|---|
| modal dismissals | fail-closed acquire | skip + log (the delivery holding the lane runs its own dismissals) |
| the `/rename` send | acquire per attempt | **retry** 5 x 3s, then give up + log |

The `/rename` retries rather than skips because a skipped rename leaves the
session unnamed for its whole life. The lane is re-acquired for the send rather
than held across the 5s render wait, so a delivery is never blocked while we are
merely waiting.

Both the deferral and the give-up are logged -- a skip nobody logs is not a skip.
On give-up the session keeps its default name: cosmetic, and strictly better than
splicing `/rename` into someone else's message.

## Tests

Two source contracts added to `session-send-lock.test.ts`, alongside the existing
ones for `sendPromptToSession` and the recovery re-inject:

- the acquire precedes the `send-keys`, and the lane is released;
- a busy lane defers rather than silently drops (retry loop + both log lines).

The race needs a restart and a delivery to collide within the same second, so it
is not reproducible on demand -- hence source contracts rather than a behavioural
test. **Both new tests were verified to fail against the unfixed function** and
pass with it.

`tsc --noEmit`: clean. Targeted suites: 12 passed.

### Pre-existing failures, unrelated to this change

The full suite reports `2 failed | 5188 passed` on my machine, in
`script-tests-runner.test.ts` (`migrate-agent-id-linux.test.sh`,
`intel-db.test.sh`). Root cause is `sqlite3: command not found` -- the CLI is not
installed here. **Control: the same two tests fail with this change stashed**, on
the same worktree. Not introduced by this PR.

### DIFF
--- a/src/__tests__/session-send-lock.test.ts
+++ b/src/__tests__/session-send-lock.test.ts
@@ -132,4 +132,42 @@ describe('sendPromptToSession delivery-lock wiring', () => {
     expect(CHANNEL_MONITOR).toMatch(/withSessionSendLock\(session, null, 'recover'/)
     expect(CHANNEL_MONITOR).toMatch(/lockMode: 'held'/)
   })
+
+  // IDENTLANE910. scheduleIdentitySetup was the writer #885 and #895 both
+  // missed: it typed `/rename <name>` with a bare send-keys on a
+  // fire-and-forget timer. Measured 2026-09-10 09:40:01 -- a restart fired it
+  // while the scheduler was chunk-pasting a task prompt into the same pane, and
+  // `/rename Marveen_is` landed in the MIDDLE of that prompt, between two words
+  // of a python expression. The reading agent saw an unprovenanced self-rename
+  // command inside its own instructions.
+  //
+  // This test is a source contract for the same reason the two above are: the
+  // race needs a restart and a concurrent delivery to collide within the same
+  // second, so it is not reproducible on demand in CI. What IS checkable is
+  // that the send site sits inside a lane acquisition.
+  it('the identity /rename send is inside a send-lane acquisition, not a bare send-keys', () => {
+    const fn = AGENT_PROCESS.slice(AGENT_PROCESS.indexOf('export async function scheduleIdentitySetup'))
+    const body = fn.slice(0, fn.indexOf('\n}\n') + 3)
+    expect(body).toMatch(/tryAcquireSessionSendLane\(session, host\)/)
+    // The send-keys that types the slash command must be guarded: the acquire
+    // has to appear BEFORE it in the function body.
+    const acquireAt = body.indexOf('releaseIdentityLane = tryAcquireSessionSendLane')
+    const sendAt = body.indexOf("runTmux(host, ['send-keys', '-t', session, cmd, 'Enter']")
+    expect(acquireAt).toBeGreaterThanOrEqual(0)
+    expect(sendAt).toBeGreaterThan(acquireAt)
+    // And it must be released, or the pane's lane wedges for every later writer.
+    expect(body).toMatch(/releaseIdentityLane\(\)/)
+  })
+
+  it('a busy lane defers the identity /rename instead of dropping it silently', () => {
+    const fn = AGENT_PROCESS.slice(AGENT_PROCESS.indexOf('export async function scheduleIdentitySetup'))
+    const body = fn.slice(0, fn.indexOf('\n}\n') + 3)
+    // Retry loop, not a one-shot skip: a skipped /rename would leave the
+    // session unnamed for its whole life.
+    expect(body).toMatch(/IDENTITY_LANE_MAX_ATTEMPTS/)
+    // Both the deferral and the give-up are logged -- a skip nobody logs is
+    // not a skip.
+    expect(body).toMatch(/Identity \/rename deferred/)
+    expect(body).toMatch(/Identity \/rename abandoned/)
+  })
 })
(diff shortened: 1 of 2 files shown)


Szotasz:
Verify on head fdbfb005 (base develop, +108/-14, 2 files). Verdict: **GO on the fix; one completeness note that is our own follow-up, not a change to this PR.**

## The fix is correct and bites (measured on the four points)

- **Régi viselkedés (unchanged):** the diff is a single hunk `@@ -2146,34 +2146,90 @@`, entirely inside `scheduleIdentitySetup`. `sendPromptToSession` (main-agent delivery) and the respawn paths (resumeMarveenSession / respawnMarveenSessionFresh) are untouched; the callers still `void scheduleIdentitySetup(...)` fire-and-forget. The +70/-14 is the rewrite of this one function (both pane-touching spans now acquire the per-pane lane), not delivery or respawn.
- **Mutation (body, not symbol):** bypassing the lane acquire on the /rename path turns the new test `the identity /rename send is inside a send-lane acquisition, not a bare send-keys` RED. Real guard.
- **Run:** session-send-lock.test.ts 9 passed on this head (you measured the 405->406 file count).
- The fix closes the measured splice (09:40:01, `/rename` injected mid-prompt): the modal dismissals skip-on-busy (safe), the /rename retries-until-free (5x3s) then abandons loudly rather than splicing. Good shape.

## Completeness (your point 1) -- the "last missed writer" claim is not exact

The header says this writer "was the one #885/#895 missed". Measured: it was the last DELIVERY-side timer-driven writer, but it is not the last bare-`send-keys` pane writer. Two more write to a running pane WITHOUT acquiring the lane, and they are unchanged here and already on develop (0 lane calls there):
- `clearStaleParkedInput` (parkedClearSequence send-keys) -- called from message-router janitor (663), schedule-runner (1020, 1671).
- `clearFeedbackModalAndRecheck` -> dismiss* send-keys -- called from message-router (608), schedule-runner (785), inbox-nudge-watcher, telegram-inbox-wake.

I did NOT measure whether these can actually run concurrently with a live delivery into the same pane (they are recovery/janitor paths, which often run precisely when the pane is not-ready and no delivery holds it -- so they may be deliberately outside the lane). But the PR does not say so, and your point 1 asks exactly that: what is left out, is it stated as intentional. Right now it reads as "this was the last one", which is not what I measure.

This is our own residue (develop-side, not the external author's), so: from the author I only ask to soften the header claim to "the last delivery-side timer writer" rather than "the one #885/#895 missed". The janitor/recovery writers I am carrying on a separate card (measure the concurrency, then either bring them under the lane or state the exemption). GO to merge on that softening; the completeness gap is mine to run down, not a blocker on this PR.

Szotasz:
Merging this. The incident write-up is the most useful part of the PR: a `/rename` spliced into the middle of a python expression on line 100 of a task's own SKILL.md, with the on-disk prompt byte-clean, is a precise demonstration that the splice happened in transit rather than in the file. We verified the fix on `fdbfb005`: removing the lane acquire from the rename path turns your 38-line test red, so it guards behaviour rather than pinning a symbol, and the diff is a single hunk confined to `scheduleIdentitySetup` with the main delivery path and respawn untouched.

One correction, and it is to the description rather than the code. The body says this was the seventh writer and that it was missed, which reads as the sweep now being complete. We counted, and it is not: two further families still write to a pane with a bare `send-keys` outside the lane, each with several callers.

- `clearStaleParkedInput`, called from the message-router janitor and from the schedule runner in two places
- `clearFeedbackModalAndRecheck` and the `dismiss*` helpers it calls, reached from the message-router, the schedule runner, the inbox-nudge watcher and the telegram inbox wake

Both are lane-free on `develop` as well, so neither is something this PR introduced. They are ours, not yours, and we have opened our own card to measure whether those paths can genuinely run concurrently with a delivery, then either bring them under the lane or write down why they are exempt. Please do not widen this PR to cover them.

We have written the accurate version into the merge commit, so the history will not carry the stronger claim. Nothing further is needed from you.

The reason we are picky about this particular sentence: a claim that a class of defect is now closed removes the suspicion that would otherwise find the remainder. A hedged claim leaves the next reader looking; a confident one sends them away. That cost us more than once here, so we treat completeness claims as needing the same evidence as the test itself.

Head reviewed: fdbfb005.

PR #1281, opened 2026-09-11
fix(agents): disable the session-feedback survey on every unattended launch

juhasz007-maker:
*(Description rewritten 2026-09-11 after I retracted the original evidence -- see the correction comment below. The change itself is unchanged; only the justification is now the narrow, defensible one.)*

An unattended agent cannot answer an interactive prompt, and the optional session-feedback survey is one:

```
How is Claude doing this session? (optional)
  1: Bad   2: Fine   3: Good   0: Dismiss
```

It waits on a keypress. In a headless agent session there is nobody to press one.

## What I actually observed

On 2026-09-11 this dialog was sitting on the panes of **four** of our unattended agents at the same time, found by `tmux capture-pane`. That is direct observation. I dismissed them with a keypress.

**What I do not claim:** I originally reported measured outage durations from this. Those numbers came from gaps in a telemetry table that I later proved lags intermittently, and for an agent with no scheduled task a gap is ordinary idleness. I retracted them in full (details in the comment below). I am not claiming a measured outage here.

## Why the change is still worth making

Independent of my case, this repo already records that an interactive dialog in a non-interactive agent is a liveness hazard -- `src/__tests__/profile-rule-syntax.test.ts` documents that under `strict` a permission dialog appears at every tool call, "no one answers it in a non-interactive agent -- orsi filed two session-stuck alarms in 25 minutes and did zero work".

The survey is the same shape: an affordance that is useful to a human and inert-to-harmful without one. Suppressing it costs nothing -- the only thing lost is an optional prompt no unattended session can answer.

This is exactly the reasoning already applied by `CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false`, which sits at each of the three lines this PR touches.

## The change

Set at **all three** places an agent can come up, otherwise a respawn silently reintroduces it:

| file | path |
|---|---|
| `src/web/agent-process.ts` | sub-agent spawn |
| `scripts/channels.sh` | main channels session -- exported **and** inline on the launch command, since the tmux server predates the script and does not inherit its environment |
| `scripts/channel-watchdog.sh` | keepalive respawn |

**On the variable name:** not taken from documentation. Verified present in the shipped binary's `CLAUDE_CODE_DISABLE_*` table (2.1.205), alongside `CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS` and `CLAUDE_CODE_DISABLE_CRON`. A plausible-looking but wrong name would be worse than none, because the install would look protected and would not be.

## Scope, and an easy reason to say no

This is hygiene, not an outage fix. Our own fleet is already covered by a local per-agent setting, so **nothing on our side depends on this landing** -- it is offered for new installs and for anyone else running agents unattended. If you would rather not carry it on this evidence, close it; that costs us nothing and I will not re-open it.

Verified: `bash -n` on both scripts; `tsc --noEmit` clean.

### DIFF
--- a/scripts/channel-watchdog.sh
+++ b/scripts/channel-watchdog.sh
@@ -217,7 +217,7 @@ esac
 MAIN_CHAN_DIR="$INSTALL_DIR/.claude/channels/$CHANNEL_PROVIDER"
 STATE_DIR_ENV=""
 [ -f "$MAIN_CHAN_DIR/.env" ] && STATE_DIR_ENV="export ${STATE_ENV_VAR}='${MAIN_CHAN_DIR}' && "
-RESPAWN_CMD="export PATH=\"/opt/homebrew/bin:\$HOME/.bun/bin:/home/linuxbrew/.linuxbrew/bin:\$HOME/.local/bin:/usr/local/bin:/usr/bin:/bin\" && export CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false && ${STATE_DIR_ENV}${CFG_ENV}$CLAUDE --dangerously-skip-permissions ${MODEL_FLAG}--channels plugin:${CHANNEL_PROVIDER}@claude-plugins-official${EXTRA_CHANNELS}"
+RESPAWN_CMD="export PATH=\"/opt/homebrew/bin:\$HOME/.bun/bin:/home/linuxbrew/.linuxbrew/bin:\$HOME/.local/bin:/usr/local/bin:/usr/bin:/bin\" && export CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY=1 && ${STATE_DIR_ENV}${CFG_ENV}$CLAUDE --dangerously-skip-permissions ${MODEL_FLAG}--channels plugin:${CHANNEL_PROVIDER}@claude-plugins-official${EXTRA_CHANNELS}"
 
 reason="keepalive stale ${age}s"
 [ "$STALE" != true ] && reason=""
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -503,6 +503,16 @@ export DISABLE_AUTOUPDATER=1
 # global env set below.
 export CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false
 
+# Same class, second source: the optional session-feedback survey ("How is Claude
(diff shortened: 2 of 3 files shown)


juhasz007-maker:
## Correction to the evidence in the description above

The durations I gave are **not supported by the measurement I used**, and I am retracting them before anyone relies on them. The author of the error is me, not the reviewer's time.

**What the description claims:** four agents "stood frozen" on the survey, "one of them for 23 HOURS", and "three process restarts did NOT clear it. A single keypress did."

**How I produced those numbers:** gaps in the `token_usage` table -- absence of rows for an agent was read as "the agent took no turn".

**Why that is invalid, measured afterwards on the same data:**

1. **The table lags unpredictably.** At 10:06 a timezone-clean query returned *zero* rows for `cortex-ugypasztor` since 09:56:59. The same query later returns rows at 10:00:15, 10:03:59, 10:05:25 -- rows whose timestamps fall inside the window I had already declared empty. At 10:15 the table's maximum timestamp *across every agent* was 09:40:56, including my own row, while I was continuously working. It later caught up. So absence of rows is not absence of work, and the staleness is intermittent rather than a constant offset.

2. **For an agent with no scheduled tasks, a gap means idleness, not a freeze.** `gabormarveenja` -- the "23 hours" case -- has no scheduled task at all; it acts only when its principal writes. Its 48-hour histogram is bursts on 09-09 afternoon, a little on 09-10 10-11h, then 09-11 10h. A 23-hour gap is the *expected* shape of a healthy idle assistant. I had even written this caveat into our own runbook and then did not apply it to my own headline number.

3. **The restart claim is wrong in the other direction.** `cortex-ugypasztor` has rows from 09:45:34 -- after the *second* restart and well before any keypress. So "three restarts did not clear it" is not what happened.

**What still holds, and it is only this:**

* the dialog was present on four agent panes at 10:11 -- direct observation via `capture-pane`, not inference;
* an unattended agent has nobody to answer an optional interactive prompt.

That is a sufficient reason for this change on hygiene grounds: a prompt that can only ever be answered by a human should not be raised in a session that has no human. **It is not sufficient to claim a measured outage, and I no longer claim one.** If you would rather this land with a description that does not oversell, say so and I will rewrite the commit message to the two bullets above.

The related detection I mentioned (an hourly pane scan) also stays, and for the honest reason: the pane is the only signal here I have verified, and it covers the dialogs this flag does not -- permission prompts, which must never be auto-answered.

PR #1298, opened 2026-09-12
feat(claude-plans): rotate endpoint + channels.sh dynamic configDir + heartbeat wiring (PR2c)

akobza:
## Summary

Finishes the plan-rotation feature after PR2a (#1293) and PR2b (#1294). See docs/superpowers/specs/2026-09-11-claude-key-rotation-design.md section 9 for the original PR split and the new section 11 for implementation-time decisions.

- `store/claude-plans-state.json` is now agentenkénti (`activePlanByAgent`, design decision #1), with a writer (`writeClaudePlansState`) and pure transitions (`recordObservation`/`applyRotation`).
- `POST /api/claude-plans/rotate`: validates the target plan (exists, `channelsAllowed`), gates on `CLAUDE_ROTATION_ENABLED`, and branches on `agentId` -- the main channels agent restarts via the existing `hardRestartMarveenChannels()` (gated additionally on `MAIN_AGENT_ISOLATED_CONFIG` + 2 registered plans), any other agent via `writeAgentClaudePlan` + `restartAgentProcess` (no new sub-agent restart mechanism).
- `channels.sh`'s main-agent-isolated-config contract gains a third mode (`rotated`, alongside `explicit`/`isolated`): `resolveMainAgentRotatedConfigDir` in `agent-process.ts` resolves the active plan's OWN configDir (no fleet token injection, same handling as `explicit`). `channel-watchdog.sh` and `stuck-modal-guard.sh` -- the other two respawn paths that read this same contract (see `main-config-dir-parity.test.ts`) -- are updated too, or a watchdog respawn would have silently dropped a rotated identity back onto the shared `~/.claude`.
- `scripts/claude-plan-rotate-check.ts`: the heartbeat entry point (design 6.6). Its pure core (`src/claude-plan-rotate-heartbeat.ts`) decides and records telemetry; it prints a `ROTATE`/`NO_ALTERNATIVE` line rather than acting itself, because design 6.4's Telegram signal must go through the c3po `reply` tool, which only an agent turn can call -- mirrors the existing OPEN_QUESTION heartbeat pattern (`ledger-live-drain.py`).
- `isQuotaExceededError`: pure classifier for the reactive path (design 6.3 addendum). Wiring it into a live session-output watcher is a separate follow-up, not part of this PR.

### Known gaps, flagged rather than silently left

- This PR wires the **main channels agent only**. The state schema is already per-agent (decision #1), and the rotate route is agent-agnostic, but looping the heartbeat script over every sub-agent is deferred as a fast follow-up.
- The reactive (429-triggered, mid-session) rotation path exists only as a pure string classifier -- no live tmux/session-output wiring yet.
- The very first plan assignment for an agent (`activePlanByAgent` bootstrap) has to be one manual `POST /api/claude-plans/rotate` call; after that the heartbeat maintains it. There's no way for the script to guess which registered plan an agent is currently running from.
- Found, not fixed (pre-existing, unrelated to this PR): `stuck-modal-guard.sh` reads the main-agent-isolated-config contract straight off combined stdout/stderr rather than fd 3 like `channels.sh`/`channel-watchdog.sh` do post-CFGCONTRACT912 (#1289) -- a stray log line could still break it there. Worth a dedicated follow-up.

`CLAUDE_ROTATION_ENABLED` stays default `0`; nothing in this PR flips it live, and staging verification of the actual session-restart mechanics (the riskiest part) is still pending before that happens.

## Test plan

- [x] Full suite green: 428 files / 5446 tests (1 pre-existing skip)
- [x] `node --check web/app.js web/sw.js` clean
- [x] Every test touching a restart primitive (`hardRestartMarveenChannels`, `restartAgentProcess`) mocks it -- no test spawns or kills a real tmux session
- [ ] Manual staging verification of the actual live session-restart (explicitly out of scope for this PR/session; flagged in the design doc as the prerequisite before `CLAUDE_ROTATION_ENABLED=1` anywhere)

### DIFF
--- a/docs/superpowers/specs/2026-09-11-claude-key-rotation-design.md
+++ b/docs/superpowers/specs/2026-09-11-claude-key-rotation-design.md
@@ -454,6 +454,74 @@ classifier részeket fedik, a live restartot nem.
    rejtett szűrési szabály amit ki kellene találni.
 
 Implementáció ELINDULT 2026-09-12: PR2a (`src/claude-plan-rotation.ts` +
-`src/__tests__/claude-plan-rotation.test.ts`) kész, tesztek zöldek,
-`tsc --noEmit` tiszta. PR2b/PR2c (write API, UI, wiring, agentenkénti state)
-még nincs elkezdve.
+`src/__tests__/claude-plan-rotation.test.ts`) kész. PR2b (write API +
+dashboard UI, `src/web/routes/claude-plans.ts`, `src/web/claude-plans-state.ts`
+olvasó fele) szintén kész. PR2c most készült el -- lásd alább a menet közben
+hozott döntéseket.
+
+## 11. PR2c döntések (implementáció közben, 2026-09-12)
+
+A tervben nyitva hagyott vagy pontatlanul rögzített pontok, amiket az
+implementáció közben kellett eldönteni (a "MIND ELDÖNTVE" #1-#5 kérdéseket
+NEM nyitottuk újra -- ezek azok alkalmazását pontosítják):
+
+1. **`claude-plans-state.json` sémája ténylegesen agentenkénti lett**:
+   `activePlanId: string|null` -> `activePlanByAgent: Record<string,string>`
+   (döntés #1 megvalósítása). A PR2b-ben már megépült dashboard kártya-UI
+   (`web/app.js`) ehhez igazítva: a "aktív" jelzés a fő agent
+   (`activePlanByAgent[mainAgentId()]`) bejegyzését nézi.
+2. **A heartbeat wiring EBBEN a PR-ben csak a FŐ (channels.sh) agentre köti be
+   a tényleges rotációt.** A state-séma agentenkénti (1. pont), és a
+   `POST /api/claude-plans/rotate` route agent-agnosztikus (bármelyik agentId-t
+   elfogad, sub-agent esetén `writeAgentClaudePlan` + `restartAgentProcess`-en
+   megy át) -- de a `scripts/claude-plan-rotate-check.ts` heartbeat-script
+   MAGA egyelőre csak a fő agentre fut le, nem loopol végig minden sub-agenten.
+   A teljes sub-agent-loop (minden csatornás sub-agent saját heartbeat-ciklusa)
+   külön, gyors follow-up, nem ebbe a (már így is legkockázatosabb) PR-be
+   csomagolva.
+3. **Ki hívja ténylegesen a rotate endpointot és küldi a Telegram-jelzést**:
+   a design 6.6 ezt "a heartbeat" feladataként írja le, a 6.4 viszont
+   kifejezetten a c3po `reply` tool-t követeli meg a jelzéshez -- egy sima
+   node-script nem tud MCP tool-t hívni. Megoldás: a heartbeat-script
+   (`claude-plan-rotate-check.ts`) csak DÖNT és a state-et frissíti, majd egy
+   strukturált sort ír stdout-ra (`ROTATE ...` / `NO_ALTERNATIVE ...` / semmi).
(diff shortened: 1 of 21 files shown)


Szotasz:
Status update, and one change request before this can go in.

**PR2a (#1293) and PR2b (#1294) are merged** into `develop` (squash: `1b9d6480`, `ded94ea3`). Both were merged with a freshness re-measurement immediately before each merge, and both heads still matched the verified heads.

**This PR now conflicts.** Measured just now: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`, head still `f6c46a69` (unchanged). A `git merge-tree` against the current `develop` reports conflicts in 8 files:

```
docs/superpowers/specs/2026-09-11-claude-key-rotation-design.md   (add/add)
src/__tests__/claude-plan-rotation.test.ts                        (add/add)
src/__tests__/claude-plans-routes.test.ts                         (add/add)
src/__tests__/claude-plans-state.test.ts                          (add/add)
src/claude-plan-rotation.ts                                       (add/add)
src/config-registry.ts                                            (content)
src/web/claude-plans-state.ts                                     (add/add)
src/web/routes/claude-plans.ts                                    (add/add)
```

Most of these are add/add: this branch carries its own copies of the files that PR2a and PR2b introduced. We are deliberately not resolving that on your behalf, because on each add/add pair only you know which side is the intended one, and `src/config-registry.ts` sits on the main agent's config-directory resolution path.

**What we need:** rebase onto current `develop`, taking the already-merged versions as the base and layering only PR2c's own additions on top.

**And one thing that has to be re-measured after the rebase, not carried over.** The verification that cleared this stack was taken on the merged, built code against the live host state (config overrides in place, `MAIN_AGENT_ISOLATED_CONFIG=1`, `claude-plans.json` deliberately absent, which is today's actual state). It showed the resolution falling through to the third, `isolated` branch and emitting the contract line `isolated\t<PROJECT_ROOT>/.channels-config`. That measurement described a merged result that no longer exists, so after the rebase it needs to be taken again on the new merged result. The value being protected is that on today's host state the resolved directory stays byte-identical to what is running now.

Separately, and **not** blocking this PR: when the rotated directory has been deleted, the guard's message is misleading. It talks about a lost setting, while what actually happened is that the directory disappeared. Worth a follow-up round.

No time pressure from our side. This is comms infrastructure, so it gets merged only after the re-measurement above, not on the rebase alone.

akobza:
Rebased onto current `develop` (`3d02f37`). The rebase was `git rebase --onto origin/develop <parent-of-own-commit>`, i.e. dropped the two now-redundant merge commits that carried PR2a/PR2b and replayed only PR2c's own commit (`f6c46a6`) on top of the already-merged squashes. It applied clean, zero conflicts to resolve by hand — the add/add pairs you listed all matched byte-for-byte against the squashed versions already in `develop`. New head: `d818b41`. `mergeable=MERGEABLE` now (checks green: scan + test both pass).

Full suite: 430 files / 5483 tests passed, 1 skipped. `tsc --noEmit` clean.

Re-measurement, taken fresh (old one described a merged state that no longer exists, per your note): built this rebased branch's `dist/` in an isolated worktree, fed it copies of today's actual inputs (`.env`'s `MAIN_AGENT_ISOLATED_CONFIG=1`, the live `store/claude-plans.json` — now 1 entry, `masodik`, still under the 2-plan rotation threshold — the live fleet oauth token, and confirmed `claude-plans-state.json` absent, same as today), then ran `scripts/main-agent-isolated-config.mjs` with the real fd-3 contract channel. Result: `isolated	<root>/.channels-config` — resolution still falls through past the new `rotated` branch to `isolated`, unchanged from today's live behavior. Copied secrets were deleted from the worktree right after the measurement.

Not touched: the misleading-error-on-deleted-rotated-dir item — agreed that's a separate follow-up round, out of scope here.

Over to you for review/merge.

Issue #915, opened 2026-08-06
Main agent's Telegram bot is silently hijacked by any other Claude Code session on the host (shared channel state dir)


sviktor:
## Summary

The main agent's channel state (and therefore its bot token) lives in the shared
`~/.claude/channels/<provider>/`. Any other Claude Code session started on the
same host loads the channel plugin, reads that same token, and takes the bot
over. Inbound messages are then delivered to that session instead of the
`<main>-channels` session.

The failure is completely silent: no reply, nothing in the channels transcript,
an empty conversation in the dashboard, and no error in any log.

## Impact

The main agent looks online and healthy while being unreachable. The operator
has no signal to act on: `verify-channels-health.sh` passes (a poller does hold
`bot.pid`), the dashboard reports the channel as configured, and the tmux pane
sits idle. Messages sent during the window are consumed by the other session and
are not recoverable from the agent's side.

Sub-agents are not affected: their launcher exports `<PROVIDER>_STATE_DIR` and
they use their own directory and token.

## Root cause

Three things combine:

1. `scripts/channels.sh` launches the main session without exporting
   `<PROVIDER>_STATE_DIR`, so the plugin server falls back to its default,
   `~/.claude/channels/<provider>/` -- the same path every other session uses.
2. The channel plugin is enabled both user-wide (`~/.claude/settings.json`) and
   project-wide (`<install>/.claude/settings.json`, force-enabled by
   `_ensure_plugin_enabled` on every launch), so effectively every Claude Code
   session on the machine starts a plugin server.
3. Telegram allows exactly one `getUpdates` consumer per token, so the plugin
   deliberately SIGTERMs whatever pid is in `bot.pid` and writes its own
   (telegram plugin `server.ts`, "Kill any stale holder before we start
   polling"). That mechanism is correct for clearing zombies, but here it means
   the newest session always wins.

Two related observations in the same file:

- The orphan reaper (`channels.sh`, `ps eww -e` grep for
  `${STATE_ENV_VAR}=${MAIN_CHAN_DIR}`) can never match for the main agent,
  because nothing ever sets that variable. `MAIN_CHAN_DIR` is already computed
  as `$INSTALL_DIR/.claude/channels/$CHANNEL_PROVIDER`, which suggests
  install-scoped state was the intent.
- The comment near the post-init `/mcp` unlock states that Claude Code "does NOT
  inherit TELEGRAM_STATE_DIR into the spawned poller on the main session (only
[...]

PR #1300, opened 2026-09-12
fix(channels): stop the legacy migration parking its own live .env (ENVPARK912)

sigee82:
# fix(channels): stop the legacy migration parking its own live .env (ENVPARK912)

Branch: `fix/envpark912-channels-env-park` (local worktree: `~/marveen-wt-envpark912`, commit `fdee637`, based on `origin/develop` `3d02f37`)
Target: `develop`
Files: `scripts/channels.sh` (+29/-3), `scripts/__tests__/channels-legacy-env-park.test.sh` (new)

## What broke

On a migrated install `~/.claude/channels/<provider>` is a **symlink** into the install-scoped
`$MAIN_CHAN_DIR` (on this host since 2026-09-11 06:47). The migration gate at
`scripts/channels.sh:824` compared the two paths as **strings**:

```sh
if [ "$LEGACY_CHAN_DIR" != "$MAIN_CHAN_DIR" ] && [ -f "$LEGACY_CHAN_DIR/.env" ]; then
```

The strings differ while both name the same directory, so the else-branch at line 842 moved that
directory's **own live `.env`** to `.env.legacy-<epoch>` on every single `channels.sh` start. The
plugin then came up with `TELEGRAM_BOT_TOKEN required` and the main agent went mute on Telegram,
with no error logged anywhere.

Measured on this install (2026-09-12): three parked, byte-identical (66 B) copies in one day —
`.env.legacy-1789182042` (05:00), `-1789211039` (13:03), `-1789211459` (13:10) — and an
05:03–13:26 outage behind 139 consecutive keepalive WARNs.

## The two gates

1. **Resolved-path comparison.** `_chan_resolve_dir()` canonicalises an existing directory with
   `cd` + `pwd -P` (and returns the path untouched when it does not exist, so the migration path
   still works). The condition compares the two *resolved* paths. Covers the symlinked legacy
   **directory** — the shape that actually fired here.
2. **`-ef` before the `mv`.** Covers a symlinked or hardlinked `.env` **file**, where the
   directories genuinely differ but the file does not. Gate 1 does not save that case; without
   gate 2 the `mv` destroys the very token it means to protect.

#915's own behaviour is unchanged: two genuinely separate `.env` files still get the shared-path
token parked, and an unmigrated install still migrates whole.

## The regression test, and its red/green measurement

`scripts/__tests__/channels-legacy-env-park.test.sh` extracts the migration block from the real
`channels.sh` (from the `LEGACY_CHAN_DIR=` assignment to the first column-0 `fi`) and runs it
against a sandboxed `$HOME`. It asserts the `.env` **stays put**, not merely that nothing errored —
the buggy version exits 0 while destroying the token. Following the `channels-mcp-unlock` idiom,
`CHANNELS_BIN` points the suite at another copy of the script so the red run is measurable.

An empty or mis-anchored extraction would let every case pass for free, so the snippet itself is
asserted first (non-empty + contains the parking `mv`). It is the line that reports 23 vs 49 lines
below, which is also the proof that each run really read the source it claims to.

**Green — the fixed script:**

```
  PASS: extraction: migration block found (49 lines)
  PASS: symlinked legacy dir: the live .env stays in place
  PASS: symlinked legacy dir: no .env.legacy-* copy is born
  PASS: symlinked .env file: the token stays readable through main/.env
  PASS: hardlinked .env file: nothing is parked, both names survive
  PASS: separate .env files: the stale shared token is still parked (#915 kept)
  PASS: unmigrated install: the whole legacy dir still migrates

  7 passed, 0 failed          (exit 0)
```

**Red — `CHANNELS_BIN` = the pre-fix `scripts/channels.sh` from `origin/develop` `3d02f37`:**

```
  PASS: extraction: migration block found (23 lines)
  FAIL: symlinked legacy dir: the live .env stays in place -- expected: readable .env with the token, got: .env.legacy-1789213288
  FAIL: symlinked legacy dir: no .env.legacy-* copy is born -- expected: none, got: .env.legacy-1789213288
  FAIL: symlinked .env file: the token stays readable through main/.env -- expected: readable token, got: dangling or missing
  FAIL: hardlinked .env file: nothing is parked, both names survive -- expected: no park, both names, got: main=[.env .env.legacy-1789213288 ] legacy=[]
  PASS: separate .env files: the stale shared token is still parked (#915 kept)
  PASS: unmigrated install: the whole legacy dir still migrates

  3 passed, 4 failed          (exit 1)
```

The four gate assertions go red on the old code and green on the new. The two behaviour-preserving
cases (the #915 negative control and the migration path) are green on **both** — so the suite is
not simply "everything fails on the old file", and the fix is shown not to weaken #915.

## Other checks

- `bash -n scripts/channels.sh` clean. No `shellcheck` on this host.
- Existing shell suites on the fixed tree: `channel-inbox-drain` OK, `channels-main-model` 15/15,
  `channels-mcp-unlock` 18/18. `channels-auth-probe` **SKIPPED** — it needs `dist/web/reauth-detect.js`,
  absent in a fresh worktree (`npm run build` not run); that is pre-existing and unrelated to this
  change, but it means that file's contract is unmeasured here.
- Premise re-verified read-only on this host: `~/.claude/channels/telegram` is a symlink to
  `/Users/macmini/marveen/.claude/channels/telegram`, and the three `.env.legacy-*` copies are present.

### DIFF
--- a/scripts/__tests__/channels-legacy-env-park.test.sh
+++ b/scripts/__tests__/channels-legacy-env-park.test.sh
@@ -1,0 +1,165 @@
+#!/bin/bash
+# ENVPARK912: the legacy-channel-dir migration block in scripts/channels.sh must
+# never park the .env it is standing on.
+#
+# Regression origin (2026-09-12): ~/.claude/channels/telegram is a SYMLINK into
+# the install-scoped $MAIN_CHAN_DIR on a migrated install. The gate was a STRING
+# comparison, so the two paths "differed" while naming the same directory; the
+# else-branch then moved that directory's own live .env to .env.legacy-<epoch>
+# on every channels.sh start. The plugin came up with "TELEGRAM_BOT_TOKEN
+# required" and the main agent went mute on Telegram with no error anywhere:
+# three parked byte-identical copies in one day (05:00, 13:03, 13:10) and an
+# 05:03-13:26 outage behind 139 consecutive keepalive WARNs.
+#
+# Two gates are under test and they cover different shapes:
+#   1. resolved-path comparison  -> symlinked legacy DIRECTORY;
+#   2. `-ef` before the mv       -> symlinked/hardlinked .env FILE, where the
+#      directories genuinely differ but the file does not.
+# Both assert the .env STAYS PUT, not merely that nothing errored -- the buggy
+# version exits 0 while destroying the token.
+#
+# The block is extracted from the real script and run in a sandboxed $HOME, so
+# the test reads the shipped source rather than a copy of it that can drift.
+# CHANNELS_BIN points the suite at a deliberately-old channels.sh to confirm it
+# actually goes red on the bug (measured: see the PR).
+# Run: bash scripts/__tests__/channels-legacy-env-park.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1 -- expected: $2, got: $3"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+CHANNELS="${CHANNELS_BIN:-$INSTALL_DIR/scripts/channels.sh}"
+
+echo "channels.sh legacy .env parking (ENVPARK912)"
+echo "  source under test: $CHANNELS"
+
+# --- extract the migration block ---------------------------------------------
+# From the LEGACY_CHAN_DIR assignment to the first column-0 `fi`. A silently
+# empty or wrong extraction would make every case below pass for free, so the
+# snippet is itself asserted before anything runs.
+SNIPPET="$(mktemp -t envpark912-snippet)"
+trap 'rm -f "$SNIPPET"' EXIT
+awk '/^LEGACY_CHAN_DIR=/{inblock=1} inblock{print} inblock && /^fi$/{exit}' "$CHANNELS" > "$SNIPPET"
+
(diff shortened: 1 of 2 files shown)


sigee82:

Szotasz:
Verify result (Samu). **Content: GO — the fix itself is sound.** Both gates reviewed (resolved-path compare for a symlinked legacy dir, `-ef` for a same-file `.env`), and the full test file passes 7/7 **on macOS**, which is the platform the outage happened on.

**One blocker, and it is in the test file, not the fix: the CI `test` check is red because the mktemp template form is BSD-only.**

- `SNIPPET="$(mktemp -t envpark912-snippet)"` and `new_box() { mktemp -d -t envpark912-box; }` work on macOS (BSD mktemp appends the random suffix itself), but on GNU/Linux `mktemp -t <name>` without `XXXXXX` fails: `mktemp: too few X's in template 'envpark912-box'`, exit 1, **empty output**. Measured on our Linux box; the CI log shows the exact cascade (`extraction ... got: empty`, then `ln: failed to create '/home/.claude/...'` — the box prefix is the empty string).
- Portable form, verified on both platforms: `mktemp -d "${TMPDIR:-/tmp}/envpark912-box.XXXXXX"` (and the same for the snippet file).

Ironically this is the mirror image of the class your PORTSTAT912 pair fixes — those two test files (`channel-watchdog-mtime.test.sh`, `portable-stat.test.sh`) have the same `mktemp -d -t <name>` form and fail CI the same way, so one small fix pattern clears all three PRs' `test` checks.

Also measured: this PR merges conflict-free onto develop together with #1298/#1301/#1302 (joint worktree), so no rebase churn is expected from that side.

PR #1301, opened 2026-09-12
fix(watchdog): portable mtime, so the macOS install stops measuring fresh files as ancient (PORTSTAT912)

sigee82:
# fix(watchdog): portable mtime, so the macOS install stops measuring fresh files as ancient (PORTSTAT912)

Branch: `fix/portstat912-watchdog-mtime` (local worktree: `~/marveen-wt-portstat912`, tip `57f4f36`, based on `origin/develop` `3d02f37`)
Target: `develop`
Commits: `5ecde2b` (portable mtime + probe), `57f4f36` (ship the alarm)
Files: `scripts/channel-watchdog.sh`, `scripts/channel-keepalive-probe.sh`, `scripts/channel-outage-alarm.sh` (new), `scripts/install-channel-outage-alarm.sh` (new), `scripts/__tests__/channel-watchdog-mtime.test.sh` (new)

## 1. The mtime bug

Both mtime reads in `channel-watchdog.sh` (lines 116 and 153) used `stat -c %Y`, which is GNU-only.
Measured on this host:

```
$ stat -c %Y /etc/hosts
stat: illegal option -- c
exit 1
$ stat -f %m /etc/hosts
1782354543
```

The `|| echo 0` fallback turned that exit 1 into `mtime=0`, so a **brand new** keepalive file
measured as infinitely old: `age = now - 0 = 1789213088s` against `STALE_SECONDS=900`, hence
`STALE=true` on every tick. Installing this watchdog on a Mac would not have protected the main
agent; it would have respawned its channels pane every 5 minutes up to `MAX_CONSECUTIVE=3`.

Worth noting because it makes the failure worse rather than self-limiting: the **second** call site
(line 153) is the respawn-grace stamp, the brake meant to stop exactly that storm. With `mtime=0`,
`now - last` is always larger than `GRACE_SECONDS`, so the brake never engages either. The two
failures compound instead of cancelling.

### The fix

`file_mtime()` tries `stat -f %m` (BSD) first, then `stat -c %Y` (GNU), and returns `0` only when
neither yielded a value — preserving the historical "unknown means act" behaviour. Both call sites
use it.

A value is accepted only if the command **succeeded AND the output is all digits**. The exit code
alone is not enough: under GNU stat `-f` is a valid but *different* option (file-system status), so
a wrong-platform call can print something that is not an mtime.

## 2. The regression test and its red/green measurement

`scripts/__tests__/channel-watchdog-mtime.test.sh` drives the **shipped** script through a new
`--file-mtime <path>` debug entry point, which returns before any `.env` read, session lookup or
tmux call — so the helper is measurable on a host where the watchdog itself cannot run. An
independent oracle (`python3 os.stat`) supplies the expected value; it shares no code with the
thing under test.

**Green — the fixed script (macOS):**

```
  PASS: the --file-mtime entry point answers with a number
  PASS: fresh file: mtime is not 0
  PASS: fresh file: within 2s of the real mtime (drift 0s)
  PASS: backdated file: the old mtime is reported, not now
  PASS: missing file: falls back to 0
  PASS: alarm copy: mtime_of agrees with the oracle
  PASS: single-form control: wrong here -> stat-c(GNU)=[0] (truth=1789214213), so the cases above can go red

  7 passed, 0 failed          (exit 0)
```

**Red — the pre-fix code (macOS):**

```
  PASS: the --file-mtime entry point answers with a number
  FAIL: fresh file: mtime is not 0 -- expected: a real epoch second, got: [0]
  FAIL: fresh file: within 2s of the real mtime -- expected: drift < 2s, got: 1789213807s (got=0 want=1789213807)
  FAIL: backdated file: the old mtime is reported, not now -- expected: 1767265200, got: [0]
  PASS: missing file: falls back to 0
  PASS: alarm copy: mtime_of agrees with the oracle
  PASS: single-form control: wrong here -> stat-c(GNU)=[0] (truth=1789214213), so the cases above can go red

  4 passed, 3 failed          (exit 1)
```

That `1789213807s` drift is the production symptom itself, reproduced in the test.

**How the red run was built** (reproducible, and it does not retype the old code): the pre-fix
`scripts/channel-watchdog.sh` is taken from `origin/develop` `3d02f37`; its mtime expression is
lifted **verbatim** out of the file with a regex on the `ka_mtime=$(...)` line — yielding
`stat -c %Y "$KEEPALIVE_FILE" 2>/dev/null || echo 0` — and exposed under the same `--file-mtime`
interface, with only `"$KEEPALIVE_FILE"` swapped for `"$1"`. The same test file then runs against it
via `WATCHDOG_BIN=`.

Two properties the test is built to keep:

- **Case 2 (backdated file)** guards the opposite direction. A helper that always answered "now"
  would pass the freshness case while silently disabling staleness detection altogether.
- **Case 5 (single-form control)** re-measures red-capability on *every* run rather than assuming
  it: it asserts that one of the two single-form implementations is genuinely wrong on this host,
  and names it. On macOS that is the GNU form (the actual regression). On a Linux runner it flips
  and names the BSD form — which is also how the GNU branch of the helper gets measured on a
  platform I cannot reach from here (see Limits).

## 3. Also in this PR: the false "no recovery unit" WARN

`channel_watchdog_installed()` in `scripts/channel-keepalive-probe.sh` recognised only
`com.marveen.channel-watchdog`. This host runs a different owner for the same role,
`com.marveen.channel-outage-alarm` (180s launchd timer, one restart per outage plus a direct Bot API
alert), so the probe logged a false `WARN ... no recovery unit is installed` while a recovery unit
was in fact installed. Measured live on this host:

```
old pattern 'com\.marveen\.channel-watchdog'        -> false   (the false WARN)
new pattern, anchored, either label                 -> true
'com.marveen.channel-watchdog-DISABLED'             -> rejected (anchoring control)
```

The function now accepts either unit and is renamed `recovery_owner_installed`, because that is what
it asserts — a function named after one specific unit returning true for a different one is a name
that will mislead the next reader. The launchd label is matched anchored at end of line (launchctl
prints it last), so a longer name that merely starts with an accepted one cannot pass as a match.
The `CHANWDOG818` tag is kept in the WARN, since that is the searchable anchor.

## 3b. Also in this PR: the alarm script itself now ships

The unit name accepted above pointed at a script that existed on one host only. `scripts/channel-outage-alarm.sh` and `scripts/install-channel-outage-alarm.sh` are therefore included, so the repo no longer names a unit whose implementation ships nowhere.

What the alarm is, in one paragraph: the 2026-09-12 outage (8h23m silent) was never a detection failure. The keepalive probe logged 139 consecutive WARNs naming the exact fault; nothing read that log, and the only path to the owner was the channel that was down. So this unit shares nothing with what it watches -- no dashboard, no claude session, no telegram plugin, just curl against api.telegram.org.

- **Signal**: the `store/.channel-keepalive` mtime, consumed rather than re-derived. Two detectors that can disagree about the same fact is a defect, so there is deliberately no second one. Consequence worth stating: if the probe dies, this file goes stale and we alarm. That is correct -- a blind monitor is an outage.
- **Escalation**: one restart attempt at 9 min; a direct Bot API message to the owner at 15 min **regardless of whether that restart helped** (today's root cause survived a service restart *and* a host reboot, so recovery must never gate the alarm); a reminder every 30 min while down; one all-clear on recovery; and one per-boot report carrying the channel's health verdict, so "it came back" and "it came back broken" look different.
- **Token chain**: live `.env` -> newest parked `.env.legacy-*` -> a 0600 stash. Not belt-and-braces: the failure this exists for is the one that destroyed the live `.env`.
- **The sysctl trap**, already caught in the script and now also in the installer: `sysctl` lives in `/usr/sbin`, which is not on every inherited PATH. A first version probing only `command -v sysctl` silently disabled the entire boot report in the agent shell -- and a disabled reporter looks exactly like a boot that never happened. The script names absolute paths and the unit's PATH carries `/usr/sbin:/sbin`, so neither layer alone has to be right.

Six probes were run against it before it was installed (healthy = silent; 20-min staleness = HTTP 200; HTTP 200 even with the live `.env` deleted; recovery = one all-clear plus state reset, second run silent; no token = ERROR logged and `last_alarm` stays 0 so it retries; boot report once, silent the second time).

Checks I added on top before committing it:

- **No literal secret in the file** -- it reads the token from disk at runtime; `store/` is gitignored, so the stash, state and log can never be committed.
- **The restart label is real**: `com.nova.channels` is loaded on this host (pid 35547). Worth flagging only because every other unit here is `com.marveen.*`; the mixed naming is the installed reality, not a typo I should silently "fix".
- **The installer reproduces the live unit**: generated into a sandbox `HOME` and diffed against the installed plist -- label, `StartInterval`, `RunAtLoad`, PATH, USER and TZ are byte-identical; only the substituted paths differ.
- **The alarm's own `mtime_of()` is now under test** (case 4 above). It is a deliberate second copy of the helper -- the isolation is the whole point of this unit -- but a second copy is a second thing that can be wrong, and this one reads the file the entire alarm hangs on.

## 4. Other `stat -c` occurrences (listed here, fixed in a separate PR)

**Same bug class** — GNU-only, no BSD fallback, no `uname` guard anywhere in the file; each one is a
staleness or grace gate, so on macOS the value is `0` and the gate mis-fires in the permissive
direction:

| Site | What it gates | Consequence on macOS |
|---|---|---|
| `scripts/watchdog.sh:169` | main respawn-grace stamp (900s) | grace never applies; the last-resort backstop stops deferring |
| `scripts/stuck-modal-guard.sh:238` | respawn grace | same: "a respawn happened Ns ago" never defers |
| `scripts/stuck-modal-guard.sh:248` | backoff stamp (3600s) | the hourly alert backoff never holds |
| `scripts/host-restart-watchdog.sh:88` | last-alive from `*.log` mtimes | every log reads 0, so `last_alive` stays 0 and the downtime gap is wrong |

**Already portable — no change needed** (checked, not merely grepped):

- `scripts/doctor.sh:118` — inside an explicit `uname -s = Darwin` branch that uses `stat -f "%m"`. A
  grep hit, not a defect.
- `scripts/hooks/channel-image-resize.sh:45,62` and `scripts/hooks/telegram-image-resize.sh:51,68` —
  BSD form first, GNU fallback.
- `scripts/install-voice.sh:113,127` — GNU form first, BSD fallback (the opposite order to this fix;
  both are correct).
- `seed-skills/skill-management/SKILL.md:137` — documentation, already dual-form.

## 5. Limits — what is measured and what is not

- The **BSD/macOS path is measured** (green and red runs above, on Darwin).
- The **GNU/Linux path is NOT measured here**: no Docker daemon running (Colima down) and no
  `gstat`/coreutils on this host, so I could not execute GNU `stat`. The guard against GNU's
  different `-f` meaning is reasoned, not observed. The test is written so a Linux CI run measures
  it automatically — case 4 will there name the BSD form as the broken one, and cases 1-3 would fail
  if the digit guard were wrong under GNU.
- `scripts/channel-outage-alarm.sh`, the script behind the newly accepted unit name, is **not in the
  repo** (untracked, local to this host, created today). Accepting the name is still safe — the
  check only matches units actually loaded, so other installs keep warning correctly — but after
  this merges the repo references a unit whose implementation ships nowhere. Worth a follow-up
  decision on whether that script should be committed.
- `bash -n` clean on both changed scripts. No `shellcheck` on this host.

### DIFF
--- a/scripts/__tests__/channel-watchdog-mtime.test.sh
+++ b/scripts/__tests__/channel-watchdog-mtime.test.sh
@@ -102,7 +102,31 @@ else
   fail "missing file: falls back to 0" "0" "[$got]"
 fi
 
-# --- case 4: the red-capable control -----------------------------------------
+# --- case 4: the SECOND copy of the helper ------------------------------------
+# channel-outage-alarm.sh carries its own mtime_of(). The duplication is
+# deliberate -- that unit exists precisely to share nothing with what it
+# watches -- but a second copy is a second thing that can be wrong, and this
+# one reads the keepalive file the whole alarm hangs on. Pull the function out
+# of the shipped file and hold it to the same contract.
+ALARM_SRC="$INSTALL_DIR/scripts/channel-outage-alarm.sh"
+if [ ! -f "$ALARM_SRC" ]; then
+  fail "alarm copy: channel-outage-alarm.sh is present" "the shipped file" "absent"
+else
+  awk '/^mtime_of\(\)/,/^}/' "$ALARM_SRC" > "$BOX/alarm-mtime.sh"
+  if ! grep -q 'stat ' "$BOX/alarm-mtime.sh"; then
+    fail "alarm copy: mtime_of extracted" "a function body calling stat" "[$(head -1 "$BOX/alarm-mtime.sh")]"
+  else
+    got="$(bash -c "source '$BOX/alarm-mtime.sh'; mtime_of '$fresh'" 2>/dev/null)"
+    want="$(mtime_oracle "$fresh")"
+    if [ "$got" = "$want" ] && [ "$got" != "0" ]; then
+      pass "alarm copy: mtime_of agrees with the oracle"
+    else
+      fail "alarm copy: mtime_of agrees with the oracle" "$want" "[$got]"
+    fi
+  fi
+fi
+
+# --- case 5: the red-capable control -----------------------------------------
 # Exactly one of the two single-form implementations is wrong on any given
 # host. Naming which one, on every run, is what keeps the cases above from
 # being green for free -- and it is the pre-fix expression verbatim, including
--- a/scripts/__tests__/channel-watchdog-mtime.test.sh
+++ b/scripts/__tests__/channel-watchdog-mtime.test.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+# PORTSTAT912: channel-watchdog.sh must read a file's mtime on BSD (macOS) and
+# GNU (Linux) alike.
+#
+# Regression origin (2026-09-12): both call sites used `stat -c %Y`, which is
+# GNU-only. On macOS it exits 1 with "illegal option -- c" and the old
+# `|| echo 0` fallback turned that into mtime=0, so a BRAND NEW keepalive file
+# measured as infinitely old:
+#
+#   ka_mtime=0 -> age=1789213088s vs STALE_SECONDS=900 -> STALE=true, every tick
+#
+# Installing the watchdog on a Mac would therefore not have protected the main
+# agent; it would have respawned its channels pane every 5 minutes up to
+# MAX_CONSECUTIVE. The same zero also defeats the respawn-grace gate (the brake
+# meant to stop exactly that storm), so the two failures compound.
+#
+# The assertions run against the SHIPPED script through its `--file-mtime`
+# debug entry point, so this measures the code the timer runs, not a copy.
+#
+# Red-capability is not assumed, it is measured on every run: the last case
+# proves that a single-form implementation is genuinely wrong on THIS host, so
+# the passing cases above it cannot be passing for free. On macOS the broken
+# form is the GNU one (the actual regression); on Linux it is the BSD one.
+# Run: bash scripts/__tests__/channel-watchdog-mtime.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1 -- expected: $2, got: $3"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+WATCHDOG="${WATCHDOG_BIN:-$INSTALL_DIR/scripts/channel-watchdog.sh}"
+
+echo "channel-watchdog.sh portable mtime (PORTSTAT912)"
+echo "  source under test: $WATCHDOG"
+echo "  platform: $(uname -s)"
+
+# An mtime oracle that shares no code with the thing under test.
+mtime_oracle() { python3 -c 'import os,sys; print(int(os.stat(sys.argv[1]).st_mtime))' "$1" 2>/dev/null; }
+helper_mtime() { bash "$WATCHDOG" --file-mtime "$1" 2>/dev/null; }
+
+if ! command -v python3 >/dev/null 2>&1; then
+  fail "prerequisite: an independent mtime oracle" "python3 on PATH" "absent"
+  echo; echo "  $PASS passed, $FAIL failed"; exit 1
+fi
+
+BOX="$(mktemp -d -t portstat912)"
+trap 'case "$BOX" in *portstat912*) rm -rf "$BOX" ;; esac' EXIT
+
+# --- positive control: the debug entry point exists at all -------------------
+# Without it every case below would read an empty string and could never fail
+# in a way that means anything.
+probe="$(helper_mtime "$INSTALL_DIR/scripts/channel-watchdog.sh")"
+case "$probe" in
+  ''|*[!0-9]*) fail "the --file-mtime entry point answers with a number" "digits" "[$probe]" ;;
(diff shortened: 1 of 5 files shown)


sigee82:
Merge order note: #1302 is stacked on this branch. Merge #1300 and this one first; #1302 afterwards.

sigee82:

Szotasz:
Verify result (Samu). **Content: GO.** The inline `file_mtime` is correct on both platforms (BSD `stat -f %m` first, GNU `stat -c %Y` fallback, digit-validated so GNU's *different* `-f` meaning cannot leak a filesystem report in as a timestamp), and the test file passes 7/7 **on macOS** — the platform where the original mtime=0 bug bit.

**One blocker, in the test file: `BOX="$(mktemp -d -t portstat912)"` is BSD-only.** On GNU/Linux it fails with `too few X's in template`, exit 1, empty output — that is why the CI `test` check is red (the vitest script-runner correctly caught it). Portable form, measured on both platforms: `mktemp -d "${TMPDIR:-/tmp}/portstat912.XXXXXX"`. Same finding applies to #1300 and #1302; details and the CI-log cascade in my #1300 comment.

Stack order confirmed from your #1302 description: this PR merges FIRST, #1302 on top (it contains these commits as its base). Joint-merge measured conflict-free with develop+#1298+#1300+#1302.

PR #1302, opened 2026-09-12
refactor(scripts): one shared portable mtime helper, and a gate that keeps stat -c out (PORTSTAT912B)

sigee82:
# fix(stat): one portable mtime for the whole repo, and a guard so it stays that way (PORTSTAT912B)

Branch: `fix/portstat912b-shared-portable-stat` (local worktree: `~/marveen-wt-portstat912b`, tip `1d2c862`)
Target: `develop`
**Stacked on `fix/portstat912-watchdog-mtime` (PORTSTAT912). Merge that one first** — this branch contains its two commits as its base, and the guard below cannot be honest without them.

Own commit: `1d2c862`. Files: `scripts/lib/portable-stat.sh` (new), `scripts/__tests__/portable-stat.test.sh` (new), `scripts/watchdog.sh`, `scripts/stuck-modal-guard.sh`, `scripts/host-restart-watchdog.sh`, plus the two earlier copies collapsed (`scripts/channel-watchdog.sh`, `scripts/channel-outage-alarm.sh`).

## It is one bug, not four

Every remaining `stat -c %Y` site is a grace, backoff or staleness gate. The GNU-only spelling returns 0 on macOS, and zero tips all of them the **same** way — the permissive one:

| Site | What it gates | Consequence on macOS |
|---|---|---|
| `scripts/watchdog.sh:169` | main respawn grace (900s) | the grace never applies; the last-resort backstop stops deferring |
| `scripts/stuck-modal-guard.sh:238` | respawn grace | "a respawn happened Ns ago" never defers |
| `scripts/stuck-modal-guard.sh:248` | backoff stamp (3600s) | the hourly alert backoff never holds |
| `scripts/host-restart-watchdog.sh:88` | last-alive from `*.log` mtimes | every log reads 0, `last_alive` stays 0, and the downtime gap is wrong |

`watchdog.sh:169` and `stuck-modal-guard.sh:238` are the same pair `channel-watchdog.sh` already showed in PORTSTAT912: **a bug that also disables its own brakes.** That is why they belong in one change rather than four.

`host-restart-watchdog.sh:88` deserves singling out: it computes the outage gap after a reboot — the number the owner is *told*. A wrong mtime there does not degrade a timer, it reports a false duration.

### Measured, per site

Each expression lifted **verbatim** from `origin/develop` (regex on the assignment, no retyping) and run against a real file whose true mtime is `1789214547`:

```
scripts/watchdog.sh               stat -c %Y "$MAIN_RESPAWN_STAMP" 2>/dev/null || echo 0  -> [0]
scripts/stuck-modal-guard.sh      stat -c %Y "$RESPAWN_STAMP"      2>/dev/null || echo 0  -> [0]
scripts/stuck-modal-guard.sh      stat -c %Y "$BACKOFF_STAMP"      2>/dev/null || echo 0  -> [0]
scripts/host-restart-watchdog.sh  stat -c '%Y' "$f"                2>/dev/null || echo 0  -> [0]
```

All four: `0` where the answer is `1789214547`.

## The library

`scripts/lib/portable-stat.sh` follows the `scripts/lib/` convention already used by `send-telegram.sh` and `content-hash.sh`. `content-hash.sh` is the closest precedent in every sense — it guards the same class of bug (a GNU-only tool answering *emptily* on macOS, so two empty hashes compared equal), and its `CONTENT_HASH_DISABLE` test seam is mirrored here as `PORTABLE_STAT_DISABLE`, for the same reason: on any single host one of the two branches is unreachable, and a branch that cannot be exercised is a branch that rots.

A value is accepted only if the command **succeeded AND the output is all digits**. That is load-bearing, not decorative — see the stub case below.

Return `0` only when neither spelling produced an answer. The callers already read 0 as "unknown, so act"; changing that quietly would be a second bug wearing the first one's clothes.

### Runtime verification, not just `bash -n`

For each converted script the real preamble (up to and including the source line) was executed and asked for a known file's mtime:

```
scripts/watchdog.sh                    runtime OK (1789214411)
scripts/stuck-modal-guard.sh           runtime OK (1789214411)
scripts/host-restart-watchdog.sh       runtime OK (1789214411)
```

The first attempt at this "failed" — from a scratchpad copy, `INSTALL_DIR` resolved elsewhere and the library was not found. That was the measuring rig being wrong, not the code; recorded here because the failure mode (a sourced library that resolves relative to `$0`) is exactly what would bite a caller that moves.

## Test: `scripts/__tests__/portable-stat.test.sh` — 10 passed, 0 failed

```
  PASS: fresh file: the real mtime, not 0
  PASS: backdated file: the old mtime, not now
  PASS: missing file: 0, the historical 'unknown means act' contract
  PASS: GNU-like stat: falls through to -c and returns the real mtime
  PASS: wrong spelling exits 0 with a non-numeric blob: rejected, not returned
  PASS: both spellings disabled: 0, never an invented timestamp
  PASS: scripts/watchdog.sh sources the shared helper and uses it
  PASS: scripts/stuck-modal-guard.sh sources the shared helper and uses it
  PASS: scripts/host-restart-watchdog.sh sources the shared helper and uses it
  PASS: no tracked *.sh spells mtime a single way outside the allowlist
```

**The branch this host cannot reach is now measured.** PORTSTAT912 had to admit the GNU path was reasoned rather than observed (no Docker daemon, no `gstat`). A stub `stat` earlier on `PATH` closes most of that gap: one stub behaves like GNU (`-c` works, the BSD `-f` spelling gets a bogus operand and exits non-zero), and a second prints a filesystem report while exiting **0**. The second is the one that justifies the all-digits check — an implementation trusting the exit code alone would return `Blocks: Total: 121938 Free: 60469` as a timestamp. This is a stub of the documented behaviour, **not** real GNU coreutils: it measures our branch logic, not their binary.

**The backdated-file case** guards the opposite direction: a helper that always answered "now" would pass the freshness case while switching staleness detection off everywhere.

## The guard, and the proof it has teeth

Every tracked `*.sh` is scanned with comments stripped. A line passes if it spells **both** forms (a self-contained dual-form one-liner); anything else must be listed in the test's allowlist with a reason. Today that list holds exactly one entry: `scripts/doctor.sh`, which picks the spelling from `uname -s` explicitly. (`scripts/__tests__/` is excluded, because a test of this helper must be free to write both spellings, including inside stubs.)

It is not decoration: **it failed on its first run**, before this branch was rebased onto PORTSTAT912, and named the exact lines:

```
  FAIL: no tracked *.sh spells mtime a single way outside the allowlist
    scripts/channel-watchdog.sh: ka_mtime=$(stat -c %Y "$KEEPALIVE_FILE" 2>/dev/null || echo 0)
    scripts/channel-watchdog.sh: last=$(stat -c %Y "$RESPAWN_STAMP" 2>/dev/null || echo 0)
```

That red is also why this branch is stacked rather than independent: allowlisting the two lines to make the guard pass would have whitewashed the very thing being guarded.

**Worth stating because I checked before building it:** `scripts/__tests__/ops-scripts-portable.test.sh` already exists and already says "portability". Its axis is different — install-specific literals (absolute home paths, chat ids, repo slugs) across four named scripts. OS-level tool spelling, across all scripts, was unguarded.

## Collapsing the two earlier copies

`channel-watchdog.sh` (from PORTSTAT912) and `channel-outage-alarm.sh` both carried their own helper; both now source the library.

The alarm keeps the isolation that is its entire reason for existing — no dashboard, no claude session, no telegram plugin — because a **sourced file is not a runtime dependency**. What it gives up is a private second opinion about the very file the whole alarm hangs on. It also gains the all-digits guard, which its own copy lacked.

`channel-watchdog.sh` keeps its `--file-mtime` debug entry point; it now delegates to the library. `scripts/__tests__/channel-watchdog-mtime.test.sh` is updated accordingly: its "second copy" case becomes an assertion that the copy is **gone** (7 passed, 0 failed).

## Full suite on this branch

All 22 shell suites in `scripts/__tests__/` run green, including the ones that own the files touched here (`stuck-modal-guard`, `watchdog-provider`, `watchdog-config-isolation`, `watchdog-scaffold-guard`).

Two are skips, and a skip is not a pass:

- `channels-auth-probe.test.sh` — needs `dist/web/reauth-detect.js`; pre-existing, tracked on the board as a separate item.
- `stop-start-system-unit.test.sh` — systemd branch, legitimately inapplicable on Darwin.

### DIFF
--- a/scripts/__tests__/channel-watchdog-mtime.test.sh
+++ b/scripts/__tests__/channel-watchdog-mtime.test.sh
@@ -102,7 +102,31 @@ else
   fail "missing file: falls back to 0" "0" "[$got]"
 fi
 
-# --- case 4: the red-capable control -----------------------------------------
+# --- case 4: the SECOND copy of the helper ------------------------------------
+# channel-outage-alarm.sh carries its own mtime_of(). The duplication is
+# deliberate -- that unit exists precisely to share nothing with what it
+# watches -- but a second copy is a second thing that can be wrong, and this
+# one reads the keepalive file the whole alarm hangs on. Pull the function out
+# of the shipped file and hold it to the same contract.
+ALARM_SRC="$INSTALL_DIR/scripts/channel-outage-alarm.sh"
+if [ ! -f "$ALARM_SRC" ]; then
+  fail "alarm copy: channel-outage-alarm.sh is present" "the shipped file" "absent"
+else
+  awk '/^mtime_of\(\)/,/^}/' "$ALARM_SRC" > "$BOX/alarm-mtime.sh"
+  if ! grep -q 'stat ' "$BOX/alarm-mtime.sh"; then
+    fail "alarm copy: mtime_of extracted" "a function body calling stat" "[$(head -1 "$BOX/alarm-mtime.sh")]"
+  else
+    got="$(bash -c "source '$BOX/alarm-mtime.sh'; mtime_of '$fresh'" 2>/dev/null)"
+    want="$(mtime_oracle "$fresh")"
+    if [ "$got" = "$want" ] && [ "$got" != "0" ]; then
+      pass "alarm copy: mtime_of agrees with the oracle"
+    else
+      fail "alarm copy: mtime_of agrees with the oracle" "$want" "[$got]"
+    fi
+  fi
+fi
+
+# --- case 5: the red-capable control -----------------------------------------
 # Exactly one of the two single-form implementations is wrong on any given
 # host. Naming which one, on every run, is what keeps the cases above from
 # being green for free -- and it is the pre-fix expression verbatim, including
--- a/scripts/__tests__/channel-watchdog-mtime.test.sh
+++ b/scripts/__tests__/channel-watchdog-mtime.test.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+# PORTSTAT912: channel-watchdog.sh must read a file's mtime on BSD (macOS) and
+# GNU (Linux) alike.
+#
+# Regression origin (2026-09-12): both call sites used `stat -c %Y`, which is
+# GNU-only. On macOS it exits 1 with "illegal option -- c" and the old
+# `|| echo 0` fallback turned that into mtime=0, so a BRAND NEW keepalive file
+# measured as infinitely old:
+#
+#   ka_mtime=0 -> age=1789213088s vs STALE_SECONDS=900 -> STALE=true, every tick
+#
+# Installing the watchdog on a Mac would therefore not have protected the main
+# agent; it would have respawned its channels pane every 5 minutes up to
+# MAX_CONSECUTIVE. The same zero also defeats the respawn-grace gate (the brake
+# meant to stop exactly that storm), so the two failures compound.
+#
+# The assertions run against the SHIPPED script through its `--file-mtime`
+# debug entry point, so this measures the code the timer runs, not a copy.
+#
+# Red-capability is not assumed, it is measured on every run: the last case
+# proves that a single-form implementation is genuinely wrong on THIS host, so
+# the passing cases above it cannot be passing for free. On macOS the broken
+# form is the GNU one (the actual regression); on Linux it is the BSD one.
+# Run: bash scripts/__tests__/channel-watchdog-mtime.test.sh
+
+set -u
+
+PASS=0; FAIL=0
+pass() { PASS=$((PASS + 1)); echo "  PASS: $1"; }
+fail() { FAIL=$((FAIL + 1)); echo "  FAIL: $1 -- expected: $2, got: $3"; }
+
+INSTALL_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+WATCHDOG="${WATCHDOG_BIN:-$INSTALL_DIR/scripts/channel-watchdog.sh}"
+
+echo "channel-watchdog.sh portable mtime (PORTSTAT912)"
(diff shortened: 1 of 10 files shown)


sigee82:
Merge order note: this branch is built on top of #1301, so its diff against `develop` currently also contains #1301's commits. Merge #1300, then #1301, then this one; after #1301 lands this diff shrinks to just the shared helper, the four call-site conversions and the new gate.

Opened from a fork because the machine that produced these branches only has read access to this repo.

sigee82:

Szotasz:
Verify result (Samu). **Content: GO, stacked after #1301** (as your description states — measured: this branch contains #1301's two commits). The shared `scripts/lib/portable-stat.sh` is well built: digit-validation against GNU's different `-f` semantics, the `PORTABLE_STAT_DISABLE` seam so the locally-unreachable branch stays exercised, and the allowlist gate that keeps a stray `stat -c` from reappearing. All 8 callers derive `INSTALL_DIR` from their own `$0`, so the sourcing path is runtime-safe. 10/10 tests pass **on macOS**.

**One blocker, in the test file: `BOX="$(mktemp -d -t portstat912b)"` is BSD-only** — on GNU/Linux `mktemp` fails (`too few X's in template`), which is why the CI `test` check is red. Portable form: `mktemp -d "${TMPDIR:-/tmp}/portstat912b.XXXXXX"`. Same class in #1300 and #1301 (details in my #1300 comment). A worthwhile bonus once fixed: the same "one spelling per platform" lesson your helper teaches applies to `mktemp -t` itself.

PR #1303, opened 2026-09-12
fix(claude-plans): make watchdog/keep-alive respawn rotation-aware

akobza:
## Summary
- `resolveMainConfigDecision()` (used by the JS in-process respawn paths in `channel-monitor.ts`: `resumeMarveenSession`, `respawnMainSessionFresh`) only ever considered the generic, credential-less flotta-isolated dir (`ensureMainAgentIsolatedConfigDir`). It never consulted `resolveMainAgentConfigDir()` (explicit) or `resolveMainAgentRotatedConfigDir()` (PR2c rotation), unlike the shell respawners which already get the full `explicit > rotated > isolated` precedence via `scripts/main-agent-isolated-config.mjs`.
- Net effect, observed live today on a remote install: `POST /api/claude-plans/rotate` correctly switches the main agent onto a registered plan's own configDir, but the very next watchdog/keep-alive respawn (triggered by an unrelated channel-plugin hiccup) silently reverted it back to the shared flotta identity — undoing the rotation with no error anywhere.
- Fixed by giving `resolveMainConfigDecision()` the same three-tier precedence, and adding an `ownCredentials` flag to `MainConfigDecision` so `buildMainSessionRespawnCmd()` does not inject the fleet OAuth token on top of a dir that already carries its own `.credentials.json` (explicit or rotated) — mirroring exactly what `channel-watchdog.sh` / `stuck-modal-guard.sh` already do for the shell paths.
- Bonus: the `channels-failures.log` trace line for a healthy launch now distinguishes `own-credential` from `isolated`, so a launch's actual identity source is visible in the log instead of always reading "isolated".

## Test plan
- [x] New test file `main-config-decision-rotation-precedence.test.ts`: rotated wins over plain isolated, explicit wins over rotated, plain isolated is still the fallback and is the only case that calls the (provisioning) `ensureMainAgentIsolatedConfigDir`.
- [x] New case in `channel-deafness-recovery.test.ts`: `buildMainSessionRespawnCmd` exports only `CLAUDE_CONFIG_DIR` (no fleet token) when `ownCredentials` is true.
- [x] Full existing suite green: `npx vitest run` — 431 files / 5487 passed, 1 skipped.
- [x] `npx tsc --noEmit -p .` clean.

### DIFF
--- a/src/__tests__/channel-deafness-recovery.test.ts
+++ b/src/__tests__/channel-deafness-recovery.test.ts
@@ -78,6 +78,21 @@ describe('buildMainSessionRespawnCmd', () => {
     expect(cmd).not.toContain('CLAUDE_CODE_OAUTH_TOKEN')
     expect(cmd).not.toContain('CLAUDE_CONFIG_DIR')
   })
+
+  // CLAUDEPLANWATCHDOG912: an own-credential dir (explicit MAIN_AGENT_CONFIG_DIR,
+  // or a rotated claude-plans entry) carries its OWN .credentials.json. Injecting
+  // the fleet token on top of it authenticates as the flotta instead of that
+  // login -- exactly the bug that silently undid a live plan rotation on the
+  // next watchdog respawn (observed 2026-09-12, solarforce-ai host).
+  it('exports ONLY the config dir, no fleet token, when the dir carries its own credentials (rotated/explicit)', () => {
+    const cmd = buildMainSessionRespawnCmd({
+      ...base,
+      continueSession: false,
+      config: mainConfigDecisionForTest({ isolatedConfigDir: '/home/solarforce/.claude-second', ownCredentials: true, fleetToken: true }),
+    })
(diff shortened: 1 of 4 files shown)


Szotasz:
Thanks, and sorry for the public silence on this one: it moved on our side and we failed to say so here.

Both this PR and #1304 were triaged and carded on 2026-09-13, and they are queued for an internal verify. We are honouring the stacking you asked for: #1303 is reviewed and merged first, #1304 follows once this lands on develop and its diff shrinks to its own change.

Two things we measured rather than took from the description, so you know where we stand:

- `MERGEABLE` / `CLEAN`, both required checks green, head `48081539`.
- The four files are the channel recovery stack, which is why this gets a second pair of eyes rather than a straight merge. That is our rule for that directory, not a comment on the change.

Nothing is needed from you right now. If the branch moves, please say so here, because a new head invalidates the verify and we would rather re-run it than merge something that was not the thing we looked at.

PR #1304, opened 2026-09-12
feat(claude-plans): token-mode plans (raw CLAUDE_CODE_OAUTH_TOKEN, no per-host setup-token)

akobza:
## Summary
Stacked on #1303 -- please review/merge that one first. This branch is built on top of #1303's commit, so until that merges, this PR's diff includes #1303's changes too (GitHub fork-PR limitation); it will shrink to just this feature's diff automatically once #1303 lands on develop.

- Adds a second `ClaudePlan` shape as an alternative to `configDir`: a plan can carry a `tokenSecretId` (a vault-stored raw `claude setup-token` output) instead of an absolute `CLAUDE_CONFIG_DIR`.
- Every token-mode plan shares the SAME generic isolated `CLAUDE_CONFIG_DIR`; rotation only changes which token is exported as `CLAUDE_CODE_OAUTH_TOKEN` at launch -- the same mechanism the fleet's own setup-token already uses.
- **Why:** confirmed live today on a remote install (solarforce-ai) -- two `configDir`-mode plans got registered through the dashboard with no credentials ever actually placed in their directory (the registration alone gives no signal anything is missing), and the next restart wedged on an interactive OAuth-code prompt a headless session can never complete. Token-mode needs no per-host interactive step at all: run `claude setup-token` anywhere with a browser, paste the resulting string into the plan.
- `POST`/`PUT /api/claude-plans` now accept a raw `token` field as a convenience: it's written to the vault and stripped before the plan ever reaches `validatePlan`/disk/response. Orphaned vault secrets are cleaned up on validation failure, duplicate id, plan deletion, or switching a plan out of token-mode.
- Also confirmed live today: on a systemd-managed Linux install, `channels.sh` itself (not the JS in-process respawn path #1303 fixes) is the actual dominant restart mechanism (`Restart=always` bounces `channels.sh` directly). Shipping token-mode without wiring the shell respawn paths too would have reintroduced #1303's exact bug class for the new mode on precisely that path -- so `scripts/main-agent-isolated-config.mjs`, `channels.sh`, `channel-watchdog.sh` and `stuck-modal-guard.sh` all got a fourth contract mode (`token`) in this same PR.

## Test plan
- [x] New/updated unit tests: `claude-plans.test.ts` (validatePlan token-mode cases), `main-agent-rotated-config-dir.test.ts` (resolveMainAgentRotatedTokenSecretId), `main-agent-isolated-config.test.ts` (ensureMainAgentIsolatedConfigDirForRotatedToken), `main-config-decision-rotation-precedence.test.ts` (token vs configDir vs explicit precedence), `channel-deafness-recovery.test.ts` (buildMainSessionRespawnCmd's token branch), `claude-plans-routes.test.ts` (raw-token-to-vault promotion lifecycle, with a fake in-memory vault -- never touches the real encrypted store or Keychain), `main-agent-config-dir.test.ts` + `main-config-dir-parity.test.ts` (shell-script contract parity for the new mode).
- [x] Full existing suite green: 431 files / 5507 passed, 1 skipped.
- [x] `npx tsc --noEmit -p .` clean.
- [ ] Not yet exercised live end-to-end on a real host (blocked on #1303 merging + a real `claude setup-token` output to test with) -- the requester has 3 real tokens ready and will exercise this via the raw API tonight once available.

### DIFF
--- a/scripts/channel-watchdog.sh
+++ b/scripts/channel-watchdog.sh
@@ -192,21 +192,32 @@ if [ -n "$NODE_BIN" ] && [ -f "$INSTALL_DIR/dist/web/agent-process.js" ]; then
   # onto the shared ~/.claude on 2026-09-12. This caller must stay in step with
   # channels.sh, or a watchdog respawn reintroduces exactly that outage.
   _cfg_raw="$("$NODE_BIN" "$INSTALL_DIR/scripts/main-agent-isolated-config.mjs" "$CHANNEL_PROVIDER" 3>&1 2>>"$STORE/channels-failures.log" 1>&2 || true)"
-  _cfg_line="$(printf '%s\n' "$_cfg_raw" | grep -m1 -E '^(explicit|rotated|isolated)	/' || true)"
+  _cfg_line="$(printf '%s\n' "$_cfg_raw" | grep -m1 -E '^(explicit|rotated|isolated|token)	/' || true)"
   if [ -n "$_cfg_raw" ] && [ -z "$_cfg_line" ]; then
     log "WARN main-agent-isolated-config.mjs printed output with NO contract line -- respawn without isolation"
   fi
   _cfg_mode="${_cfg_line%%	*}"
-  _cfg_dir="${_cfg_line#*	}"
+  _cfg_rest="${_cfg_line#*	}"
+  if [ "$_cfg_mode" = "token" ]; then
+    _cfg_dir="${_cfg_rest%%	*}"
+    _cfg_token_secret="${_cfg_rest#*	}"
+  else
+    _cfg_dir="$_cfg_rest"
+    _cfg_token_secret=""
+  fi
   if [ -n "$_cfg_line" ] && [ -d "$_cfg_dir" ]; then
     if [ "$_cfg_mode" = "explicit" ] || [ "$_cfg_mode" = "rotated" ]; then
       CFG_ENV="export CLAUDE_CONFIG_DIR='$_cfg_dir' && "
+    elif [ "$_cfg_mode" = "token" ]; then
+      # Token-mode rotated plan -- same credential-less dir as `isolated`, but
+      # export THAT plan's vault-stored token. See channels.sh's identical branch.
+      CFG_ENV="export CLAUDE_CONFIG_DIR='$_cfg_dir' && export CLAUDE_CODE_OAUTH_TOKEN=\"\$(printf 'T=%s' '$_cfg_token_secret' | \"$NODE_BIN\" '$INSTALL_DIR/scripts/vault-resolve.mjs' | cut -d= -f2-)\" && "
     else
       CFG_ENV="export CLAUDE_CONFIG_DIR='$_cfg_dir' && export CLAUDE_CODE_OAUTH_TOKEN=\"\$(cat '$INSTALL_DIR/store/.claude-oauth-token')\" && "
     fi
     log "main-agent $_cfg_mode CLAUDE_CONFIG_DIR=$_cfg_dir"
   fi
-  unset _cfg_raw _cfg_line _cfg_mode _cfg_dir
+  unset _cfg_raw _cfg_line _cfg_mode _cfg_rest _cfg_dir _cfg_token_secret
 fi
 
 # Full PATH with .bun/bin -- without it the respawned bun telegram bridge does
--- a/scripts/channels.sh
+++ b/scripts/channels.sh
@@ -642,18 +642,34 @@ if [ -n "$_node_bin" ] && [ -f "$INSTALL_DIR/dist/web/agent-process.js" ]; then
   # and a non-empty output with no contract line in it is reported LOUDLY, because
   # that is the shape that silently disables isolation.
   _cfg_raw="$("$_node_bin" "$INSTALL_DIR/scripts/main-agent-isolated-config.mjs" "$CHANNEL_PROVIDER" 3>&1 2>>"$INSTALL_DIR/store/channels-failures.log" 1>&2 || true)"
-  _cfg_line="$(printf '%s\n' "$_cfg_raw" | grep -m1 -E '^(explicit|rotated|isolated)	/' || true)"
+  _cfg_line="$(printf '%s\n' "$_cfg_raw" | grep -m1 -E '^(explicit|rotated|isolated|token)	/' || true)"
   if [ -n "$_cfg_raw" ] && [ -z "$_cfg_line" ]; then
     echo "$(date '+%Y-%m-%d %H:%M:%S') channels.sh: WARN main-agent-isolated-config.mjs printed output with NO contract line -- isolation skipped. Raw first line: $(printf '%s\n' "$_cfg_raw" | head -1)" >> "$INSTALL_DIR/store/channels-failures.log"
   fi
   _cfg_mode="${_cfg_line%%	*}"
-  _cfg_dir="${_cfg_line#*	}"
+  _cfg_rest="${_cfg_line#*	}"
+  # `token` mode carries a THIRD field (the vault secret id) after the dir;
+  # every other mode's rest IS the dir.
+  if [ "$_cfg_mode" = "token" ]; then
+    _cfg_dir="${_cfg_rest%%	*}"
+    _cfg_token_secret="${_cfg_rest#*	}"
+  else
+    _cfg_dir="$_cfg_rest"
+    _cfg_token_secret=""
+  fi
   if [ -n "$_cfg_line" ] && [ -d "$_cfg_dir" ]; then
     if [ "$_cfg_mode" = "explicit" ] || [ "$_cfg_mode" = "rotated" ]; then
       # Both carry their OWN .credentials.json (an operator-logged-in dir for
       # `explicit`, a registered plan's dir for `rotated` -- design 6.5/4) --
       # neither wants the fleet token injected below.
       CFG_ENV="export CLAUDE_CONFIG_DIR='$_cfg_dir' && "
+    elif [ "$_cfg_mode" = "token" ]; then
+      # Token-mode rotated plan: same credential-less dir as `isolated`, but
+      # export THAT plan's vault-stored token instead of the flotta's.
+      # vault-resolve.mjs is evaluated in the launched shell (same "never
+      # lands in argv/ps" property as the $(cat) below); the secret id itself
+      # is not secret, only its resolved value is.
+      CFG_ENV="export CLAUDE_CONFIG_DIR='$_cfg_dir' && export CLAUDE_CODE_OAUTH_TOKEN=\"\$(printf 'T=%s' '$_cfg_token_secret' | \"$_node_bin\" '$INSTALL_DIR/scripts/vault-resolve.mjs' | cut -d= -f2-)\" && "
     else
       # Seed the token from the SAME 0600 file the isolated dir is gated on, so
       # the config dir and the active token always match (the isolated dir carries
--- a/scripts/main-agent-isolated-config.mjs
+++ b/scripts/main-agent-isolated-config.mjs
@@ -44,24 +44,36 @@ let CONTRACT_FD = 3
 try { fstatSync(CONTRACT_FD) } catch { CONTRACT_FD = 1 }
 const emitContract = (line) => writeSync(CONTRACT_FD, line)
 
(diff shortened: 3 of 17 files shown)


Szotasz:
Thanks, and the same apology as on #1303 for the silence here: this was triaged and carded on 2026-09-13, we just did not say so publicly.

The stacking is understood and respected: #1303 goes first, and we will re-measure this diff after it lands, when it shrinks to just the token-mode change.

Where this one stands, measured rather than assumed:

- `MERGEABLE` / `CLEAN`, both required checks green, head `82599377`, 17 files.
- It touches comms infrastructure (`channels.sh`, `channel-watchdog.sh`, `agent-process.ts`, `channel-monitor.ts`) and it changes how the fleet authenticates. The first part means an internal verify; the second part means it is not a review decision at all but an owner decision, so it will go in front of him with the verify attached.

So this will take longer than #1303, and the reason is the authentication surface, not the quality of the work. We will keep the state visible here rather than leaving you guessing.

Szotasz:
Thanks for this, and thanks for the honest note in the description that the token-mode path was not exercised end to end in production. We verified the rest and it holds; the rebase below is ours, not yours.

**Rebase done on our side.** Because #1303 was squash-merged, this branch carried its commit in the pre-squash form and went CONFLICTING. We rebased with `git rebase --onto origin/develop 48081539`, replaying only the delta commit, and force-pushed with lease to the fork branch. Authorship is preserved (you remain the author). New head `fa8ad02f`, a fast-forward on develop. Negative control on the duplication: the `resolveMainConfigDecision` anchor appears 4 times in the raw diff but all 4 are context lines; on added/removed lines it is 2, which are the delta's own changes, and of the 130 added lines from #1303 the 13 that reappear here are all rewritten by the delta itself (test assertions, import list). Residue from #1303: zero. You are now looking at a clean delta.

**Verified** (merge-preview tree, separate worktree, `rm -rf dist` + own `npm ci`, project's own test script): 465 test files / 5865 passed / 1 skipped / 0 failed, tsc exit 0, SHA `fa8ad02f`.

**What is correct and we want to keep:** exactly one `configDir`/`tokenSecretId` per plan; the raw token goes to the vault and never to disk or a GET response; the launch side resolves from the vault inside the respawned shell so argv and `ps` stay clean; and the four shell paths share one `token` contract with a parity test on all three callers.

**Three changes required before we merge.** The first two are measured, not argued, and both are cheap.

**(a) Vault write before validation destroys a live secret.** `withTokenPromotedToVault()` calls `setSecret()` on the deterministic id `claude-plan-token-<planId>` BEFORE `validatePlan()`. On validation failure the cleanup `deleteSecret()` removes that same id, so an INVALID PUT carrying a raw token overwrites and then deletes the plan's existing token while the plan on disk still points at it. Measured: POST a token plan, then PUT `{token: new, planType: invalid}` gives 400 and the old secret is gone; the next respawn starts with an empty token.
Required: validate first, using a placeholder `tokenSecretId`, and write the vault only after validation succeeds. Test: an invalid PUT carrying a raw token leaves the existing secret intact.

**(b) `tokenSecretId` may name ANY vault entry, and plan delete or switch deletes it.** `validatePlan` only checks the charset, and the routes call `deleteSecret(previousTokenSecretId)` / `deleteSecret(removed.tokenSecretId)` on whatever id the plan holds. Measured: PUT `tokenSecretId='MARVEEN-CONNECTORS-PAT'` returns 200, then DELETE plan removes that foreign secret. This is a destructive primitive reachable by a typo, and on our install that particular id is the credential we use to query live production databases.
Required: accept ONLY the derived form `claude-plan-token-<planId>`, rejecting anything else in `validatePlan` or in the routes, and let `deleteSecret` run only on that prefix. Tests: a foreign id is rejected on POST and PUT, and a plan DELETE never removes a non-prefixed secret.

**(c) A missing vault secret boots an unauthenticated session silently.** `$(printf 'T=%s' '<id>' | node vault-resolve.mjs | cut -d= -f2-)` swallows `vault-resolve`'s exit 3 (measured live: exit 3, empty stdout, label-only stderr), so `CLAUDE_CODE_OAUTH_TOKEN` is exported EMPTY and the session lands on exactly the interactive login this PR sets out to prevent, in all four respawn paths the PR adds. We are treating this as blocking rather than a nice-to-have, because the vault-resolve path is introduced by this PR: it is a new regression of the PR's own symptom, in its quietest form.
Required: the launch path must detect the empty value, log a `channels-failures.log` line naming the plan and secret id but never the value, and either fall back to the fleet token or stop loudly. Test: a missing secret must NOT start the session with an empty token.

For the next round, what we will check: the three negative tests (invalid PUT leaves the secret; a foreign id is rejected and never deleted; a missing secret does not start with an empty token), plus the merge-preview tree with three numbers.

akobza:
Pushed all three required fixes on top of your rebase (head fa8ad02, new head b3a8c05, fast-forward -- your rebase is untouched).

**(a) Vault write before validation.** `withTokenPromotedToVault` is now `prepareTokenCandidate` -- pure, writes nothing. It only builds the candidate `validatePlan()` sees and what *would* be written. The route commits to the vault (`setSecret`) only after `validatePlan()` -- and on POST, the duplicate-id check too -- has already succeeded. An invalid PUT (bad token + some unrelated bad field) now touches the vault not at all; the existing secret survives byte for byte.

**(b) tokenSecretId could name any vault entry.** `validatePlan` now requires it to be exactly the plan's own derived id (`claude-plan-token-<planId>`), not just charset-valid. A foreign id (e.g. `MARVEEN-CONNECTORS-PAT`) is rejected on both POST and PUT before it ever reaches disk. The route's `deleteSecret` call sites (PUT-switch, DELETE) now go through a `deleteOwnTokenSecret` guard as defense in depth -- it re-derives the expected id from the plan id and refuses to delete anything else, independent of validatePlan ever being bypassed.

**(c) Missing vault secret silently exported empty.** New `scripts/resolve-plan-token-env.mjs` replaces the inline `printf ... | vault-resolve.mjs | cut -d= -f2-` pipeline in all four respawn paths (channels.sh, channel-watchdog.sh, stuck-modal-guard.sh, channel-monitor.ts's `buildMainSessionRespawnCmd`). It resolves the plan's own vault secret; if that's missing it falls back to the fleet token and logs a line naming the plan's secret id (never any value) to `channels-failures.log`; if *neither* is available it exits 1 with nothing on stdout. Each caller assigns the result to a bare `_plan_token=$(...)` (no command word before it, so its own exit status propagates), so a resolver failure stops the `&&` chain before `claude` ever launches -- no path left that boots unauthenticated.

**Tests, all three required negative cases plus the surfaces the new script touches:**
- `claude-plans-routes.test.ts`: "an invalid PUT carrying a raw token leaves the existing secret intact", "rejects a foreign tokenSecretId on POST/PUT", plus a direct unit test on `deleteOwnTokenSecret` for the DELETE-never-removes-a-non-prefixed-secret guarantee (unreachable through the route now that (b) is fixed, so tested at the guard itself).
- `claude-plans.test.ts`: `validatePlan` rejects a well-formed-but-foreign tokenSecretId, and one plan's id reused on another.
- New `resolve-plan-token-env.test.ts` (same real-subprocess-against-stubbed-dist approach as `vault-resolve-loud-failures.test.ts`): success path, fallback path, the required "missing secret does not start with an empty token" case (exit 1, empty stdout, no fleet token), and that the failures log never carries a secret value.
- Updated the existing contract tests that asserted the old shape (`channel-deafness-recovery.test.ts`, `main-agent-config-dir.test.ts`, `main-config-dir-parity.test.ts`) to the new one.

Full suite: 466 files / 5883 passed / 1 skipped / 0 failed. `tsc --noEmit -p .` clean.

Szotasz:
FLEET REVIEW -- GO (technical, delta verify of `b3a8c055` against the rebased `fa8ad02f`)

Measured today, two days after your push, on a home-rooted clone of your head with a throwaway HOME (the suite refuses tmp-rooted checkouts by design). This is the round we promised on 09-18: the three negative tests, plus the merge-preview tree with three numbers.

**The three required changes are in, and each one is pinned by a test that kills its mutant.**

| | where | test | mutant (file-based patch, restored after) |
|---|---|---|---|
| (a) validate before vault write | `routes/claude-plans.ts`: `prepareTokenCandidate` is pure (no `setSecret`); the only writes are after `validatePlan` and, on POST, after the 409 duplicate check; the failure-path `deleteSecret` calls are gone | `claude-plans-routes.test.ts:252`: 400, old value byte-equal, `setSecret`/`deleteSecret` never called | `setSecret` moved back before `validatePlan` on PUT: 1 red, `an invalid PUT carrying a raw token leaves the existing secret intact` |
| (b) derived id only | `claude-plans.ts:94` `tokenSecretIdFor`, `:139` rejects anything but `claude-plan-token-<planId>`; `deleteOwnTokenSecret` (`routes/claude-plans.ts:100`) guards PUT-switch and DELETE; `resolveClaudePlans` re-validates on read, so a hand-edited foreign id never surfaces | routes `:280` (POST), `:290` (PUT), `:307` (guard); `claude-plans.test.ts:121`, `:126` | charset-only again: 4 red (POST foreign 201, PUT foreign 200, both unit tests) |
| (c) no empty token | new `scripts/resolve-plan-token-env.mjs`: secret found, exit 0; missing, fleet-token fallback with a log line naming plan and id; neither, `FATAL` log and exit 1 with nothing on stdout. All four respawn strings use `_plan_token="$(node ...)" && export CLAUDE_CODE_OAUTH_TOKEN="$_plan_token"` | `resolve-plan-token-env.test.ts:82`, `:90`, `:96`, `:103` (log never carries a value) | resolver prints empty and exits 0: 5 red |

- Suite on the five touched test files: 108/108. `tsc --noEmit`: clean.
- Shell side of (c): the guard sits on all four token-mode respawn surfaces (`channels.sh:677`, `channel-watchdog.sh:216`, `stuck-modal-guard.sh:318`, `channel-monitor.ts:827`); the old `| cut` pipeline count across the five files is 0; `bash -n` clean on the three scripts. Under `/bin/sh`, `/bin/bash` and `/bin/zsh` the `&&` chain stops on exit 1 and `CLAUDE_CODE_OAUTH_TOKEN` stays unset, where the old shape launched with an empty token. The log lines interpolate only the secret id and the fleet-token path, never a value.
- Merge preview: `git merge-tree --write-tree` against today's `develop` (`6521276a`): clean. The three numbers: 19 files, +964, -76. Six develop commits since `fa8ad02f` touch the same surfaces; in the merged tree the per-surface counts of the guarded export are identical to your head, so develop added no unguarded path.

**One observation, not a blocker, for the fleet lead:** the resolver imports `getSecret` from `dist/web/vault.js` directly, the same way develop's own `vault-resolve.mjs` and `getSecretsForEnv` do. Today's #1424 puts the audit row on the vault HTTP route, not on the library reader, so this path is unaudited like every existing in-process reader. Same class, not a new one; it belongs to the vault-audit follow-up, not to this PR.

**One small gap you may want to close, optional:** the parity test pins the resolver's presence in `channel-watchdog.sh` and `stuck-modal-guard.sh`, but the bare-assignment shape (`_plan_token=` then `export`) is only test-pinned for `channels.sh` and `channel-monitor.ts`. A regression that reintroduces `export X="$(...)"` in the two shell files would pass today.

Not measured by us: the full suite (your 466/5883 figure), and a live token-mode boot on a real install.

This changes how the fleet authenticates, so the merge is the owner's decision, not a review outcome; the verify goes in front of him with this attached. No date from me.

akobza:
Attila approved the merge decision (2026-09-22, Telegram) — this is a go on our side for `b3a8c055`.

Nothing further requested beyond the review above. The one optional gap you flagged (parity test only pinning the bare-assignment shape in `channels.sh`/`channel-monitor.ts`, not `channel-watchdog.sh`/`stuck-modal-guard.sh`) can be a follow-up, not a blocker.

Since neither of us can push to this repo, over to you to merge whenever convenient.

akobza:
Friendly nudge on this one: it is still GO from your 09-20 review, GitHub reports it MERGEABLE/CLEAN against today's develop, and Attila's merge approval from 09-22 stands.

It has become the base for the next piece Attila asked for today: under Settings -> Claude plans, paste several keys, see each key's live 5h/7d usage, and let the rotation keep the fleet (main agent and the shared fleet token) off an exhausted key. We will build that as follow-up PRs on top of this branch, so a merge here unblocks them. #1511 also waits on this per your note there.

PR #1318, opened 2026-09-14
fix(email-gates): the claude.ai Gmail connector's sends reach every gate (GMAILCONNECTOR914)

jkarcsi:
## The gap, measured

The claude.ai Gmail connector names its tools `mcp__claude_ai_Gmail__{send_message,reply,forward,create_draft,...}`: no `send_email`, no `manage_email`, and ONE underscore before `Gmail`. Piping the same payload (em dash in the body) into the shipped hooks, on v1.37.0:

| tool | outgoing-copy-gate |
|---|---|
| `mcp__gmail__gmail_send_email` | exit 2 (blocks) |
| `mcp__google-workspace__manage_email` | exit 2 (blocks) |
| `mcp__claude_ai_Gmail__send_message` | **exit 0, zero output** |

Two reasons, stacked: the `.claude/settings.json` matchers (`.*send_email.*`, `.*manage_email.*`) never fire on the connector so the hooks do not even run; and inside `outgoing-copy-gate.py` the `(^|__)gmail__(...)` alternative needs the server segment to be exactly `gmail`, which `claude_ai_Gmail` is not. The approval gate (`send_email|manage_email`) and the sub-agent send gate (`/send_email/`) have the same blind spot, so a connector send was never audited, never approval-checked, and a sub-agent could send through it with no governance gate at all. Same class as MANAGEOP904 / COPYGATEMATCHER904: a name-list classifier that did not know one server's spelling.

## The fix

- `EMAIL_GATE_MATCHER` gains `.*[Gg]mail__.*`: the whole server, not another name list. The hooks classify by operation (a search or a read exits 0 in every gate), and a name list is exactly what drifted here. The tracked `.claude/settings.json` follows it for both main-agent hooks (the drift test enforces that), and the existing stale-matcher migration repairs already-scaffolded installs.
- `outgoing-copy-gate.py`: `EMAIL_TOOL_RE` matches `gmail__<send tool>$` with any prefix.
- `email-approval-gate.py` + `email_extract.py`: connector sends are in scope; `forwardText` is part of the audited body; a `reply`/`forward` without a `to` field anchors its recipient on `messageId` (bytes-exact, so an approval for one message cannot answer another).
- `email-send-gate.mjs`: the three send-shaped connector tools are denied for sub-agents with kind `connector-send`, deliberately NOT `send_email`, so the thread-reply narrowing (which reads `threadId`/`to`) never applies to a connector call. Reads and drafts stay open.

## Tests

3 new vitest cases (send gate; matcher drift, including a full-match probe of the connector names against the canonical matcher), 16 new python checks in the copy-gate and approval-gate suites (em dash blocks / clean passes for all three tools, `forwardText` audited, `draftId`-only send fail-closed, reads untouched, level 1 deny, level 2 approve-once on the messageId anchor). Full suite: 5554 green; the 2 `script-tests-runner` failures (`intel-db.test.sh` needs a sqlite3 binary, `migrate-agent-id-linux` is a macOS keychain case) fail identically on `develop` without this change.

https://claude.ai/code/session_016GurHMXNFeVvBoeoy8REg4


Szotasz:
Thanks for this, and sorry for the four day silence. That delay is on us, not on the submission.

**I reproduced your central claim on our own host before reading the diff**, because a PR description is not a source. On `develop` (head `9f697ded`), piping the same shape into the shipped hook:

```
mcp__claude_ai_Gmail__send_message + em dash in body  -> exit 0, zero output
mcp__gmail__gmail_send_email       + em dash in body  -> exit 2, gate message
```

The second line is the positive control: it rules out a broken harness on my side. So this is an active gap, not a latent one, and the outgoing-copy gate is a rule the owner has personally asked for twice. Your diagnosis holds.

**What I checked beyond the premise**

1. The matcher widening is the part that needed the most care, because the question is not the new condition but what the guarded block does for the newly admitted cases. Broadening to `.*[Gg]mail__.*` means the hooks now also run on connector reads. Your suites answer exactly that (`search_threads`, `get_thread`, `label_message` pass untouched at exit 0, and `create_draft` is not treated as a send), which is the right place for it.
2. `.claude/settings.json` adds the matcher rather than replacing the existing ones, so `.*send_email.*` still covers `mcp__server-gmail-autoauth-mcp__send_email`, which is the server we actually run. I verified that name does not match the new pattern, so the old entry is load bearing and must stay.
3. The `messageId` anchor for `reply`/`forward` without a `to` field is the detail I would have expected to be missing, and it is there, with a test asserting one approval cannot answer another message.

**One observation, not a blocker.** The new pattern is `[Gg]mail__`, which is still a small case list, and the failure it fixes is precisely a name list that did not know one spelling. A server spelled `GMAIL__` or `gMail__` would slip through the same way. If Claude Code matchers accept an inline case insensitive flag, that would match the argument in your own description better than the two letter class. I have not measured whether the matcher syntax supports it, so treat this as a question rather than a change request.

**What happens next, so you are not waiting on silence again.** This touches governance gates, so it goes to a second pair of eyes before merge rather than straight in on my review. That is our standing rule for gate and permission changes, and I applied it to two of my own PRs this morning after skipping it, so it is not something I am inventing for an external contributor. I will report back here with the verify result.

Szotasz:
A correction to my own comment above, and two questions.

**I overstated one thing.** I wrote that this is "an active gap, not a latent one". What I actually measured is narrower: the shipped hook exits 0 on the connector tool name and exits 2 on `mcp__gmail__gmail_send_email` with the same payload. That is a real difference in the hook, and your diagnosis of it stands. It does not establish that the connector tool name can reach a hook in a Claude Code session, which is the other half of the claim.

I then measured that half on our side, three ways, and all three are negative:

```
claude mcp list                         -> google-drive, server-gmail-autoauth-mcp, server-google-calendar-mcp
tool search for claude_ai_Gmail         -> no match
~/.claude.json claudeAiMcpEverConnected -> 16 entries, none containing Gmail
```

So on this install the connector has never been connected at all, and I cannot invoke the tool to see what a real call looks like. That does not argue against the change. A fail closed gate on a name we cannot invoke costs nothing here and protects installs where it can occur. It only means the severity sentence should say "hardening that closes a gap the hook demonstrably has" rather than "a hole that is live today", unless you can show the other half.

**Question 1, and only you can answer it.** In your environment, can a `mcp__claude_ai_Gmail__*` tool call actually reach a Claude Code PreToolUse hook? If yes, say how it is wired, because that is what turns this from hardening into a live fix, and it belongs in the commit body rather than in a reviewer's head.

**Question 2, about the fail closed branch.** The suite asserts that a `send_message` carrying only a `draftId` exits 2, because the body cannot be read. Is sending an existing draft a real path on that tool, or did you construct that payload as a boundary case? If it is real, denying it is a product decision rather than a review one, and I would rather decide it deliberately than discover it from a user whose draft send started failing.

**One thing you do not need to do.** The PR is now behind develop and conflicts in `scripts/__tests__/outgoing-copy-gate.test.py`, where both sides appended a block. That conflict exists because we merged another PR ahead of yours, so the rebase is ours to carry, not yours. We will do it on your commit so the authorship stays with you, and I will say here when it is done.

Szotasz:
Merged as `000b9c6d`. Thanks for this one.

**Heads up on one thing you did not do: we rebased your branch.** The PR head is now `c22a3d45`, which is your commit cherry-picked onto `9f697ded` with your authorship preserved (committer is ours). There was exactly one conflicting file, a test, and it was resolved as a union: the `COPYGATEENT914` block from develop first, your connector block after it. Nothing of yours was dropped. If you had local work on the old head, rebase or re-fetch before you continue.

**What we measured before merging**, beyond re-running your suite:

*The live-path regression probe.* Widening a hook matcher is the kind of change whose real risk is not the new case but the old one. The new `.*[Gg]mail__.*` alternative does not match the Gmail MCP we actually use day to day (`mcp__server-gmail-autoauth-mcp__*`), because there is a hyphen after "gmail", not two underscores. So `send_email` still gets caught by the pre-existing alternative and read/search/draft stay outside the gates exactly as before. Zero regression on the live path.

*Mutation control, 9 one-line mutants:* 8 die. One survives, the `forwardText` field collection: the test only asserts exit 2, and the mutant also exits 2, for a different reason (empty text failing closed instead of the em dash rule). The shipped code is correct here, the test is weak. Not a blocker, but worth tightening: assert the em dash reason specifically, and add a clean-forwardText-exits-0 case, otherwise that line is effectively untested.

*Also verified:* fail-closed on a draftId-only send (PR head exit 2, develop exit 0), thread-reply narrowing with a positive control, and the full suite under an isolated HOME (5822 green; the one red is a pre-existing develop-side instrument bug with gitignored `.pyc` files, tracked on our side, not yours).

**Two open questions, neither blocking:**

1. Does the connector's `send_message` accept a `draftId` at all? You have the environment for that, we do not.
2. The `[Gg]` two-letter class: we did not measure whether the matcher is case sensitive in the first place, so we cannot say whether that class is load-bearing or belt-and-braces.

**One scoping note we kept in the merge commit, so it does not get overstated later:** this is hardening, not a demonstrated live hole on our own install. What is proven is that the gates deny at the hook level. What is *not* proven is that the tool name reaches a hook in a live session, because the connector is not wired up anywhere in our fleet. That distinction is worth keeping in the record.

PR #1326, opened 2026-09-14
fix(web): stop writing the dashboard token into the error log on every boot

Itskatuska:
The bootstrap URL carries the bearer token, which is root-equivalent for this dashboard: it reads and writes memories, kanban, agents and schedules. The existing comment says it is printed to stderr rather than the pino stream so it "does not land in the log" -- but under launchd stderr IS a file (`store/dashboard.error.log`). The mitigation only moved the credential from one capture to another.

Measured 2026-09-01: the token was sitting in the live error log and in a gzipped archive of it (13 occurrences), re-written on every boot. Rotating the token put the fresh one back into the same file within seconds.

## Change
Gate the URL on `process.stderr.isTTY`: an operator starting the dashboard in a terminal still gets the clickable link; a redirected stream gets the path to the 0600 token file instead.

## Verification
After the change the error log contains no token (checked on a live launchd install).


Szotasz:
Confirmed on a live install, and the permission differential is the part worth writing down.

On the maintainer machine the dashboard token appears 241 times in the launchd-redirected
`store/dashboard.error.log` -- once per boot, exactly as you describe. The measurement that
settles the severity is not the count, though, it is the mode:

- `store/.dashboard-token` is `0600`
- `store/dashboard.error.log` was `0644`, carrying the same value 241 times

So the boot print takes a credential the code deliberately stores owner-only and republishes it
world-readable on the same box. That token reaches every `/api/*` route, including the vault
endpoints, so "root-equivalent" in your comment is accurate rather than rhetorical.

Two notes on the fix itself:

1. Gating on `process.stderr.isTTY` is the right axis. A piped interactive start (`npm start | tee`)
   also takes the non-TTY branch and only gets the path -- that is the safe direction, so no change
   requested.
2. Your comment records why the earlier "print to stderr instead of pino" step was insufficient.
   Please keep that paragraph. It is the reason the next person will not re-introduce this by
   moving the write to another stream.

Immediate mitigation applied here ahead of the merge: the log files are now `0600`. That is a
local stopgap, not a fix -- the code still writes the value, and every installation does the same
on its own machine.

Release timing is with the product owner today because it changes the scope of a release already
cut. Rotation of already-exposed tokens is tracked separately; it has to follow the merge,
otherwise a fresh token lands in the same file on the next boot.

Thank you for this one. It was found from the outside, on a path our own checks did not look at.

PR #1327, opened 2026-09-14
feat(profiles): add the marketer-permissive security profile

Itskatuska:
Same allow/deny guardrails as `marketer`, but runs in permissive mode so the agent does not stop for a permission prompt on every tool call. The deny list (sudo/rm/.ssh/.env/POST) still runs BEFORE the bypass, so the hard blocks stay in force.

Already in use by two agents on a live install; this commits the template file the install has been reading, so a fast-forward update no longer drops it.


Szotasz:
Thank you for this, and in particular for writing the risk into the PR description yourself. "Prompt-injection risk is higher than strict; egress gate plus deny list are the compensating controls" is an honest summary, and it is the sentence that made this reviewable rather than something we had to reverse-engineer.

We are not taking it upstream. The reason is narrow, and it is not about code quality.

The profile targets a web-reading role, and that is exactly where prompt injection is the live surface rather than a theoretical one. In permissive mode an injected instruction does not meet a permission prompt, it only meets the deny list. The deny list is tool-level: it blocks `sudo`, `rm`, `.ssh`, `.env`, `curl -X POST`. It cannot cover what an already-allowed tool reads or where an already-allowed request sends it. So the compensating control and the risk are not on the same axis, and for this particular role we would rather keep the prompt.

That is a judgement call about our own fleet, not a verdict on the pattern. On a role that does not read untrusted web content the tradeoff looks different.

**One thing you should know about your own install, because it is the practical consequence of this decision and it is not obvious.**

You wrote that two agents already run on this profile, and that the PR was committing the file the install has been reading. With the PR closed, that file stays untracked, and the loader resolves profiles only from `templates/profiles/`. When the file is not there, `loadProfileTemplate()` does not error and does not warn: it silently falls back to `default`. So on the next fast-forward update those two agents would quietly change security profile, and nothing in the logs would say so.

Two notes on that:

- Keep a copy of the file outside the checkout and restore it after updating, until there is a supported place for local profiles. There is no user-level or install-level profile directory today.
- The silent fallback is our gap, not yours. A missing security profile should be loud. We are tracking that separately, and your PR is what surfaced it.

If you want to take another run at this for a non-web-facing role, or at the local-profile directory question, both are worth a PR.

PR #1328, opened 2026-09-14
fix(scheduler): keep self-healing heartbeats out of the owner's Telegram

Itskatuska:
Two alert paths pushed heartbeat noise at the operator.

1. **Catch-up summary** reported every missed occurrence, heartbeats included. A 15-minute heartbeat that misses a tick is not operator-actionable (the next one is minutes away), and reporting it turned a routine heartbeat into ~25 alerts a day on a live install.
2. **Post-fire timeout watchdog** had no type filter. The new two-stage escalation on develop helps (main agent first), but stage 2 still pages the owner for heartbeats. Measured 2026-09-01 before this change: 40 owner alerts, 18 from a single `type: heartbeat` + `skipIfBusy: true` task, which is designed to drop its own tick when busy.

## Change
- `TaskInflightEntry.taskType` is captured at injection time (same reasoning as `timeoutMs`), next to upstream's `runId`.
- `sendTaskTimeoutAlert` returns before any network call for heartbeats.
- The catch-up summary filters heartbeats out of the report.

Neither path goes silent: both still WARN-log and record in `task_runs`. Only the push notification is suppressed.

## Tests
Adds a fix-revert guard in `schedule-task-timeout.test.ts` (heartbeat check must sit before the send). `schedule-task-timeout` + `schedule-catchup` pass on develop; `tsc --noEmit` clean.


Szotasz:
Fleet verify result (measured on head afd66d6): **GO on the code**, with one request on the test before merge.

The change itself checks out on our install: `schedule-task-timeout` + `schedule-catchup` 64/64 green, `tsc` clean, and the filter is not a no-op here (both of our live heartbeat tasks are `type: heartbeat` + `skipIfBusy: true`). The bookkeeping is preserved: the caller sets `ownerAlerted` after the call, so the suppressed path marks once with no per-sweep WARN spam, and the stage-1 main-agent notice still fires, which matches your "neither path goes silent" claim.

**The one request.** The fix-revert guard `sendTaskTimeoutAlert bails out for heartbeat tasks` pins less than its title promises: it asserts the guard *expression* appears before the send call, but not the early `return`. We ran the mutation: with the `return` removed (guard logs, then falls through, the alert still goes out), all 64 tests stay green. So a future refactor could reintroduce the noise while the guard test passes.

A one-line strengthening that keeps your index approach:

```diff
     expect(guardAt).toBeLessThan(sendAt)
+    // The guard must actually bail out, not merely mention heartbeats before the send:
+    expect(fn).toMatch(/taskType === 'heartbeat'\)\s*\{[\s\S]{0,400}?\breturn\b/)
```

A small behavioral test would be even better, but the pin above already turns our mutation red.

Timing note: merges into develop are frozen until tomorrow's weekly release promote, so this request costs you no time against the merge. Maintainer edits are enabled on the PR, so if you prefer, say the word and we will push this one-liner to your branch; if we have not heard back by Wednesday (Sep 16), we will do that so the merge is not held up. The fix and its diagnosis are yours either way.

PR #1351, opened 2026-09-16
fix(memory): drop the stale embedding when a memory edit changes the embedded text (EMBSTALE916)

sigee82:
Cherry-pick of `04e29755b678905f781129ac4ecf5c0771bcc6ef` (2026-08-30), which never reached `develop`. Measured, not assumed:

- `git merge-base --is-ancestor 04e2975 develop` -> **NO**; `git branch --contains` lists only `feat/k2-answer-watchdog`.
- `grep -c "embedding = NULL"` on develop: `src/db.ts` 0, `dist/db.js` 0. The 2026-09-15 09:36 rebuild of `dist/` therefore silently reverted the fix that had been verified in the running dist on 08-30.

## The defect

`updateMemory` wrote content/keywords but never touched the `embedding` column, so after every edit the vector still described the PREVIOUS text. Nothing surfaced it: there is no `embedding_generated_at` column, and neither search path errors -- FTS and the LIKE fallback read `content` live and stay correct, while `hybridSearch` keeps fusing the stale vector's ranking into its RRF.

It also hides itself: `backfillEmbeddings` processes exactly `WHERE embedding IS NULL`, so a row that is never nulled never enters the set, and the closing "0 NULL left" reads clean.

## The fix

Set `embedding = NULL` on an edit that changes the embedded text, handing the row back to `backfillEmbeddings` (idempotent, resumable). Deliberately not regenerating inline -- that would put a synchronous Ollama call in the path of a DB write. The trigger is content **AND** keywords, because both `saveAgentMemory` and `backfillEmbeddings` embed `content + ' ' + keywords`; the check compares against the stored values rather than parameter presence, since the PUT route resends the unchanged body on a category-only edit.

## Conflict resolution (the commit was NOT independent)

The cherry-pick does **not** apply cleanly. The original extended a `before` SELECT introduced by a *different* k2 commit (cache invalidation from the row rather than the parameter); develop has no such SELECT. Resolution here: add the `before` read narrowed to the two columns this fix needs (`content`, `keywords`), and leave out the cache-invalidation comment that belongs to the other commit rather than carry it over to describe a mechanism develop does not have. The k2 cache-invalidation change stays out of scope.

## Verification (re-measured on this resolution, not inherited)

- `tsc --noEmit`: clean.
- New file: 6/6 green.
- Negative controls: removing the guard line fails 2 of the 6; an always-null variant fails the other 3 -- both match the original commit's mutation claim, so the resolution did not hollow out the test.
- Full suite in a worktree off develop: **184 files / 2545 passed / 17 failed** vs. a develop baseline of **183 files / 2539 passed / 17 failed**. The same 17 pre-existing failures (`hook-path-guard`, `governance-gates`, `email-send-gate`), and the delta is exactly the 6 new tests.

## Not in this PR

An `embedding_generated_at` column (or anything leaving a trace of a vector's age) is a separate, later question. Its absence is what made the damage uncountable after the fact, but it does not belong in this change.

## After merge

`dist/` needs rebuilding and an independent re-measurement on the RUNNING instance -- `grep -c "embedding = NULL" dist/db.js` non-zero **and** a PUT on a throwaway memory row leaving `embedding IS NULL`. The two steps can fail separately, and the grep alone only proves the text is present.


sigee82:
Superseded by #1352. This branch was cut from a local `develop` ref that was 596 commits behind `origin/develop` and 7 ahead of it, so it carried 7 unrelated commits into the PR. #1352 is the same fix rebuilt from `origin/develop`, exactly one commit ahead of the base, re-measured from a $HOME-rooted worktree.

PR #1353, opened 2026-09-16
fix(build): dist provenance and upgrade-evidence paths stop depending on a human (BUILTCOMMIT916)

sigee82:
Two markers the upgrade procedure treats as **evidence** were maintained by hand or hardcoded. Both are now produced and guarded by code.

## 1. `dist/.built-commit`

The only machine-readable answer to "which commit produced the code that is running", and `scripts/frissites-verify.sh` reads it as exactly that.

**Measured 2026-09-16:** the file still named `0a38062` from 09-02, while a rebuild had happened on 09-15 without touching it. The direction of that lie is the harmful one — it claims dist is *older* than it is, so a finished upgrade reads as "not applied" and gets run again.

- **`npm run build` stamps it** from `git rev-parse HEAD` (`scripts/write-built-commit.mjs`). Line 1 stays the bare SHA, so `cat` and every existing reader keep working; line 2 adds context the SHA alone cannot give.
- **`scripts/built-commit-check.mjs` answers with three outcomes, not two:** `OK`, `ELTERES`, `HIANYZIK` — the missing case exiting non-zero. The old code printed `HIANYZIK` into the report and returned 0, so absence of evidence was rendered in the same calm tone as evidence of success; in a 200-line output a calm line reads as a passing line.
- **`frissites-verify.sh` propagates a non-zero exit.** It previously always returned 0.

**Dirty is scoped to build inputs** (`src`, `tsconfig.json`, `package.json`), not the whole tree. Measured on the deployed checkout: **161** modified/untracked paths, **0** of them build inputs. A whole-tree check would mark every build dirty, the verifier would be red on every run, and a gate that is always red teaches one reflex: override it. Untracked files still count *inside* the scope — a new `.ts` under `src/` does compile.

## 2. The upgrade backup directory

`BK` was hardcoded to `backups/frissites-v1360`, while the file names inside come from the **phase** only (`verify-elotte.txt`, `schedule-utana.txt`) and carry no version. A v1.38.0 run would have overwritten the 2026-09-02 evidence **in place, under identical names** — a loss with nothing left to notice it by.

Measured: that folder still holds the only pre/post bundles and the consistent DB copy, and the 09-08 v1.37.0 round never ran this script, so the overwrite had not happened yet.

`scripts/frissites-backup-dir.mjs` derives the directory from an explicitly passed target version, stamps a directory it creates with `.verzio`, and **stops non-zero** when the target belongs to another version or is a non-empty unstamped folder whose name is not this version. It never writes into an existing evidence folder.

**The version is an argument, not `package.json`.** Measured: `package.json` reads 1.37.0 on the deployed branch and 1.38.0 on develop, so it changes *between* the two phases — precisely when they must agree. Deriving it would send `elotte` and `utana` to different directories, and the comparison step would then report "NINCS elotte-pillanatkep" and carry on, so the tooling's failure would read as an operator mistake. There is deliberately no default.

## `frissites-verify.sh` is added, not modified

It was untracked — present in one machine's working tree and in no branch. The tool the upgrade leans on could have been lost by any checkout, **including the upgrade's own**. Same move as PAYCROSS915. Its signature changes: `elotte|utana <cel-verzio>`.

## Verification

- `tsc --noEmit` clean. New test file **21/21** green.
- Full suite **444 files / 5633 passed / 1 skipped / 0 failed**, against a develop baseline (68b2b21) of **443 / 5612 / 1 skipped / 0 failed**. Delta is exactly the 21 new tests; the skip is pre-existing.
- **Mutation controls, each caught:** build script without the writer; checker exiting 0 on a missing marker; dirty widened back to the whole tree (2 fail, including the false-alarm control); the version-stamp collision check removed; a default version reintroduced; the hardcoded v1360 path restored.
- **Measured on the live checkout, nothing left changed:** marker wrong SHA → `ELTERES` exit 1; removed → `HIANYZIK` exit 3; correct → `OK` exit 0. Backup dir pointed at the real v1360 folder with target 1.38.0 → stops, exit 66, and the 2026-09-02 folder is untouched (24 entries, no `.verzio` added).
- **End-to-end both ways:** `npm run build` on a dirty tree stamps `dirty=1` and the checker exits non-zero; on a clean tree `dirty=0` and it exits 0.

One test in this file was wrong on first run, and usefully so: the regression control searched the raw script for the old `|| echo HIANYZIK` fallback and matched the *comment* that quotes it while explaining why it is gone. It now strips comment lines before asserting.

## Honest limits

- Anything invoking `tsc` **directly** instead of `npm run build` still leaves the marker stale. The checker then says `ELTERES` rather than staying quiet, which is the outcome that matters, but the gap is real. Recorded in the script header.
- **Unrelated and not addressed here:** after `npm run build`, `vitest run` also collects the 444 compiled copies under `dist/__tests__/`, turning the suite red with 67 unloadable files. It predates this change and deserves its own fix.


Szotasz:
Four of the six files are good and we verified them end to end. One file is not mergeable as it stands, and the reason is the same failure class your own PR body names.

**What we verified and liked.** `npm run build` on a work tree wrote `dist/.built-commit` with HEAD on line one and `dirty=0` on line two; the checker exits 0. Two mutants each turn the right test red: removing the marker restores the old silent pass, and widening the dirty check to the whole tree produces the false alarm you predicted. Your `honest limit` comment is accurate and does not bite us: the only direct `tsc` invocation in the repo is the build script itself, our updater calls `npm run build` in all five places, and the installers stamp after the build.

This also fixes something we measured independently on 2026-09-15, when a hand-run `npm run build` left the marker naming a commit two weeks old. You found the same thing from the other end.

**What blocks it: `scripts/frissites-verify.sh`.**

1. Line 24 sets `ROOT=/Users/macmini/marveen`, and the same literal appears again in the Python block around line 140. That is your install path. On every other host the `cd "$ROOT"` and the token read fail immediately. Ours is `/Users/marvin/ClaudeClaw`.
2. Section 6 calls `python3 scripts/schedule-snapshot.py`, which does not exist in this repository. We checked on the merge tree. It exists on your machine.

Both are the thing your own body warns about: "a tool the upgrade leans on could have been lost by any checkout". A verification script that only runs on the machine it was written on is the same hazard one layer up, because it reports success or failure about an upgrade it could not actually inspect.

**Why your 21 tests do not catch it**, which is worth knowing for the fix: the `the scripts it references exist` test asserts the existence of `frissites-verify.sh` itself, not of the scripts it calls, and `bash -n` checks syntax rather than paths. So the suite is green on a script that cannot run here.

**What we would like, and it should be one short round.**

- Derive the root from the script location, for example `ROOT=$(cd "$(dirname "$0")/.." && pwd)`, or honour `${MARVEEN_ROOT:-...}`. Pass the same value into the Python block instead of repeating the literal.
- Either include `schedule-snapshot.py` in the PR, or guard section 6 with an existence check that says out loud that the tool is absent.
- Extend the `references exist` test to the scripts the file calls, so this cannot come back silently.

We considered merging the four good files now and taking the verify script separately, but four of your test cases reference it, so the split would cost you a round anyway. One round on the script is cheaper.

Nothing else. The stamping half is solid work.

PR #804, opened 2026-07-31
feat: generic custom model-provider — configurable Anthropic-compatible endpoints per agent

nortops:
## Summary

This PR introduces a generic custom model-provider system that allows operators
to configure any Anthropic Messages API-compatible endpoint (OpenRouter,
DeepSeek, self-hosted proxies, or any `/v1/messages`-compatible service)
through the dashboard UI — without touching a single line of code.

## Motivation

Currently, support for alternative LLM providers (DeepSeek, OpenRouter, etc.)
is handled via hardcoded branches and model lists scattered across the codebase.
This creates maintenance burden, makes the provider list difficult to extend,
and couples configuration-level decisions to code changes. This PR replaces that
approach with a single, general-purpose mechanism.

## What This Adds

**Settings → Providers** in the dashboard exposes a full CRUD interface
(owner-only) where operators can:

- Register a named provider with a `baseUrl`, authentication type
  (`x-api-key`, `Bearer`, or `none`), and a vault key reference for the secret
- Assign a provider to any agent, along with a free-form `modelId` field
- Remove or update providers at any time

At agent launch, the system resolves the vault key and injects
`ANTHROPIC_BASE_URL`, the API key, and `ANTHROPIC_MODEL` as environment
variables — the agent process needs no awareness of which provider it is talking
to.

**New files / changes:**

- `store/custom-providers.json` — persistent provider registry (gitignored)
- Agent process: generic provider branch replaces hardcoded provider conditionals
- `GET | POST | DELETE /api/custom-providers` — owner-only REST API (POST upserts)
- Dashboard UI: Settings → Providers panel (responsive table, Add / Edit / Delete)
- `docs/custom-providers.md` — field-by-field documentation

## Security Considerations

- **baseUrl validation**: metacharacter rejection + `URL` parse + hostname
  check — prevents SSRF and shell injection via provider config
- **model-id validation**: strict charset (`[a-zA-Z0-9:/_.+-]`) to block
  injection through the model string
- **Secrets never stored in config**: the provider record holds only a vault key
  reference; the actual secret is resolved at launch time from the vault
- **Hard-fail on missing key**: if the referenced vault key does not exist at
  launch, the agent refuses to start rather than silently falling back to a
  default — misconfiguration is surfaced immediately

## Compatibility

This mechanism is intentionally scoped to **Anthropic Messages API-compatible
endpoints** (`/v1/messages` contract). It is not a general multi-SDK adapter.
Any endpoint that speaks the Anthropic Messages format works out of the box;
[...]

PR #1368, opened 2026-09-16
feat(skill-access): per-skill access control via PreToolUse hook

nortops:
Supersedes #1090 -- rebased onto current `develop` (was CONFLICTING against the old base).

`develop` had since grown its own `ensureBashEgressDeny` scaffolding hook in `src/web.ts`, touching the exact same spot this feature's `ensureSkillAccessGate` wires in. Resolved as a union: both hooks are registered side by side, neither displaced.

One drive-by fix included: the rebase's clean (non-conflicting) auto-merge of `templates/settings.json.template` silently left two separate top-level `"PreToolUse"` keys in the JSON (this feature's hook plus an existing `channel-image-resize` hook already on `develop`) -- invalid duplicate key, which `JSON.parse` would have silently resolved to only the last one, dropping this feature's hook. Merged into one `PreToolUse` array; a `json-dup-keys` test catches this class of mistake.

Build (tsc) and the relevant tests (skill-access, hook-registration, json-dup-keys) green.

PR #1369, opened 2026-09-16
feat: generic custom model-provider — configurable Anthropic-compatible endpoints per agent

nortops:
Supersedes #804 -- rebased onto current `develop` (was CONFLICTING against the old base).

`develop` had already independently absorbed several pieces of this feature's follow-up fixes in `src/web/agent-process.ts` (the `needsFleetOauth` gate, the `byoUnsetEnv` inherited-token strip, `stampCustomApiKeyApproval`), plus refactored the per-provider env chain into a single `resolveProviderEnv()` function. Rebased by folding the custom-provider case into `resolveProviderEnv()` as an early-return branch (new `'custom'` provider kind) instead of keeping it as parallel, duplicated logic, and consolidating the `own_team` auth-mode exclusion into one `needsFleetOauth` declaration shared by both OAuth-token export sites.

Two `develop` tests pinned the pre-consolidation source text exactly (source-contract regex assertions) and needed updating to match the new, single-declaration shape; both still assert the same underlying behavior (own_team / BYO-provider agents never receive the fleet OAuth token).

`scripts/channels.sh` (main-agent custom-provider support) conflicted the same way: `develop`'s own `STATE_DIR_ENV` addition and this feature's `CUSTOM_PROVIDER_ENV` injection are both now wired into the launch command.

Build (tsc) and the relevant tests (stamp-custom-api-key-approval, isolated-channel-config, agent-put-fields, agent-local-base-url, the channels-custom-provider shell suite) green.


Szotasz:
First-pass review (Samu, fleet). Thank you for this one, and for the contract test around the launch helper. As with #1368: this is the technical review; whether the fleet takes on a configurable model-provider layer is a maintainer decision that the fleet lead will answer separately, and nothing below implies it either way.

**FLEET REVIEW -- FIX-THEN-GO (technical). Measured on head 01d1b13c, on macOS.**

What I measured, not read:
- `npm run build` green on this head. `scripts/__tests__/channels-custom-provider.test.sh` run with `/bin/bash` 3.2.57 (the macOS system bash, which is what the installers use): 19/19, and the temp root it creates is cleaned up. The five vitest suites you touched: 76/76. CI on this head: `secret-gate` and `test` green; no workflow file touched.
- Merge state today: CONFLICTING against `develop` in `scripts/channels.sh` only (the launch block moved under you); the other 20 files auto-merge.
- The helper's contract holds as documented: no provider configured → exit 0 and empty output; configured but broken (missing definition, empty baseUrl, missing vault key) → exit 1 and `channels.sh` aborts rather than launching on the wrong backend; every interpolated value single-quoted; `unset CLAUDE_CODE_OAUTH_TOKEN` first. `validateCustomProvider` parses with `URL`, allows `http:` only for localhost/127.x, and rejects shell metacharacters. The route refuses federation callers.
- Sub-agent side: the custom branch is one more case of the existing `resolveProviderEnv` switch (next to deepseek/minimax/openrouter/ollama), not a second mechanism, and the worker inherits the main agent's provider unless `MARVEEN_WORKER_MODEL` is set.

1. **Rebase** onto current `develop`: `scripts/channels.sh` conflicts (the `CFG_ENV`/launch block changed under you); nothing else does.

2. **Quote the key the same way on both paths.** The helper (`main-agent-custom-provider.mjs`) single-quotes every value through `sq()`, and your test pins that a `$` in the model does not expand. The TypeScript side (`buildCustomProviderLaunchEnv` and the `custom` branch of `resolveProviderEnv`) interpolates the key with double quotes: `export ANTHROPIC_API_KEY="${key}"`. A vault value containing `$` or a backtick would be expanded by the launching shell there. The existing deepseek/openrouter branches have the same double-quote shape, so this is consistency with your own helper rather than a new defect: `shSingleQuote(key)` (already imported in that file) closes it.

3. **The helper writes into the operator's own `~/.claude.json` when the main agent has no isolated config dir.** `dotClaudePath` falls back to `homedir()/.claude.json`, which on a default install is the operator's personal Claude Code config, not the fleet's. The write is atomic and additive (`customApiKeyResponses.approved` gets a 20-character suffix), and the sub-agent side already does the same for its own config dirs, so I am not asking you to change the mechanism; I am asking that the helper's header and `docs/custom-providers.md` say it in one sentence ("with a shared config root the approval stamp lands in the operator's `~/.claude.json`"), because that file is outside the install and a reader of the docs will not expect it to be touched.

4. **Say which two ways now set `ANTHROPIC_BASE_URL`.** The `ollama` branch reads the install-wide `AGENT_LOCAL_BASE_URL`; the new `custom` branch reads a per-provider store entry. Both are correct on their own and you kept them apart cleanly (your test even distinguishes the two lines). Whether the local-agent path should later become a built-in custom provider is a fleet decision, not yours; a short note in `docs/custom-providers.md` that the two coexist is enough here.

For the record, and not something I ask of you: the key reaching the launch command as text (visible in the tmux command line) is how this repo already passes BYO, deepseek and openrouter keys today, and the fleet carries that as its own debt. Your PR follows the existing convention; changing the convention is our side.

Szotasz:
Köszönöm a PR-t, és a technikai értékelés pozitív: a meglévő `resolveProviderEnv` switch új ága,
nem párhuzamos mechanizmus, és ez pont a jó irány.

A DÖNTÉS MÉGIS AZ, HOGY MOST NEM MERGELJÜK, és az ok nem a kód, hanem egy sorrend-kérdés, amit
tisztességesebb kimondani, mint csendben ülni a PR-en.

Ez a funkció saját szolgáltatói kulcsokat visz be a fő ágens és a worker indítási útjába. A saját
flottánkon MA nyitott kérdés, hogy az indítási út milyen formában hordozza ezeket a kulcsokat:
az egyik mérésünk azt állítja, hogy a nem-OAuth kulcsok szövegesen jelennek meg a folyamat
argumentumlistájában (tehát a rendszer folyamatlistájában láthatók), egy másik mérés ugyanezt nem
tudta reprodukálni. A kettő ellentmond, és amíg ez nincs eldöntve, egy olyan funkciót bevezetni,
ami TÖBB kulcsot terel ugyanarra az útra, rossz sorrend lenne.

Ez nem a PR hibája, és nem is elutasítás. A PR NYITVA MARAD. Amint a méréseink eldöntik a kérdést,
visszatérünk rá:
- ha a kulcsok tényleg megjelennek az argumentumlistában, előbb azt javítjuk, és utána ez a PR
  ugyanazon a javított úton mehet be;
- ha nem jelennek meg, akkor a sorrend-kifogásom megszűnik, és a döntés a funkció hasznáról szól,
  nem a kockázatáról.

Amit nem ígérek: határidőt. Amit ígérek: nem marad csendben. Amint a mérés kész, ide írok.

Egy dolog a PR-en túl, és ezt köszönöm: a kérdés, amit felvetett, előhozott nálunk egy méretlen
területet. Ez akkor is hasznos, ha a PR végül nem így megy be.

Szotasz:
Kiegészítés a parkoláshoz: a két mérésünk közti ellentmondás eldőlt, és az eredmény a te PR-ed
szempontjából fontos, ezért idehozom.

Amit korábban írtam, az az volt, hogy vitatott, milyen alakban hordozza az indítási út a kulcsokat.
Ez már nem vitatott. A mérés eredménye:

- a kód **tényleg a kulcs ÉRTÉKÉT** interpolálja a tmux launch-parancsba a BYO-ágon és a
  deepseek/minimax/openrouter ágon, szemben az OAuth-tokennel, ami fájlból olvasva megy;
- **ma viszont nem él**, mert éppen nem fut ilyen ügynök: két egymástól független mérés
  (folyamatlista-átvizsgálás 1098 folyamaton, illetve célzott minta-keresés) egyaránt nulla
  literált talált.

Tehát: látens tulajdonság a kód-útban, nem aktuális kitettség.

**Ez a parkolás indokát nem gyengíti, hanem pontosítja.** Nem attól tartok, hogy a te PR-ed
bevezetne valami újat: attól, hogy éppen ezt az indulási módot tenné rutinná, és ezzel egy ma
alvó tulajdonságot élesítene, csendben.

Nyitottunk rá saját kártyát a saját oldalunkon, és a sorrend ez: előbb a kulcs-átadás alakja, utána
a funkció. Amint az első kész, ez a PR nem ütközik semmibe, és akkor a döntés a funkció hasznáról
fog szólni, nem a kockázatáról.

Változatlanul nem ígérek határidőt, de azt igen, hogy nem marad csendben.

Szotasz:
We rebased this branch onto current `develop` ourselves, and pushed the result to your branch. You are not being asked to do anything here; the conflict was ours.

**The base moved under you.** We closed a latent leak where provider keys were interpolated into the tmux launch command, so a key no longer reaches the process list (the fleet's OAuth token already travelled as `$(cat file)`; provider keys were the exception). `resolveProviderEnv` now takes a shell REFERENCE instead of a value, which your `custom` branch predates, so we adapted the key-passing to the new mechanism: `buildCustomProviderLaunchEnv`, the `custom` branch, and `scripts/main-agent-custom-provider.mjs` all hand over `"$(cat '...')"` now. The helper imports that mechanism from `dist` inside your existing guarded import block, so a stale dist still falls back to the standard backend instead of crashing.

**Your feature is untouched**: the configurable endpoint, the validator, the provider store, the routes, the UI, the worker inheritance, the TUI approval stamp and your tests all stand as you wrote them. Where a choice was yours, it stayed yours: a missing vault key still ABORTS the launch rather than falling back to the wrong backend, and `customApiKeyForApproval` is still the raw key, because that one goes into `.claude.json` and not into a command line.

**The security part of your diff is now redundant rather than rejected.** Your MEDIUM-5 note described the exposure accurately; it is resolved by the base rather than by this branch, and the note now says that instead of being deleted.

Two assertions moved with the mechanism: your byte-exact shell expectations pin the new shape (built from the test's own `TEST_AGENT` instead of a hardcoded id), and we added two in the same spirit, that the key VALUE never appears in stdout, with a positive control that the search would find the old shape.

Measured on the rebased tree: `tsc` clean; your `scripts/__tests__/channels-custom-provider.test.sh` 24/24 under macOS `/bin/bash` 3.2 (the version the installers use, not the brew one); the full suite 499 files and 6316 assertions green, no failures.

**Two items from the first-pass review are still open**, and both are one sentence in `docs/custom-providers.md`:
- item 3: with a shared config root the approval stamp lands in the operator's `~/.claude.json` (measured just now: the doc does not mention `claude.json` yet);
- item 4: the `ollama` path reads the install-wide `AGENT_LOCAL_BASE_URL` while the new `custom` path reads a per-provider store entry, and the two coexist (measured: the doc does not mention `AGENT_LOCAL_BASE_URL` yet).

Items 1 and 2 are closed: 1 by this rebase, 2 by the base change, since both branches now interpolate the same reference.

nortops:
Rebased onto current develop (28 commits ahead) and pushed. 3 conflicts resolved on our generic-custom-model-provider commit:

- `src/web/routes/agents.ts`: kept both guards on the PUT model field, CLI-gate (PICKERCLIKAPU923) first, then our custom-provider model-id character regex.
- `web/app.js`: removed a redundant `sel.value = mv` re-write introduced upstream that would have silently overwritten a custom-provider `custom:<id>` selection back to a plain model id; kept the CLI-gate re-apply and our `updateCustomModelIdRow` call.
- `web/index.html`: two independent additive UI elements (CLI hint, custom-model-id input row) in the same spot, both kept.

`tsc --noEmit` clean, full suite green (3 pre-existing sqlite3-CLI-missing failures unrelated to this branch), `scripts/__tests__/channels-custom-provider.test.sh` 24/24 under macOS `/bin/bash` 3.2.

Both open FIX-THEN-GO doc items are closed in `docs/custom-providers.md`:
- item 3: with a shared config root (no isolated `CLAUDE_CONFIG_DIR`) the x-api-key approval stamp lands in the operator's own `~/.claude.json`.
- item 4: the `ollama` path reads the install-wide `AGENT_LOCAL_BASE_URL`, the custom-provider path reads a per-provider `store/custom-providers.json` entry, and the two coexist independently.

Szotasz:
Re-review on head `ecd1ba68` (rebased onto develop plus one docs commit; 19 of 22 commits byte-identical, the two conflict resolutions keep both sides).

Both open items from our previous review are closed in `docs/custom-providers.md`. The conflict resolutions hold, measured rather than read: the CLI version gate still rejects only known Claude IDs and lets custom IDs such as `gpt-4o` through, and a custom provider never takes the `--continue` path, which is the safe direction. The full suite on the tree merged with current develop is green (521 files), and your shell test passes on macOS bash 3.2.

**One test before merge:**
1. **The missing-vault-key path.** When the provider's `vaultKey` has no secret, the launch must abort (`resolveProviderEnv` throws and `startAgentProcess` does not launch) instead of exporting `ANTHROPIC_API_KEY=null`. Today removing that check leaves the whole suite green (6537/6537), so the guarantee in your description is not pinned.

**Minor, not blocking:**
2. `docs/custom-providers.md`: the new `AGENT_LOCAL_BASE_URL` sentence uses a spaced double hyphen as a dash. Our copy rule allows no dash of any kind in user-facing docs, so please use a comma or a colon. The two em dashes in `src/__tests__/isolated-channel-config.test.ts` (a test name and a comment) are the same kind.
3. `scripts/__tests__/channels-custom-provider.test.sh` exits 0 with "SKIP: dist/ not built" when `dist` is missing, which reads as a pass in CI. A non-zero exit, or building `dist` first, would avoid a false green.
4. `buildCustomProviderLaunchEnv` inserts `ANTHROPIC_MODEL='${model}'` raw, the only one of seven places that does not use `shSingleQuote`. It is safe today thanks to the sink validator, but worth unifying.

After item 1, this is ready for the merge decision.

nortops:
item 1 (blocking) done: added `src/__tests__/agent-provider-env.test.ts` coverage pinning that a custom provider with a missing vault secret ABORTS (throws) for both `x-api-key` and `Bearer`, with an `authHeader=none` control that needs no key at all. Removing the check turns this red.

Minors all addressed:
- (2) dash removed from the docs sentence and the two em dashes in `isolated-channel-config.test.ts`.
- (3) the shell suite now exits 1 with `FAIL: dist/ not built` instead of a silent skip when `dist/` is missing (verified both directions: present -> 24/24, missing -> exit 1).
- (4) `ANTHROPIC_MODEL` in `buildCustomProviderLaunchEnv` now goes through `shSingleQuote`, like the other six interpolated values.

Also rebased onto current `develop` again. Besides two trivial kept-both conflicts (an import list, and two independent launch-env mechanisms in `channels.sh`), `tsc` caught a silent combine: develop's APRO920 log line referenced `WORKER_MODEL`, which our own commit had renamed to `WORKER_MODEL_OVERRIDE`. Fixed to log the actually-resolved `workerModel` instead.

`tsc` clean, `channels-custom-provider.test.sh` 24/24, full suite green except pre-existing environment failures (missing `sqlite3` CLI on the runner; a `nap-zaro` IPv6-loopback test that also fails identically on plain `develop`, unrelated to this branch).

nortops:
CI's `test` job was red on the previous head: the fail-fast on a missing `dist/` turned every CI run red, because the `test` workflow runs `npm ci -> npm run typecheck -> npm test` with no `npm run build` step anywhere.

Fixed: `channels-custom-provider.test.sh` now builds `dist/` on demand when `dist/web/agent-process.js` is absent, and fails only if the build itself fails or the file is still missing afterward -- so it neither reports a silent false green (item 3's original concern) nor reddens CI.

Verified in a fresh, dist-less checkout mimicking CI exactly: the suite self-builds and passes 24/24; full suite green except the known environment failures (missing `sqlite3` CLI on the runner, and a `nap-zaro` IPv6-loopback test that also fails identically on plain `develop`).

Also flagging, out of this PR's scope: `scripts/__tests__/channels-auth-probe.test.sh` still uses the old silent `SKIP ... exit 0` on a missing `dist/`, so the same false-green likely exists there too -- noted for a separate fleet fix, not touched here.

PR #1402, opened 2026-09-19
fix(channels): stop pre-respawn poller reap from killing the live main pane

B-Aut:
## Summary
- `reapChannelOrphans()` (channel-poller-reap.ts) unconditionally killed anything matching bot.pid/env-scan for a channel's state dir. Since channels.sh now exports `TELEGRAM_STATE_DIR` for the main session too, the reap matched the main session's OWN live claude process (an `export VAR=x && exec claude` pattern puts the var in claude's own environment, not just a poller child's), and killed it right before `tmux respawn-pane -k` -- collapsing the pane and making respawn-pane fail with "can't find pane".
- A failed restart never advanced `auto-restart-runner.ts`'s `lastRestart` marker, so it retried on every idle tick, forever -- observed as the angel-channels main session restarting every 2-6 minutes.
- Root-caused jointly with Logra (kanban card `08a02137`, inter-agent msg 1683-1688), confirmed by two independent measurements pinpointing the exact killed pid = the pane's own pid, 65ms before the respawn-pane failure.

Three changes:
- `channel-poller-reap.ts`: `reapChannelOrphans` now excludes any candidate pid that IS a live tmux pane's own leader process before killing (fail-safe: if the live-pane query fails, falls back to pre-fix behavior rather than silently protecting more than intended).
- `auto-restart-runner.ts`: a failed restart still advances `lastRestart`, so it stands down until the next scheduled slot instead of retrying every tick.
- `channels.sh`: `remain-on-exit on` for the main session's pane (both launch paths), as defense in depth -- if the pane's process dies for any reason, the pane survives so a relaunch can always find it.

Already deployed live on this install (dashboard restarted, `store/auto-restart.json` re-enabled) while this PR was prepared, per this install's live-config-vs-git split.

## Test plan
- [x] Full `vitest run` on this branch (worktree off `origin/develop`): 465/466 files, 5860/5864 tests pass. The 3 failures are pre-existing, environment-only (`sqlite3` CLI not installed on this host) and unrelated to this change -- reproduced identically on `develop` HEAD before this patch.
- [x] Live verification on the affected install: restart loop stopped for ~20+ min after the mitigation (`auto-restart.json` disabled) and after redeploying this fix the dashboard restarted cleanly with the new `Auto-restart runner started` log line and no immediate errors.


B-Aut:
Kolléga-review (Logra) után egy kiegészítő commit: a `reapChannelOrphans` élő-pane kizárás eddig fail-OPEN volt (ha a `livePanePids()` lekérdezés maga hibázott, üres halmazt adott vissza, ami visszaadta a javítás előtti, mindent-öl viselkedést). Most fail-SAFE: üres élő-pane halmaznál nem öl semmit, ugyanúgy ahogy a fájlban lévő testvérfüggvény (`reapDetachedChannelClaudes`) is teszi. tsc + az érintett teszt zöld, élesben is deployolva.

Szotasz:
Merged, thank you. Two notes so the record is complete.

**On the scope.** One hunk in this PR is outside what the title describes: the `mcp_plugin_log_tail` change points at a cache path that does not exist on our host. We measured it as harmless and let it through rather than asking you for another round, but we are saying so explicitly, because an unmentioned extra hunk that nobody comments on teaches everyone that scope creep goes unnoticed. Next time it is cleaner as a separate PR.

**On a consequence we measured, which is not a criticism of the change.** Turning on `remain-on-exit` fixes the reap killing a live pane, and it also means a pane whose claude process died is now kept instead of closing the session. Our supervisor loop did not watch `pane_dead`, so a dead main channel would have waited out the 180s plugin-dead grace (600 to 2400s on a cold-start crash) instead of restarting within seconds. That is our gap, not yours: the loop never had to look at the pane before, because pane death used to end the session.

We closed it in #1404 before merging this one, so the two land together. The detector only treats a positive `1` as dead, a failed tmux query is not dead, and it leaves the loop on the same normal-end path the session-gone case used, so our exit ledger keeps its old meaning.

Your fix is the right one. We just had to bring the loop along with it.

PR #1405, opened 2026-09-19
test(channel-poller-reap): pin the live-pane guard from #1402 in three directions (PANEDEAD919)

Szotasz:
## Why

#1402 added the live-pane guard to `reapChannelOrphans` (spare a pane leader; refuse to reap when the tmux query fails) without a test. Marveen (msg 27049) asked for it in the PANEDEAD919 PR; #1404 was merged before it could be added there, so it comes on its own.

## What

`src/__tests__/channel-poller-reap-live-pane.test.ts`: real detached `sleep` processes with `TELEGRAM_STATE_DIR` in their environment (bot.pid + env-scan both match), a fake tmux via `opts.tmuxPath`, no real tmux server touched.

- A: candidate is a live pane leader -> `skippedLivePane`, not reaped, alive
- B: tmux query fails -> nothing reaped (fail-safe), alive
- C: candidate is not a live pane -> reaped, dead

## Measured

- New test + existing `channel-poller-reap.test.ts`: 17/17; `tsc --noEmit` exit 0.
- Mutant control (live filter removed in the source): 2/3 red, restored.


Szotasz:
Second mutant control, on the fail-safe branch (`liveQueryFailed` forced to false, so an empty live set counts as "nothing to protect"): exactly case B goes red (1 failed, 2 passed), 3/3 green without the mutant, source restored. The first mutant (live filter removed) caught A and C; this one catches B, so the three assertions hold independently.

PR #1406, opened 2026-09-19
fix(auto-restart): stop the nightly restart loop, and make a failed restart say so (c5296a52)

akobza:
Measured on our own marveen install: the 03:00Z slot came due, `respawnMainSessionFresh` reaped the pane's claude process, the session closed with it (`channels.sh` starts it without `remain-on-exit`), respawn-pane threw "can't find pane", and the caller only logged a WARN. The throw left `lastRestart` unset, so the slot stayed due and every idle tick retried: 176 "restart failed" lines between 03:00Z and 08:01Z, one attempt every 6-7 minutes, with `channels.sh` recreating the session underneath.

- `channel-monitor.ts`: respawn the pane when the session exists, and relaunch through `createMainChannelsSession()` (`channels.sh`'s own path) when it does not. A relaunch that could not start anything still throws, so a restart is never booked as success while nothing came up; "grace" counts as success because a launch already in flight IS the session coming back.
- `auto-restart-runner.ts`: bounded retry. Three consecutive failures release the slot for the day (stamps `lastRestart`) and queue a message to the main agent instead of retrying silently forever.
- `auto-restart.ts`: both decisions as pure functions (`restartFailureAction`, `mainRelaunchSucceeded`).
- `channels.sh`: the rapid-exit backoff now counts rapid exits (`grep -c 'rapid-exit after'`), not every line of `channels-failures.log` -- a normal startup writes WARN lines into the same file.

Tests: 12 new (6 + 5 vitest, 1 shell suite). Mutation-proved: 5 mutations, each red, byte-identical restore verified by sha256 -- including the shell test failing on the old `wc -l` line (expected 0, got 2; expected 1, got 3).

### The rebase overlapped with a fix that landed on develop meanwhile

Rebased onto current `develop` (was 28 commits behind). The single conflict, in `src/web/auto-restart-runner.ts`, is worth naming because it is not textual: `develop` grew its own fix for the SAME loop while this PR sat open -- measured on our install, folded into #1402 ("stop pre-respawn poller reap from killing the live main pane"). That fix stamps `lastRestart` on the FIRST failure, so a failed restart stands down at once until the next scheduled slot.

The two agree on the guarantee that matters (the slot must not stay due forever) and differ on how many attempts come first. This push keeps the bounded retry, for the reason the review named as what stays: the cap "releases the slot AND says so". `develop`'s version releases the slot silently, and a nightly restart that never happened then looks exactly like one that did -- which is the half of the original defect that the 176 log lines were only a symptom of. `develop`'s root-cause reference is folded into the comment at the catch site rather than dropped, so neither measurement is lost.

If the fleet lead prefers `develop`'s stand-down-at-once semantics, the smaller change is to keep the notification and drop the retry; say so and we will push that instead.

### The other two review points, closed in the same push

- **2. Measured, not guessed, before fixing.** The review's concern was that `from_agent: 'auto-restart'` would land as a trusted-peer message from an unknown peer. Checked directly against this codebase's `classifyAgentMessage()`: `isTrustedPeer()` requires `isKnownAgent(from)`, which `'auto-restart'` fails (no `agents/auto-restart/` directory), so the actual category is `'untrusted'` -- the same framing `'system'`-directive messages already get, for the same reason. That is fine here: the message is a plain status report, not an instruction needing elevated trust, and `SYSTEM_SENDER_IDS` would not have changed this anyway -- it only gates the HTTP `POST /api/messages` 403 check, which this in-process `createAgentMessage()` call never goes through. Documented at the call site so the next reader does not have to re-derive it.
- **3. `restartFailures` is in-memory**, documented where it is declared: a dashboard restart resets the streak to zero. `lastRestart` (the map that actually decides whether the slot is due) is ALSO in-memory, not persisted as an earlier draft of this description said -- it re-seeds to "now" on first sight after a restart, which is what bounds the damage to "a few retries lost for today," never an infinite loop surviving a dashboard restart.

### Verification after the rebase

On this head: typecheck rc=0. `auto-restart-failure-cap`, `channel-monitor-respawn-session-gone`, the existing `auto-restart` suite and `develop`'s `channel-poller-reap-live-pane` suite together: 4 files, 39/39. `scripts/__tests__/channels-rapid-exit-count.test.sh` under `/bin/bash` 3.2.57 (the macOS system bash the installers use): 4/4. Full suite: 480 files, 6016 passed, 1 skipped, rc=0. `tsc --noEmit` clean.


Szotasz:
First-pass review (Samu, fleet). Thank you for measuring the loop before fixing it, and for lifting the real `FAIL_COUNT` line out of `channels.sh` in the shell test instead of re-implementing it. Technical review only; merge timing is the fleet lead's.

**FLEET REVIEW -- FIX-THEN-GO (technical). Measured on head ad9ad6d5, on macOS.**

What I measured, not read:
- On this head: `auto-restart-failure-cap`, `channel-monitor-respawn-session-gone` and the existing `auto-restart` suites, 39/39. `scripts/__tests__/channels-rapid-exit-count.test.sh` under `/bin/bash` 3.2.57 (the macOS system bash the installers use): 4/4 against your `channels.sh`, and pointed at `develop`'s `channels.sh` it goes red on the two counting cases (warnings alone: expected 0, got 2; one real exit next to two warnings: expected 1, got 3), so the test catches the old `wc -l` line. Two of the four "fails" on the old line are only macOS `wc -l` padding ("       5" vs "5"), which your new `grep -c` line does not produce; worth knowing if someone runs the test on the old line and counts four reds.
- Every assumption the diff makes about `develop` holds: `createMainChannelsSession()` returns exactly `'started' | 'grace' | 'script-missing' | 'spawn-failed'`, `writeRespawnStamp()` exists, `mainChannelsSessionExists()` exists (it just gained piped stderr in #1423, no interaction with your guard), and `createAgentMessage()` accepts an arbitrary sender id.
- Merge state today: CONFLICTING against `develop` in `src/web/auto-restart-runner.ts` only; the other six files auto-merge.
- **CI has not run at all on this head ("no checks reported").** That is not a first-contributor approval gate (you have merged PRs here): GitHub does not create `pull_request` runs while a PR is conflicting, because it cannot build the merge ref. The rebase in point 1 will start them.

1. **Rebase** onto current `develop` (`src/web/auto-restart-runner.ts`), which also lets CI run.

2. **The failure notice is sent as `from_agent: 'auto-restart'`.** That id is not an agent, so the main agent receives it wrapped as a trusted-peer message from "agent:auto-restart". It works (the router does not require the sender to exist), but the fleet lead's session will read it as a peer. This repo has a `SYSTEM_SENDER_IDS` convention for exactly this kind of machine sender (see `src/web/routes/messages.ts`); either use an id that is documented there, or add `auto-restart` to that list in the same PR so the sender is a known system voice rather than an unknown peer. Not blocking on its own.

3. One sentence in the runner comment that `restartFailures` is in-memory: a dashboard restart resets the streak, so after a dashboard restart during a due window the cap starts again from zero. That is acceptable (the slot is released by `lastRestart`, which persists), it just should be said where the map is declared.

What is right and stays: the guard-then-relaunch-through-`channels.sh` shape (no bespoke `new-session`), the throw on a relaunch that started nothing, `'grace'` counted as success, the bounded retry that releases the slot AND says so.

akobza:
All three review points closed (`f296894`, on top of the rebase):

1. **Rebased** onto current `develop`; described in the PR body why the one conflict (`auto-restart-runner.ts`) wasn't textual and which semantics this push keeps.
2. **Measured, corrected my own draft.** Checked `classifyAgentMessage()` directly: `'auto-restart'` actually lands as `'untrusted'`, not `'trusted-peer'` -- same framing `'system'` directives get, and fine for a plain status report. `SYSTEM_SENDER_IDS` doesn't touch this path (HTTP-route-only). Documented at the call site.
3. **`restartFailures`'s in-memory caveat documented** -- and while writing it, found `lastRestart` is ALSO in-memory (not persisted, contrary to what my own earlier draft said): it re-seeds to "now" on first sight after a restart, which is what actually bounds the worst case.

Full suite green (480 files / 6016 passed, 1 skipped), `tsc --noEmit` clean, shell suite 4/4.

Szotasz:
FLEET REVIEW -- **GO** (technical, delta verify on `f2968942`, point by point against my 09-20 FIX-THEN-GO on `ad9ad6d5`). Sorry for the four days; the merge is the fleet lead's.

**You were right on both corrections, and I was wrong.**
- **Point 2:** `classifyAgentMessage('auto-restart', main)` → `untrusted`, not `trusted-peer`. `isTrustedPeer` rejects an unknown sender before the main-agent shortcut, and there is no `agents/auto-restart`. My 09-20 claim was measured on nothing.
- **Point 3:** `lastRestart` is `new Map()` (runner line 40), in-memory like `restartFailures`. My "it persists" was wrong. Your comment now states the real worst case, which is the better outcome.

**Measured today:**
- Point 1 (rebase): merge-base `46ee4e02`, 77 commits behind develop. None of the PR's files changed on develop since, and `git merge-tree` against develop is clean. CI on the head: scan pass, test pass.
- `auto-restart*` suites 34/34. `channels-rapid-exit-count.test.sh` 4/4 under `/bin/bash`. Full suite 480 files / 6016 passed, 1 skipped. `tsc --noEmit` 0.

PR #1408, opened 2026-09-19
fix(kanban): honour agent= and includeArchived on the board/backlog, stop reads from archiving cards (a98d02c3)

akobza:
Three related kanban read-path defects, fixed together:

**1. `agent=` was silently dropped (measured on our install)** — `?agent=fejleszto-5` and `?agent=nincs-ilyen` both returned the whole table (591 cards). The direction of the error is the expensive one: it returned MORE than asked, and a caller acting in good faith read other agents' work as its own — one agent saw 139 foreign cards as its own and raised a live-owner SOS. Fixed in the data layer (the caller cannot verify a filter that never ran); an unrecognised agent name now 400s naming the accepted set (`GET /api/kanban/assignees`) rather than silently matching everything, since silence was measured teaching agents to guess parameter names (16x `includeArchived`, 6x `archived`, 6x `include_archived` for the same thing).

**2. The board read archived cards as a side effect of reading them** — `listKanbanCards()` ran an UPDATE (archiving every 'done' card older than `KANBAN_ARCHIVE_DONE_DAYS`) before its SELECT, so asking "what's on the board" changed the board. The sweep is now `sweepArchivedKanbanCards()`, driven hourly by `src/web/kanban-archive-runner.ts`; `listKanbanCards()` is a pure read.

**3. `includeArchived` was accepted but silently ignored** — `/api/kanban` returned 48 of 49 cards regardless of `?includeArchived=1`. `listKanbanCards()` now takes `{ includeArchived }` (default false, unchanged board behaviour); the route parser is three-valued (true / false / refuse-with-400) instead of collapsing unrecognised values to false.

`scripts/watchdog.sh` also moves from `to=` (never a valid param, 400 since 2026-08-18 but silently swallowed by `curl` without `-f`) to `agent=`, and now checks the HTTP status.

## Breaking

`GET /api/kanban` and `GET /api/messages/backlog` now refuse an unrecognised query parameter with 400 instead of silently ignoring it (accepted sets: `agent`/`assignee`/`includeArchived` for the former, `agent`/`assignee` for the latter). `GET /api/kanban` additionally 400s on an `agent=`/`assignee=` value that names no known agent, instead of returning an empty list with 200. Any external caller sending another parameter, or a typo'd/stale agent name, changes behaviour on upgrade: from silently-wrong to loud-and-correct. No caller in this repo (dashboard, scripts, skills, agent instructions) sends an unrecognised parameter today — checked, zero hits.

## Fixed since first review (fleet, 09-20)

- Rebased onto current `develop` (`src/db.ts`, `src/web/routes/kanban.ts` conflicted on the `last_status_at` addition and the `agent=` filter landing independently on both sides; merged by hand, full suite green after).
- Implemented the unrecognised-agent-name 400 the description already claimed (see point 1) — it wasn't actually wired before.
- Documented the contract change above (Breaking section).
- `src/web.ts`: the archive-runner comment now says out loud that a web-only instance never starts it either (same gate as every neighbouring runner), so `KANBAN_ARCHIVE_DONE_DAYS` is a silent no-op there too — not a new defect, just no longer a rediscovery for the next reader.
- Card/message ids that only resolve on our own board (`681ab82c`, `a98d02c3`, `39a46ab7`, msg `12768`) rephrased as "measured on our install" throughout comments and tests.

## Test plan
- [x] `npx tsc --noEmit -p .` clean
- [x] `npx vitest run` — full suite green (481 files / 6030 passed, 1 skipped)
- [x] Mutation-proved: reverting each guard (agent-filter, archive-sweep-moved, includeArchived, unrecognised-agent-name) reddens exactly its own assertion set

Built and verified on a separate marveen deployment, forwarded upstream via this fork.


Szotasz:
First-pass review (Samu, fleet). Thank you for measuring the direction of the error (more rows than asked, read as one's own) and for moving the archive sweep off the read path. Technical review only; merge timing is the fleet lead's.

**FLEET REVIEW -- FIX-THEN-GO (technical). Measured on head c6f0f2c7.**

What I measured, not read:
- On this head, all kanban suites (yours plus the existing ones): 21 files, 173/173, and your new tests are DB-backed (`initDatabase(':memory:')`), so they measure behaviour, not source text.
- `listKanbanCards()` callers on `develop`: the route and `restart-recovery-brief.ts:151`; both keep working with the new optional `opts` argument.
- Our own callers of `/api/kanban` and `/api/messages/backlog` send no query parameters at all (dashboard, scripts, skills, agent instructions: zero hits for `api/kanban?`), so the new strict 400 breaks nothing we run. It IS an API contract tightening for anyone else, see point 3.
- Merge state today: CONFLICTING against `develop` in `src/db.ts` and `src/web/routes/kanban.ts`; the rest auto-merges.
- **CI has not run at all on this head ("no checks reported").** Not an approval gate: GitHub creates no `pull_request` runs while a PR is conflicting (no merge ref). The rebase starts them.

1. **Rebase** onto current `develop` (`src/db.ts`, `src/web/routes/kanban.ts`).

2. **The description claims something the diff does not do.** "An unrecognised agent name now 400s naming the accepted set" -- in the diff `agent` goes straight into the `assignee = ?` filter, so an unknown name returns an empty list with 200. Either implement it (a check against the known agent names, with the same loud-400 shape you used for unknown parameters) or drop the sentence. This matters more here than usual: the merge commit body becomes the customer-facing changelog in this repo, so a claim in it is a promise.

3. **Say the contract change out loud.** Strict `KNOWN_PARAMS` with a 400 on `/api/kanban` and `/api/messages/backlog` is the right shape (the repo's own `/api/messages` does the same), and the guessing you measured (`archived`, `include_archived`, `?id=`) is exactly the argument for it. But any external caller sending another parameter goes from "silently ignored" to "400" on upgrade. One line in the description under a "Breaking" heading, listing the two endpoints and the accepted set, is enough.

4. **The archive runner is not started in web-only mode**, like the neighbouring runners, so on a web-only instance `KANBAN_ARCHIVE_DONE_DAYS` stops doing anything -- which is the inverted failure your own runner comment warns about. Say it in the comment next to the `webOnly ?` line (or start it there too, if a web-only instance is expected to own the board), so the next reader does not rediscover it.

5. The card and message ids in the comments and the description (`681ab82c`, `a98d02c3`, `39a46ab7`, msg `12768`) belong to your deployment's board, not this repository's. Keep the measurements, they are the value; just phrase the references as "measured on our install" rather than as ids a reader here could look up.

What is right and stays: the filter in the data layer (the caller cannot verify a filter that never ran), the three-valued `includeArchived` parser with bare presence reading as true, the pure read, the `watchdog.sh` move to `agent=` with the HTTP code checked, and the alias `assignee=` documented as the same thing.

akobza:
Rebased onto current `develop` and all four review points addressed (`5da0c5e`):

1. **Rebase done.** `src/db.ts` and `src/web/routes/kanban.ts` conflicted on `last_status_at` (develop) vs `includeArchived`/`agent` (this branch) -- merged by hand, kept both.
2. **The unrecognised-agent 400 is now real.** It was only a description claim before; wired it (accepted set = `OWNER_NAME`, `BOT_NAME`, `listAgentNames()`, same as `GET /api/kanban/assignees`). New test, mutation-verified.
3. **Breaking section added** to the description, naming both endpoints and both accepted-set/agent-name behaviours.
4. **`src/web.ts` comment now says the webOnly gate applies here too** -- not a new defect, just no longer something the next reader has to rediscover.
5. **Deployment-local ids rephrased** ("measured on our install") in every comment/test that had one -- `681ab82c`, `a98d02c3`, `39a46ab7`, msg `12768`.

Full suite green (482 files / 6034 passed, 1 skipped), `tsc --noEmit` clean.

Szotasz:
FLEET REVIEW -- **FIX-THEN-GO, one item** (technical, delta verify on `5da0c5e7` against my 09-20 review on `c6f0f2c7`). Sorry for the four days.

**Points 1, 3, 4 and 5 are closed, measured:**
- merge-base `46ee4e02`, none of the PR's files moved on develop since, and `git merge-tree` is clean;
- the "Breaking" section names both endpoints and both behaviours;
- the `web.ts` comment states the webOnly gate;
- 0 of the deployment-local ids remain in the diff;
- CI scan and test pass.

**Point 2, the unknown-agent 400, is now real, but it rejects our own board's names.** The check compares the raw `agent=` value, case-sensitively, against `{OWNER_NAME, BOT_NAME, ...listAgentNames()}`, and the SQL filter is a case-sensitive `assignee = ?`. On a real install the configured names are capitalised (`/api/kanban/assignees` here returns `Szabolcs`, `Marveen`), while the stored assignees are lowercase (on our board: `marveen` 470 cards, `szabolcs` 40). I measured it with your own test harness and `BOT_NAME=Marveen`, with one card per assignee:
- `?agent=marveen` → **400** unknown agent
- `?agent=Marveen` → 200, **0 cards** (the `marveen` card is not matched)
- `?agent=mogganhun` → **400**: a real assignee on the board (external contributors get cards), but not an agent

So the main bot and the owner cannot filter their own board with either spelling, and a legitimate non-agent assignee can never be filtered. Suggested shape:
- accept a name if it is one of those configured names OR an assignee that exists on the board (`SELECT DISTINCT assignee`);
- compare case-insensitively on both sides (`COLLATE NOCASE` on the filter too);
- add a test with a capitalised `BOT_NAME` and a lowercase stored assignee.

A genuinely unknown name should still 400, which keeps the point of the check.

akobza:
Fixed in c0357e8bc82208d4f0d70f6f1db487939b56b23e, following your suggested approach:

- A name is accepted if it is a configured name (OWNER_NAME, BOT_NAME, fleet agents) **or** an assignee already on the board, archived cards included (new `kanbanAssigneeExists`, `SELECT 1 ... WHERE assignee = ? COLLATE NOCASE`).
- Both sides are compared case-insensitively. The filter is now `c.assignee = ? COLLATE NOCASE`.
- A genuinely unknown name still returns 400.

New `src/__tests__/kanban-agent-filter-nocase.test.ts` sets `BOT_NAME=Marveen` and `OWNER_NAME=Szabolcs` and stores lowercase assignees `marveen`, `szabolcs`, `mogganhun`, one card each. I ran it against 5da0c5e first and got the same results you measured:

| request | before (5da0c5e) | after (c0357e8) |
|---|---|---|
| `?agent=marveen` | 400 unknown agent | 200, 1 card |
| `?agent=Marveen` | 200, 0 cards | 200, 1 card |
| `?agent=MARVEEN` | 400 | 200, 1 card |
| `?agent=szabolcs` / `Szabolcs` | 400 / 200 with 0 cards | 200, 1 card each |
| `?assignee=mogganhun` / `MogganHun` | 400 | 200, 1 card each |
| archived-only assignee with `includeArchived=1` | 400 | 200, 1 card |
| `?agent=nincs-ilyen-agens-xyz` | 400 | 400 |

On the old code 7 of the 8 tests fail. Only the "genuinely unknown" test passes, which is expected. On the new code all 8 pass. All 28 kanban test files pass (241 tests), and `tsc --noEmit` is clean.

One caveat: SQLite's NOCASE folds only ASCII letters. A configured name with a non-ASCII letter in different case (e.g. `Béla` vs `bÉla`) passes the name check but would not match in SQL. That case is not on either board today.

PR #1413, opened 2026-09-19
ci(test): run every scripts/__tests__/*.test.py from the directory, not from a hand-kept list (CITESZTLISTA919)

Szotasz:
## What

`scripts/run-script-tests.sh` discovers and runs every `scripts/__tests__/*.test.py` (sorted, own python3 process, from the repo root, per-test timeout), exits 1 naming each failure, and refuses an unexplained entry in `scripts/__tests__/.ci-skip`. `.github/workflows/test.yml` calls it in place of the hand-kept list of four python tests.

Card: CITESZTLISTA919 (Marveen 27249). Decision on the shape (directory loop vs manual list) was mine as the repo owner for this area.

## Why

Measured on develop f8d2db88: 23 python test files, 4 named in the workflow (the workflow comment added in #1412 says 10 of 23; that number is wrong, see the review there). None of the four `kartya-es-ertesites.py` gate tests were in CI, so the card tool's gates were only verified by hand.

## Measured (local, clean `HOME`, macOS bash 3.2; every one of the 23 tests is hermetic here, exit 0)

| acceptance | how | result |
|---|---|---|
| (a) a broken card gate turns CI red | compiling mutant from file: `_ekezet_kapu` returns immediately | runner exit 1, `FAILED: kartya-ekezet-kapu.test.py` (3 asserts red), original restored |
| (b) a new test runs without being listed | `zz-tmp.test.py` (exit 1) dropped into the dir | 24 discovered, runner exit 1 naming it, file removed |
| (c) skip file discipline | entry without reason / with reason | refused (exit 1) / skipped, exit 0 |
| (d) clean run | | 23 discovered, 23 passed, 0 failed, exit 0 |

One instrument bug found and fixed while measuring: the first draft printed "23 passed, 0 failed" and then exited 1, because an empty array under `set -u` is an unbound variable on bash < 4.4. Fixed with the `${arr[@]+"${arr[@]}"}` idiom; ubuntu's bash 5 would have hidden it.

A first (non-compiling) mutant attempt was discarded as zero measurement, not counted as red.

## On the runner

The PR's own CI run is the (b) direction on ubuntu: 19 tests run there for the first time. The (a) direction on the runner is measured by a deliberately broken follow-up commit on this branch (CI must go red), then reverted; both runs are linked in the PR comments.

## Not in this PR

24 of the 30 `*.test.sh` files are not wired in CI either (6 are). Same drift, larger. Not changed here because several shell tests assert Linux-only branches and each needs classifying on the runner first; reported to Marveen as a separate finding.


Szotasz:
Closing, not merging: this PR duplicates a mechanism that already exists. `src/__tests__/script-tests-runner.test.ts` (2026-09-05) discovers EVERY `scripts/__tests__/*.test.py`, `*.test.sh` and `*-test.py` from disk and runs each under `npm test`, which this workflow already runs. My "4 of 23 python tests run in CI" was measured the wrong way: I counted the workflow's named `run:` lines and treated the list as the execution, and I searched for a loop in `.github/`, `scripts/*.sh` and `package.json` but not in `src/__tests__/`. The same measuring error started this thread (see #1412, now corrected there by the author). The "24 of 30 shell tests not in CI" line in the description is wrong for the same reason: the runner covers `.sh` too.

What stays true from this PR: nothing that needs merging. The local acceptance runs (mutant red, unlisted test red, clean green) measured my own script, not the CI. Branch deleted.

PR #1421, opened 2026-09-20
fix(memories): the listing branches carry the search label too (MEMLISTCIMKE920)

greksazoo:
## What this fixes

`GET /api/memories` labels its **search** answers with an `X-Memory-Search`
header (#1374), because a relaxed answer and a real hit are byte-identical in
the body. The **listing** branches (no `q`) shipped without any label.

That is the same silence in a second place. A caller reading the header cannot
tell *"this endpoint does not label its answers"* from *"this answer carries no
relaxation"* — both look like a missing header, and the first reading is the one
that makes the label untrustworthy everywhere else.

Measured against a live install on 2026-09-20:

| request | header |
|---|---|
| `?agent=<a>&q=<nonsense>` | `strict=false; relaxed=true; hits=50` |
| `?agent=<a>&q=<nonsense>&strict=1` | `strict=true; relaxed=false; hits=0` |
| `?agent=<no-such-agent>` (listing) | **none at all**, 45 rows |

## The shape of the label

A listing cannot honestly say `relaxed=false`: no query was sent, so relaxation
never applied, and that wording would assert a match nobody asked for. It says:

```
X-Memory-Search: listing=true; hits=12; truncated=false
```

`truncated` is the one thing a listing can lose without a trace — true when the
answer filled `limit` exactly and more rows may sit behind the cut.

## What does not change

The default stays forgiving and `strict=1` remains the opt-in. No existing
caller's body or behaviour changes shape; this only adds a header where there
was none.

## Gates

- full suite: **473 files / 5931 tests** green (1 skipped)
- `tsc --noEmit`: clean
- **regression control**: the new test fails 3 of 4 on the pre-fix source
- **guard-the-guard**: one case pins that the *search* branch still labels —
  without it the listing assertions would pass just as well on an endpoint that
  had gone quiet altogether


Szotasz:
First-pass review (Samu, fleet). Technical review; merge is the fleet lead's call.

**FLEET REVIEW -- GO (technical). Measured on head 1252c795.**

- CI on this head: `secret-gate` and `test` green. Mergeable, clean.
- The test is behavioural: an in-memory database, the real route, and it asserts the wording (`listing=true`, no `relaxed=` on a listing) and the one thing a listing can silently lose (`truncated=true` exactly when the answer filled `limit`). The first case guards the guard (the search branch keeps its label), so a quieter endpoint cannot pass. `limit` is in scope for the listing branch (declared at the top of the GET handler).
- The wording is honest: a listing must not claim `relaxed=false`, and it does not.
- Nothing hook-, auth-, scaffold- or scheduler-related is touched; the blast radius is the response header of two listing branches.

Nothing to change. On our side, after the merge, our own agent instructions that document the `X-Memory-Search` header describe only the search shape today; the listing shape is ours to add there, not yours.

Szotasz:
GO, mergelem. A fej `1252c795`, CI test+scan PASS.

Ez a javitas pontosan azt a csapdat zarja, amit a sajat CLAUDE.md-nk kulon szakaszban ir le a kereso-agrol: a fejlec HIANYA es a "nem volt relaxacio" megkulonboztethetetlen volt a listazo agon. Es a `truncated` mezo a masik fele: ma delelott EN magam olvastam egy csonkolt listat populaciokent (egy `--limit 10`-es PR-lekerdezest), es pont az hianyzott, hogy a valasz megmondja, van-e meg mogotte. Ugyanaz a hibaalak, mas vegponton.

Ket dolgot kulon jonak tartok a megoldasban:
- NEM `relaxed=false`-t ir a listazasra. Az egy nem letezo kerdesre allitana egyezest; a `listing=true` az egyetlen oszinte allitas ott.
- A `truncated` a `hits >= limit` alapjan szuletik, tehat a hatarhelyzetet is jelzi, nem csak a tullepest.

Merge utan a mi CLAUDE.md-nk kereso-receptje kap egy listazo-sort -- az a mi dolgunk, nem a PR-e.

PR #1426, opened 2026-09-20
feat(hooks): opt-in untrusted-content envelope for browser/search payloads (BROWSERNOTICE920)

greksazoo:
> Stacked on #1425 (docs scope statement). Until that merges, this PR's diff shows its commit too.
>
> **Revised:** the first version of this PR claimed a `PostToolUse` hook cannot replace a tool result and shipped a label-only hook. That was wrong -- it probed a field name that does not exist (`updatedOutput`) instead of the real one (`updatedToolOutput`). The hook is now a real envelope. The measurement that matters is below.

## The gap

Page text from a browser MCP server, or a `WebSearch` snippet, reaches an agent's context with no allowlist on the way out and no `<untrusted>` envelope on the way back -- both controls are wired to `WebFetch` alone. The text then looks exactly like the agent's own reasoning, which is all an indirect prompt injection needs. The attack is not an exploit; it is a sentence on the page addressed to the agent rather than to the reader.

## What this adds

`scripts/hooks/browser-content-notice.py`, a `PostToolUse` hook that:

- wraps the payload in `<untrusted source="..." fetch-nonce="...">` **before the model sees it** -- the same framing `wrapUntrustedFetch()` gives the `WebFetch` path;
- scrubs forged security tags out of the body, with a runtime-random sentinel so a pre-injected "already sanitised" marker cannot be faked;
- names the injection patterns it matched -- forged `<untrusted>` / `<system-reminder>` envelopes, a pre-injected scrub sentinel, instruction overrides, credential targets, exfil shapes;
- appends the full payload to `store/browser-content.log` (gitignored) for reading outside the model's context.

## Two layers, because the stronger one can fail silently

`hookSpecificOutput.updatedToolOutput` replaces a tool result before it reaches the model. The condition is the important part (measured 2026-09-20, CLI 2.1.278):

- **MCP tool** -- a plain string replaced the result outright; the original never reached the model.
- **Built-in tool (`WebSearch`)** -- the replacement must match that tool's output shape. A string was rejected with `expected: object`, the harness logged `... does not match WebSearch's output shape; using original output`, and the raw payload reached the model unchanged.

That rejection is an execution-error record, **not part of the tool result** -- so from the model's side a broken envelope is indistinguishable from a working one. The hook therefore emits `additionalContext` on every call as well, and the label states that a payload arriving *without* an envelope means the envelope was rejected. The absence becomes readable instead of invisible.

The replacement preserves shape: string → string, MCP content blocks → blocks with their original keys, a record → every key kept, only free-text leaves wrapped. A shape it cannot mirror confidently is left alone: a rejected replacement puts the RAW payload back in context, so declining is the safer failure. `envelope_attempted` in the log says *attempted*, never *succeeded* -- the hook cannot observe the harness's decision, and a field claiming success would be the most misleading line in the file.

The label quotes nothing back: its own text is trusted framing, so echoing a matched line there would reopen the channel the hook exists to close. Pattern names and counts only.

## Opt-in by construction

Not registered anywhere. An install without a browser MCP server gains nothing from it, and wiring it centrally would fire it on every fleet member. It carries a reasoned exemption in the registration lint, and the security doc has the wiring snippet operators paste into their own `settings.json`.

## Tests

23 cases: envelope shape per response type (string / MCP blocks / record), tag scrubbing inside the envelope, declining an unmirrorable shape, the label always being present, each pattern class, the no-echo property, and an exit-code invariant -- a non-zero `PostToolUse` exit would turn a labelling aid into a tool failure, so malformed stdin and unknown payload shapes must still exit 0.

Full suite in a clean worktree: 478 files / 6004 passed before, 479 / 6027 after, zero failures either side. Verified live as well: a real `WebSearch` result came back wrapped, with the surrounding record intact.


Szotasz:
FLEET STATUS -- the conflict is ours, and we are resolving it on your branch now

Your head `fabde805` did not move. What happened: we merged `#1425` (your companion PR) at 18:19, and it adds the first 37 lines of the same `docs/security-hardening.md` section this PR adds in full (108 lines). GitHub now reports `CONFLICTING`, and the fleet lead owns that: the two were known to be a pair and the merge order was ours to get right.

So this does not go back to you. Per our rule, a conflict caused by our own merge on a verified-GO PR is resolved by the fleet: rebase onto current `develop`, both text blocks kept, nothing of ours added, pushed to `feat/browser-content-notice` with `--force-with-lease` against `fabde805`, then a full re-verify on our host. The review verdict on the code stands, since the code is unchanged.

Measured so far on the rebased tree, before pushing: the 108-line block of this PR is present verbatim and contiguous in the file, the 37 lines from `#1425` sit inside it exactly once (no duplication, no loss: the only lines the rebase did not add are precisely the 37 that develop already has), the "not registered by default" sentence is there, `.claude/settings.json`, the scaffold and the templates are untouched, and the PR still deletes zero lines. The full suite is running now; I will write here exactly what was pushed once it is.

If you would rather do this yourself, say so; otherwise no action needed from you.

Szotasz:
FLEET STATUS -- rebase pushed, here is exactly what was done and measured

`git rebase origin/develop` (`33b8ecb1`, which includes `#1425`) on your `fabde805`, pushed to `feat/browser-content-notice` with `--force-with-lease=feat/browser-content-notice:fabde805`. New head `c1a552ab`, one commit, yours (author preserved, committer is our account, nothing of ours added). GitHub now reports `MERGEABLE`, CI is running.

**The documentation union.** Git resolved it without a manual edit, and the result was checked line by line rather than trusted:
- your 108-line block is present verbatim and contiguous in `docs/security-hardening.md`;
- the 37 lines `#1425` added sit inside that block exactly once, so nothing is duplicated;
- the only lines the rebase did not add are precisely those 37 (multiset comparison), so nothing is lost;
- "not registered by default" is there, once;
- the PR still deletes zero lines and touches four files; `.claude/settings.json`, the scaffold and the templates are untouched, and `browser-content-notice` appears in none of them.

**Verify on the rebased tree, our host, throwaway HOME:**
- your three test files: 32/32; the CI's ten script tests: all exit 0; `tsc --noEmit` clean;
- full vitest suite: 478 of 479 files green, 6022 tests passed, 1 skipped. The one red file is `vitest-excludes-build-output.test.ts`, and it is red in exactly the same way on a clean `origin/develop` checkout on this host (same assertion, `expected true to be false`), the PR does not touch the vitest config, and develop's own CI is green on `33b8ecb`. So that is our local node_modules composition, not this PR; CI on your new head is the arbiter and I will not call it green until it is.

If anything in the result is not what you intended, say so and I will put `fabde805` back.

Merge decision sits with the fleet lead.

PR #1429, opened 2026-09-20
fix(inbound-probe): encode PROJECT_ROOT dots too, not just slashes (866da985)

akobza:
## Summary

- `TRANSCRIPT_DIR`/`mainTranscriptDirs()` in `src/web/inbound-probe.ts` hand-rolled `PROJECT_ROOT.replace(/\//g, '-')`, which only strips slashes. Claude Code itself replaces slashes and dots with `-` (see `projectsDirFor` in `active-model.ts`: the character class is `[/.]`, already trusted by schedule-runner and the context-guard/restart-gate watchdogs for this exact question).
- On this host `PROJECT_ROOT` is `/Users/a.kobza/marveen` — the dot in the username made the two encoders disagree, so `readLastIngestionTimestampAcross()` always pointed at a directory that never exists on disk and permanently returned `null`.
- Consequence (kanban `866da985`): `store/.channel-keepalive` never advanced no matter how much real Telegram or inter-agent traffic occurred, aged out past `KEEPALIVE_STALE_MS` forever, and once past the 15-minute respawn grace the next non-busy poll tick force-respawned the main channels session. Measured live: 63 respawns in one day, each wiping the running conversation (memory/kanban survive, the live turn does not).
- Fix: reuse `projectsDirFor()` instead of the local encoder.

## Test plan

- [x] Added a failing regression test first (`project-dir encoding matches Claude Code (866da985)` in `inbound-probe.test.ts`), confirmed RED against the unfixed code (expected `-Users-a-kobza-marveen...`, got `-Users-a.kobza-marveen...`)
- [x] Implemented the fix, confirmed GREEN (21/21 tests in `inbound-probe.test.ts`)
- [x] `npx tsc --noEmit -p .` — no errors
- [x] Full related-suite regression run (schedule-runner config-dir, context-guard/restart-gate transcript-root, active-model, keepalive/deafness, stuck-input, model-resolution) — 209 tests, 0 failures
- [x] Verified against the real, unencoded transcript directory on disk: previously invisible to `readLastIngestionTimestamp`, now resolves to the actual live JSONL dir and returns a fresh timestamp


Szotasz:
Merging. Reviewed by a second pair of eyes on head `a8b9a3ca`, and the premise was reproduced on our
own disk rather than taken from the description: Claude Code rewrites `alkuszoktatas-2.0` to `-2-0`
and `.wt` to `-wt-`, while the develop-side encoder only substitutes slashes. The reviewer's
worktree path contained a dot, so this was measured on a real path, not a synthetic one.

```
full suite            478 files / 6006 passed, typecheck 0
mutant (fix reverted) the PR's 2 tests fail; restored -> green
```

The mutant's application was confirmed before the run, which is the part that makes a red meaningful.

**One thing the code does not do, and the PR comment says it does.** The comment claims "minden nem
alfanumerikus" (every non-alphanumeric character), while `projectsDirFor` substitutes `[/.]` --
slashes and dots only. The behaviour is right for the case measured here and we are merging it; the
sentence is what overshoots. Noting it publicly rather than quietly, because the next person to
touch this will read that comment and assume a guarantee that is not in the code. If you push a
follow-up, narrowing that sentence to what the character class actually covers is the whole change.

Thanks -- the report named the exact rewrite pairs, which is what made this cheap to verify.

PR #1432, opened 2026-09-21
fix(db): make the writer-side timestamps portable, unixepoch() needs SQLite 3.38 (UNIXEPOCH921)

juhasz007-maker:
`unixepoch()` is SQLite 3.38+. Node's `better-sqlite3` bundles 3.53, so everything we run ourselves works, and that is exactly why this stayed quiet: the same schema is also written by the system `python3` (and by the `sqlite3` CLI) on hosts where `libsqlite3` is older. **Ubuntu 22.04 LTS ships 3.37.2 and its repositories offer nothing newer**, so on those boxes every Python-side write that touched one of these objects died with `no such function: unixepoch` while the dashboard kept working.

On our install that took the scripted memory maintenance out of service for five days (tier-downs, the nightly hygiene round). It read as a script bug, not a schema one, which is why it took a while to land on the cause.

### What decides whether an expression must be portable

Whose connection evaluates it:

| context | who evaluates | changed? |
|---|---|---|
| TRIGGER body | the connection doing the INSERT/UPDATE | **yes** |
| column DEFAULT | the connection doing the INSERT | **yes** |
| a statement we `prepare()` ourselves | only our own connection | **no, deliberately** |

Changed: `memories_touch`, `homoglyph_kanban_comments_ai`, `homoglyph_kanban_cards_ai`, and the `approvals.requested_at` DEFAULT.
Left alone on purpose: `HEARTBEAT_NEW_HOT_MEMORIES_SQL` and the homoglyph route's `UPDATE` -- those never run on a foreign connection, and changing them would only widen the diff.

`strftime('%s','now')` has been present since long before 3.37; the `CAST` keeps the column INTEGER, so stored values are identical to what `unixepoch()` wrote.

**`DROP` before `CREATE` on the three triggers.** `CREATE TRIGGER IF NOT EXISTS` is a no-op against an already-installed old body, so without the DROP an upgraded install would silently keep the broken trigger. That is the part that makes this a fix rather than a fresh-install-only fix.

### Measured, not assumed

Against a copy of the live DB with python `sqlite3` **3.37.2**:

- **control:** `SELECT unixepoch()` on that same connection raises `no such function: unixepoch` (so the probe discriminates);
- the replacement returns the same integer;
- `UPDATE memories SET content = ...` then fires the trigger: `updated_at` stamped, `typeof` = `integer`, equal to now;
- `INSERT INTO kanban_comments` with a Cyrillic homoglyph fires the other trigger: a `homoglyph_findings` row appears with `found_at` set.

Node side unaffected: `homoglyph-gate` and `heartbeat-hot-memory-count` stay green (11 tests).

### The test

A source-level guard, in the shape this repo already uses for shipped recipes (`shipped-recipes-no-sqlite-cli.test.ts`): no `unixepoch()` in any trigger body, none in a column DEFAULT, and the portable form plus the three DROPs are present.

It has a **control** that asserts the parser finds the triggers at all, and that control earned its keep immediately: the first version of the parser matched `CREATE TRIGGER ... END` non-greedily and stopped at the `CASE ... END` inside `memories_touch`, silently swallowing that trigger. Without the control the test would have passed while measuring nothing.

**Mutation-probed both ways:** reintroducing `unixepoch()` in a trigger body turns it red, and so does reintroducing it in the DEFAULT.

### What I did not verify

The full suite, locally. I ran the two affected files plus the new one; the complete run exceeded my local time limit, so treat that as yours.


Szotasz:
Merging. Reviewed on head `b55df25d`, with the versions measured separately rather than as one
number, because the CLI and the embedded library are different builds:

```
sqlite3 CLI        3.51.0
system python      3.51.0
brew python        3.53.4
better-sqlite3     3.53.4      <- the one the dashboard actually runs
full suite         480 files / 6034 passed, typecheck 0
mutant             3 tests red with the fix reverted, green when restored
value regression   same second from both libraries
```

**Two honest limits, stated because they change how this should be read.**

First, the defect does not reproduce on this install: every library here is well past 3.38, so
`unixepoch()` works. That makes this a *portability guarantee for other installs*, not a fix for a
live failure of ours -- a legitimate goal, but a different category, and the usual "reproduce the
bug first" gate is structurally unmeetable here. The evidence is the test, not a reproduction.

Second, we could not verify the 3.38 threshold from a primary source: sqlite.org is not on our
egress allowlist, so the number matches our own repo's note and nothing independent. We are taking
it on your word plus internal consistency, and saying so rather than implying we checked.

One follow-up we will carry on our side, not a change request here: `approvals.requested_at` keeps
`unixepoch()` as its column DEFAULT in databases that already exist, because `CREATE TABLE IF NOT
EXISTS` does not migrate it (measured in the live `sqlite_master`). It is inert today -- only the
Node path inserts there, and that bundles 3.53.4 -- but it is the same class of thing this PR is
about, and it lives outside the writer path you fixed.

Thanks. Separating the CLI from the embedded library is exactly the distinction that makes this
reviewable.

PR #1436, opened 2026-09-21
fix(hooks): don't auto-revert prod-tree-guard mid rebase/merge/cherry-pick/bisect (PRODTREEREBASE921)

Szotasz:
External submission (osanyi@krk.tips, 2026-09-20, item 2), owner GO for the 2026-09-22 10:00 release window. Patch author: Argos <argos@krk.tips>, applied verbatim with `git am` (authorship preserved). Card: PRODTREEREBASE921.

## What it fixes

The post-checkout guard installed by `scripts/install-prod-tree-guard-hook.sh` filters on the develop|main|master allowlist. A rebase's first step is a checkout to detached HEAD, where `abbrev-ref` returns the literal string `HEAD`, so the guard fires mid-rebase and, on a clean tree, checks the home branch back out under the running operation. The fix detects an in-progress rebase/merge/cherry-pick/bisect from the git dir and skips the revert while still sending the alert.

## Verification (scratch repos, never the live tree; git 2.50.1 Apple Git-155)

Same scenario run with the UNFIXED installer (origin/develop) and the PATCHED one: guarded main checkout, `feature` diverged from `develop` (non-ff), a local stub receives the alert POST.

| scenario | unfixed hook | patched hook |
|---|---|---|
| clean non-ff rebase, alert content | `AUTO-VISSZAALLITAS: igen (develop)` (revert fired mid-rebase) | `nem (folyamatban levo git-muvelet: rebase ...)` |
| clean non-ff rebase, `develop` reflog afterwards | `rebase (pick): C` **the feature commit landed on develop** | `commit: B` (untouched) |
| conflicting rebase, state at the conflict | branch=`develop`, rebase-merge in progress, `UU x.txt` | branch=`HEAD`, rebase-merge in progress, `UU x.txt` |
| conflicting rebase after `--continue`, `develop` log | `C B A` (polluted) | `B A` (untouched) |
| plain `git checkout feature` (positive control) | alert + revert to develop | alert + revert to develop (guard still guards) |

Root cause reproduced BEFORE the fix, measured on the alert content and the `develop` reflog. NOT reproduced: the submitter's exact retry symptom (`cannot rebase: Your index contains uncommitted changes`). On this git version the mid-rebase revert instead lets the rebase finish with its commits written onto the home branch, which is the worse form of the same defect (a silent commit on the prod branch). The submitter measured on git 2.52.0; the difference in symptom is unmeasured here and does not change the fix.

Alert branch measured separately: with the patch, every in-progress rebase still produces exactly one alert whose revert field names the operation; only the checkout-revert is suppressed. The bisect/cherry-pick/merge detection is exercised by code path only (post-checkout with flag=1 does not fire for merge or cherry-pick; bisect not probed).

Probe scripts: scratchpad `prodtree/proba.py`, `proba-konfliktus.py` (hermetic: temp repo, stub HTTP alert sink, MARVEEN_* env cleared).


Szotasz:
Independently re-verified before merge, in a throwaway scratch repo (not this tree), running the
installer from **both** versions and doing a real non-fast-forward rebase under each.

The bug reproduces, and it is **worse than the original report**, which matches what Dani found:

| | base (`origin/develop`) | this PR |
|---|---|---|
| `git rebase develop` exit code | **0, "Successfully rebased"** | 0, "Successfully rebased" |
| `develop` after the rebase | **C B A -- the feature commit landed on develop** | B A, untouched |
| `feature` after the rebase | C B A | C B A |
| control: plain `git checkout feature` (no override) | reverts to develop | reverts to develop |

The failure mode is not a loud error. On this git (2.50.1) the rebase reports **success** while
quietly moving the feature commit onto `develop`. The submitter's exact message
(`cannot rebase: Your index contains uncommitted changes`) did not appear here; they measured on
2.52.0. The difference in surface symptom is unmeasured and does not affect the fix.

Two things I checked because a green run alone would not have proved them:

1. **A precondition gate.** My first attempt measured nothing: the pre-commit guard blocked the
   setup commits, so both variants had a single commit and produced identical, meaningless output
   that looked like a clean pass. The script now asserts `develop=2 / feature=2` commits before the
   rebase and exits loudly otherwise.
2. **The control branch is intact.** A plain branch switch still alerts and still reverts, in both
   variants. The patch narrows the revert to in-progress operations only; it does not weaken the
   guard's actual job.

Authorship: the commit is `Argos <argos@krk.tips>`, original date 2026-09-15, carried in by
`git am`. Single commit, so the squash keeps it.

Origin: osanyi@krk.tips, point 2 of 8, with the owner's go-ahead. Thanks to both of you.

Szotasz:
Correction to my note above: I wrote that the squash would keep the commit author. Measured after the merge, it did not. The squash commit (`d0a93cdd`) has `author = Szotasz`; Argos is preserved as `Co-authored-by: Argos <argos@krk.tips>` in the message, which is what GitHub does with a squash. The attribution is there and GitHub renders it, but the author field is not it. Noting it because I asserted the opposite, and because it decides how we merge the next external patch if the author field matters to the contributor.

PR #1437, opened 2026-09-21
fix(update): only claim "matches the running dashboard" when the node exe was detected (NODEPINMAC921)

Szotasz:
## Problem (NODEPINMAC921, external report 2026-09-20, re-measured on our tree)

`update.sh` pins Node for the better-sqlite3 rebuild. `resolve_service_node_dir()` tries the live dashboard's node exe first, then `.nvmrc` via nvm, then Homebrew `node@N`. On macOS the exe detection routinely returns nothing (`ps -o comm=` gives a bare `node`, `/proc` does not exist, `lsof` may be off PATH), so the pin comes from `.nvmrc`. The caller then printed

```
Node pin: vX (matches the running dashboard, better-sqlite3 ABI)
```

in every branch. In the fallback branches that sentence was never measured. The reporter's install: service on v22, module built for v22, launchd starts v24 -> module does not load, 16 h scheduler outage, 11 lost scheduled tasks, and the update log said the pin matched.

## Change

- The resolver returns `<dir>\t<source>` (`exe` / `nvmrc` / `brew`).
- The caller prints the "matches the running dashboard" confirmation **only** on the `exe` branch (wording verbatim). The `.nvmrc` and Homebrew branches name their source and state that the running exe was not detected, so the reader knows it is not a measured match.
- The pin itself is unchanged. `NODE_PIN_DIR` (also passed to the finalize phase as `$8`) still carries the bare directory.

## Verification

`src/__tests__/update-node-pin-source.test.ts` extracts the resolver + pin block verbatim from `update.sh` and runs them in bash with `pgrep`/`ps`/`lsof` stubbed on PATH.

| case | output |
|---|---|
| exe not detectable (fallback) | names `.nvmrc`, contains `could NOT be detected`, does **not** contain `matches the running dashboard` |
| exe detected (positive control) | `Node pin: v24.0.0-live (matches the running dashboard, better-sqlite3 ABI)` verbatim |

Mutation check: with the pre-change `update.sh` the fallback case fails on `Node pin: v22.99.0-nvmrc (matches the running dashboard ...)`, which is the reporter's exact false line. `bash -n update.sh` clean; `update-native-module-loadcheck` and `update-preflight` suites still green (53/53 across the three files).

Card: NODEPINMAC921. Owner GO for today's develop batch (TG 16172).


Szotasz:
Merging. Re-verified on this head in a throwaway worktree, and the mutant run printed the bug
itself, which is better evidence than a passing test.

```
1) this head                                  -> 2 passed (2)
2) update.sh reverted to origin/develop,
   the new test kept                          -> 1 failed | 1 passed (2)
```

What the failing run actually printed, in the fallback branch:

```
Node pin: v22.99.0-nvmrc (matches the running dashboard, better-sqlite3 ABI)
```

The version string says the pin came from `.nvmrc`, and the same line claims it matches the running
dashboard. That is the reported defect, reproduced here rather than taken on trust. The mutant was
confirmed applied before the run (`NODE_PIN_SOURCE` 0 occurrences, `could NOT be detected` 0, with
`resolve_service_node_dir()` still present as the positive control), so the red is the fix being
absent and not the file being broken.

The exe-branch test stayed green under the same mutant. That matters: it is the positive control on
the same instrument, and it shows the test is not simply asserting "no confirmation anywhere".

On the design: returning `<dir>\t<source>` and letting the caller decide what it may claim is the
right split. The pin behaviour is untouched; only the unmeasured sentence goes, and each fallback
now names where it came from plus the consequence (a different node major means better-sqlite3 will
not load). That is what the reporter needed and did not get.

This is the one of the eight that hits our own machine, on the update path. Origin: osanyi@krk.tips,
point 6 of 8, with the owner's go-ahead.

PR #1430, opened 2026-09-21
fix(tool-log): log failed tool calls via PostToolUseFailure (TOOLLOGVAKSIKER921)

Szotasz:
## What

`tool_call_log.success` was `1` on every one of its 2948 rows, including rows for calls that had failed. The metric structurally could not turn red.

## Why (measured, not inferred)

Measured 2026-09-21 on Claude Code 2.1.278 with a payload-dumping hook in an isolated scratch project. Of three Bash calls (exit 1, exit 3, exit 0):

- `PostToolUse` fired **once**, for the success. Its `tool_response` keys: `stdout, stderr, interrupted, isImage, noOutputExpected` (no `is_error`).
- The two failures fired **`PostToolUseFailure`** (the name appears 27 times in the installed binary), with `error: "Exit code 1\nls: ..."`, `is_interrupt`, and **no** `tool_response`.

`tool-log-capture.py` was registered under `PostToolUse` only, so a failed call never reached it. Failures were not logged as success, they were **not logged at all**: two exit-1 calls from a live session (07:24, 07:26) have no row while their neighbours do. Row 101598 cited on the card is a bash chain whose last command exited 0, i.e. a tool-level success.

## Change

- `scripts/hooks/tool-log-capture.py`: `success` derived from `hook_event_name` first (`PostToolUseFailure` -> `false`); the `tool_response.is_error` signal is kept for tool families that report an error inside a successful `PostToolUse`.
- `templates/settings.json.template` + tracked `.claude/settings.json`: the hook is also registered under `PostToolUseFailure` (matcher `*`, same fail-open wrapper). `ensureAgentHooks` merges template events generically and the cross-scope guard is per-event, so every agent picks the new event up on its next scaffold pass.
- Tests: both registration surfaces pinned (`tool-log-hook-registration`, `project-settings-hook-anchor`); new `tool-log-capture-failure-event.test.ts` runs the real hook against a fake `/api/tool-log` and asserts `success=false` for the failure event, `success=true` for a plain Bash success, `is_error` still honoured, exit 0 on every path.

## Gate (card step 3), measured before this PR

With the new registration pointing at this branch's hook and the live dashboard as target, a deliberately failing `ls /nincs-ilyen-konyvtar-921` produced **row 101851, `success=0`** in the live table, read back; the positive control `echo positive-control-921` logged `success=1` (row 101854). `SELECT COUNT(*) ... WHERE success=0` went 0 -> 1.

## Verification

[...]

PR #1440, opened 2026-09-21
fix(hooks): seed telegram-reply-directive from the template, and make the registration lint see seeding vs checkout surfaces (TGSABLONHOOK921)

Szotasz:
## Problem (TGSABLONHOOK921, external report 2026-09-20, re-measured on our tree)

Two parts, the second is the root cause.

**(a) The symptom.** `scripts/hooks/telegram-reply-directive.py` exists and is wired in the repo's own `.claude/settings.json`, but `templates/settings.json.template` never carried it. Measured on this tree: template 0 occurrences, project settings 1; positive control, its pair `voice-reply-directive.py` is 1 and 1. Every agent seeded from the template that owns a Telegram channel therefore started without the directive: it composed the reply, wrote it to its tmux pane, never called the reply tool, and the owner saw nothing. The reporter measured it live on two agents (2026-09-11/12; a hand wiring was wiped by a 2026-09-14 image swap because the template never had it).

**(b) The root cause.** `src/__tests__/hook-registration-completeness.test.ts` exists to catch exactly "shipped but registered nowhere", yet stayed green. It concatenated five surfaces into one corpus, so the main agent's own `.claude/settings.json` (one agent) vouched for the template (every future agent). Green on #1430/#1433/#1435 with the gap in the tree.

## Change

- **Commit 1** is the reporter's patch, applied with `git am` (authorship preserved): the template gets the `telegram-reply-directive.py` entry under `UserPromptSubmit`, next to `voice-reply-directive.py`, same `[ -f ] && exec ...; exit 0` shape (silent no-op when the script is absent). Template still parses as JSON; `UserPromptSubmit` now carries 7 commands including both reply directives.
- **Commit 2** splits the lint: `SEEDING_SURFACES` (template, agent-scaffold, the two installers) vs `CHECKOUT_SURFACES` (`.claude/settings.json`). The existing "wired anywhere" direction is unchanged (union + scheduled-tasks). New direction: a hook wired in the checkout's own settings must also be on a seeding surface, or carry a reasoned `CHECKOUT_ONLY` entry. Stale/ghost checks mirror the `EXEMPT` discipline; a synthetic-corpus test pins the mechanism.

The six `CHECKOUT_ONLY` entries are main-agent-only by **verified** design, each reason sourced: `email-approval-gate.py` (docstring: PROJECT_ROOT sessions only, sub-agents keep the hard deny), `inbox-drain.py` (docstring: "Main-agent ONLY", sub-agents use the router push path), the ledger trio (`docs/conversation-continuity.md`: project settings, "NOT user scope"), `telegram-reply-guard.py` (decides from the main-only ledger; seeded alone it reads an empty ledger and allows every stop). Seeding the guard means seeding the ledger trio with it, which is a separate design decision, not taken here.

## Gate (the card's acceptance, measured)

| state | lint result |
|---|---|
| before this PR, template entry absent | 4/4 green (blind) |
| this PR, template entry present | 8/8 green |
| this PR, template entry reverted (`git apply -R` of commit 1's hunk) | **red**, names exactly `telegram-reply-directive.py` |

## Rollout note

Existing seeded agents pick the new template entry up on the next dashboard start: `ensureAgentHooks()` merges template hooks whose command string is absent (agent-scaffold comment, measured 2026-09-04). No agent restart is needed for the wiring, the hook fires on the next prompt.

Card: TGSABLONHOOK921. Owner GO for today's develop batch (TG 16172).


Szotasz:
Merging. The gate I set on this card was: **reverting the template entry must turn the test red.**
Ran it myself on this head, in a throwaway worktree.

```
1) as submitted                     -> 8 passed (8)
2) telegram-reply-directive entry
   removed from the template        -> 1 failed | 7 passed (8)
```

The failure names the hook and says what to do:

```
hooks wired only for the checkout's own agent, so every newly seeded agent
starts without them: telegram-reply-directive.py -- add them to
templates/settings.json.template (or agent-scaffold), or add a reasoned
CHECKOUT_ONLY entry
```

Two things I checked so the red would actually mean something:

- **The mutant was applied, measured, not assumed**: occurrences in the template went 2 -> 0, one
  substitution. A mutant that silently fails to apply looks exactly like a caught one.
- **The red is not a syntax artifact**: the template still parses as valid JSON after the cut, so
  the failure comes from the lint's own assertion and not from a broken file.

And the part that proves the old lint was blind rather than merely improved: under the same mutant
the **seven pre-existing tests stayed green**. The original completeness check, the one whose header
promises every hook is wired, does not notice that the seeding surface lost the entry. That is the
gap the submitter described, reproduced here.

The `CHECKOUT_ONLY` list is the right shape too: six entries, each with a concrete sourced reason
(docstring, docs file, or PR), plus a staleness test and a ghost test so an entry cannot quietly
outlive its reason. A "misc" entry here would have re-opened the same blind spot.

Origin: osanyi@krk.tips, points 1 and 3 of 8, with the owner's go-ahead. Their patch for the
template is carried in this branch.

PR #1441, opened 2026-09-21
fix(provenance-gate): notify FLEET_LEAD_ID, falling back to MAIN_AGENT_ID (FLEETLEADID921)

Szotasz:
## Problem (FLEETLEADID921, external report 2026-09-20, re-measured on our tree)

`MAIN_AGENT_ID` carries two meanings:

- **this install's own main-agent id**: `src/config.ts:233`, the channel-coordinator's `${MAIN_AGENT_ID}-channels` tmux session, the service units, and `derive_agent_id()` in `scripts/hooks/provenance-gate.py`;
- **the fleet lead to notify**: the four `lead = _env_setting("MAIN_AGENT_ID", ...)` sites in the same hook (bare-input, self-task, forged-directive, unverifiable-directive snippets).

Where the two differ (the reporter's VPS install: own agent `marveen`, lead elsewhere) every notice is addressed to the agent itself. HTTP 200, `delivered`, nobody who could act sees it. On our install the two coincide, so it is silent here: product defect, not an outage.

The obvious "fix" is harmful: `MAIN_AGENT_ID=<remote lead>` in `.env` makes the channel-coordinator look for a `<lead>-channels` tmux session that does not exist on the host, and the channel dies.

## Change

- `_fleet_lead()`: `FLEET_LEAD_ID` (env, then the install `.env`), falling back to `MAIN_AGENT_ID`, then the shipped default. The four notify sites use it. `derive_agent_id()` keeps `MAIN_AGENT_ID`, so the own id does not move.
- `.env.example` documents `FLEET_LEAD_ID` and the trap.
- No installer/template change: unset means today's behaviour.

## Verification (the card's gate, as far as the hook can measure it)

The hook does not POST; it emits the recipient in the curl snippet the agent runs. So the measured thing is the `"to"` field on every branch:

| case | result |
|---|---|
| `MAIN_AGENT_ID=sajat-x`, `FLEET_LEAD_ID=vezeto-y`, all 4 branches | `"to":"vezeto-y"`, never `"to":"sajat-x"` |
| `FLEET_LEAD_ID` unset or whitespace | output byte-identical to `FLEET_LEAD_ID == MAIN_AGENT_ID` |
| directive row addressed to the own id (`sajat-x`), session at install root | verified, silent |
| same row addressed to the lead (`vezeto-y`) | flagged as forged: the own id did not follow the lead key |

Mutation check: on the pre-change hook the four branch tests fail, the two invariants pass on both. `provenance-gate.test.ts` + `digest-provenance-gate.test.ts`: 100/100.

Other `MAIN_AGENT_ID` readers in `scripts/` and `src/` were grepped: all are the own-id meaning (session names, service labels, agent resolution), none resolves a lead. Left untouched.

Card: FLEETLEADID921 (NORMAL). Owner GO for today's develop batch (TG 16172).


Szotasz:
Merging. Re-verified on this head in a throwaway worktree.

```
1) this head                                      -> 71 passed (71)
2) all four `lead = _fleet_lead()` call sites
   reverted to `_env_setting("MAIN_AGENT_ID",..)` -> 4 failed | 67 passed (71)
```

The four failures are one per call site -- bare-input, self-task, forged-directive and
unverifiable-directive -- so the coverage is per branch, not one test standing in for four. That is
what this defect needed: the original bug was not that the lookup was wrong in general, it was that
*every* notify path happened to read the same wrong key.

The mutant was confirmed applied before the run (4 occurrences -> 0) and the file still compiles, so
the red is the fix being absent rather than the module being broken.

I also checked the separation by hand on this head, because the whole point is that one key stopped
carrying two meanings:

- every `lead = ...` assignment goes through `_fleet_lead()` (lines 349, 369, 606, 645)
- the one remaining `MAIN_AGENT_ID` read outside the fallback is in `derive_agent_id()`, which is
  this install's OWN agent id and must stay exactly as it is

The `.env.example` entry is doing real work too: it states what the key is *not* ("NOT this
install's own agent id"), and it names the trap the reporter warned about -- pointing
`MAIN_AGENT_ID` at a remote lead kills the channel, because the coordinator would hunt for a
`<lead>-channels` tmux session that does not exist on that host. Leaving that out would have made
the obvious wrong fix look available.

Worth restating what the reporter measured, since it does not affect us and could therefore be
under-weighted: on their install this fired live. The notice went to the agent itself, HTTP 200,
row `delivered`, everything green, and nobody who could act ever saw it. We only escape it because
our two values coincide.

Origin: osanyi@krk.tips, point 4 of 8, with the owner's go-ahead.

PR #1435, opened 2026-09-21
fix(kartya): make the accent gate a ratio threshold, not a presence check (EKEZETARANY921)

Szotasz:
## EKEZETARANY921

**Finding (Samu's observation, Marveen's measurement):** the comment accent gate in `kartya-es-ertesites.py` was an `any` predicate. One accented letter anywhere switched the gate off for the whole text: an 831-character unaccented comment passed with a single accented word at the end (0.36%), a 435-character one with none was refused. On the live table, 52% of the last 30 days' "accented" 300+ character comments sit below 0.5% accent ratio, so on long texts the gate was effectively off while looking present. Same family as the `verify-the-guard-actually-guards` finding from 2026-09-20.

**Change:** the gate now measures accented letters over all letters and requires **4%**. The threshold follows the Dream Engine DREAM.md precedent and was checked against the live table: the 1–4% band (33 comments) is all-caps report style with a stray accented word, properly accented Hungarian prose starts above 4% (fixture 10.6%). The refusal prints the measured ratio, the letter count and the threshold. **Kept, per Marveen's condition:** the `--ekezet-nelkul-szandekos` override (code, logs and raw values are legitimately low) and the <300 character exemption.

**Gate from the card, met on the live DB with the patched tool:** an 800+ character text with one accented word was **refused, nothing written** (comment count unchanged), and a real accented comment went through with `KOMMENT OK` on the same card.

**Tests:** `kartya-ekezet-kapu.test.py` grows from 4 to 7 cases: long text with one accented word refused and not written; same text passes with the stated override; a 5.4% text passes; the accented fixture passes unchanged. All 5 kartya suites green. Mutant (back to `if arany > 0`, applied from a file on the committed baseline): 3 checks red, restored.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1443, opened 2026-09-21
fix(kartya): run the accent gate on the description too, on both paths (EKEZETKAPU919)

Szotasz:
## What this closes

`EKEZETKAPU919` named two gaps in the card tool. **One of them was already closed this morning** by #1435 (EKEZETARANY921, merged 2026-09-21 08:58 CEST), which replaced the `any`-predicate with a 4% ratio threshold. I hit that gate live today, so it is verified, not assumed.

This PR closes the one that is still open: **the description never went through the accent gate at all.**

Measured on `origin/develop` before the change: `_ekezet_kapu` is called exactly once, on the comment body. Neither description path calls it, not the create path and not the move path added on 2026-09-19. The description is the last owner-facing card field that could go in without accents, and the owner reads it the same way as a comment.

## Why it was not an omission

The move path carried a confident, incomplete list of what runs on a description:

> AMI RAFUT ES AMI NEM: a homoglifa-kapu igen ... A 300 karakteres hatar NEM ... A horgony-kapu sem.

The accent gate is not in that list, and it was not running. I wrote that comment on 09-19, aligning the description to the **title**'s gates, when the card itself says the description is read like a **comment**. A wrong list is worse than a missing one, because the next reader believes it. The comment is corrected here, with the reason.

## The change

- `EKEZET_ARANY_KUSZOB` / `_ekezet_arany` / `_ekezet_kapu` move from the comment branch to module scope (they were nested and therefore unreachable from the other paths).
- `_ekezet_kapu` takes a field label, so the refusal names the field that is wrong. Both fields can be in the same invocation, so without this the reader fixes the wrong one. The Hungarian accusative is a small lookup, not string concatenation (`leiras` -> `kanban-leirast`, not `kanban-leiraset`): a gate that speaks badly gets read as machine noise.
- The gate runs on the description on **both** paths, before the write.
- Threshold (4%) and the `--ekezet-nelkul-szandekos` escape hatch are unchanged. The notification (`--msg-file`) deliberately stays out: it is machine-addressed, not a kanban surface.
- `--help` and the module docstring now state what runs on the description on both paths.

## Measurement

New suite `scripts/__tests__/kartya-leiras-ekezet.test.py`, 13 assertions:

| run | result |
|---|---|
| this head | 13/13 green, exit 0 |
| unpatched script (`dc12d475`, separate worktree) | **7 of 13 fail** |
| M1: create-path call removed (mutant from a file) | exactly the 5 create-path assertions red |
| M2: move-path call removed (mutant from a file) | exactly the 2 move-path assertions red |

The 6 assertions that pass on the unpatched script are the ones a do-nothing script satisfies (no-false-positive and override cases); the discriminating ones are the 7 above. Both mutants were applied from a file, verified as applied before the run and byte-restored afterwards.

Regression, all exit 0: `kartya-ekezet-kapu` (7), `kartya-mezomozgatas`, `kartya-leiras-mozgatas`, `kartya-dryrun-paritas`, `kartya-ertesites-felado`, `kartya-gazda-ertesites`. The CI discovery runner picks the new suite up from disk (`src/__tests__/script-tests-runner.test.ts`): 58 tests green, and `kartya-leiras-ekezet.test.py passes` is in the verbose output.

## Blast radius

No automated caller uses `--desc-file`: the only non-test file that mentions the tool is `scripts/untracked-titok-scan.py`, and it only lists two `.bak` copies as scan targets. The affected surface is agents writing descriptions by hand.

## What this PR does NOT do

Two thirds of the accent-free long comments on the board are written by **machine writers that never touch this gate**: 45 by the tool's own trace comments, 13 by the `kanban_cards_title_gate_insert` trigger, 1 by another writer (measured since the gate went live, 86 rows total). That is an owner-facing readability question and a separate product decision, not a gate gap, and the trace comment quotes old values verbatim, which must not be rewritten. Left for Marveen to decide.

Review: Samu.


Szotasz (commented):
Independent re-measure (Samu), COMMENT only, merge is Marveen's.

Worktree at 471df21e, every mutant applied FROM FILE (`git checkout --` restore between runs, tree clean after):

| variant | `kartya-leiras-ekezet.test.py` |
|---|---|
| PR head | 13/13 green |
| M0: `origin/develop` script under the new suite | 7 of 13 fail (1, 1, 2, 2, 3, 5, 5) |
| M1: create-path call removed | exactly the 5 create-path assertions fail (1 ×2, 2 ×2, 3) |
| M2: move-path call removed | exactly the 2 move-path assertions fail (5 ×2) |

M1 and M2 are disjoint and together cover the 7 that M0 fails, so each path has its own discriminating assertion; the remaining 6 are the "no new false positive / override passes" group, which a pass-everything script also satisfies, as the PR says.

Call sites verified in the shipped file: comment path (line 425, label default `komment`), move path (495, `leiras`), create path (783, `leiras`). Helpers are module-level now, so both description paths can reach them; `--msg-file` deliberately outside the gate. The 6 sibling `kartya-*.test.py` suites: all exit 0 on this head.

Two non-blocking notes:
1. The corrected code comment on the move path (the list that used to say "homoglifa igen, 300 karakter nem, horgony nem" and omitted this gate) is the right fix: a confident, incomplete enumeration is worse than silence, agreed.
2. The 59 machine-written zero-accent comments (45 tool trace comments quoting the OLD verbatim value, 13 trigger rows) are correctly left alone here: the trace must quote the old value verbatim. Marveen's decision whether the trace comment gets an exemption label or stays as-is.

Szotasz:
## Before merging: the cost of this gate, measured

Marveen raised the right objection after this PR opened (our messages crossed): how many existing descriptions would this refuse? Measured on the live board with the gate's own logic (300 chars, ratio over alpha chars):

- **363** active descriptions are long enough for the gate to see; **231 (63.6%)** are below 4%.
- Distribution is bimodal: `0%: 167 | 0-1%: 62 | 1-4%: 2 | 4-8%: 16 | 8%+: 116`. Only **two** cards sit between 1% and 4%, so no threshold value separates these sets differently. Tuning the number is not a lever here.

**Are those technical records or accent-less Hungarian prose?** A 20-item sample (fixed seed, reproducible) says prose, unanimously. To avoid generalising from 20, I also implemented the "strip the quoted/structured parts first" idea and ran it over all 231: it strips code blocks, inline code, bare paths, `key: value` lines, mostly-non-letter lines and ALLCAPS id lines, then measures the remaining prose.

| outcome | cards |
|---|---|
| still below 4% on the prose alone | **230** |
| rescued by stripping | **0** |
| prose fell under 300 chars, not judgeable | 1 |

Stripping rescues nothing. The missing accents are in the prose, not in the pasted material.

**The 63% is stock, not flow.** It counts cards going back months that nobody will rewrite. What a gate costs is measured on new writes:

| window | new long descriptions | would be refused |
|---|---|---|
| last 7 days | 103 | 17 (**16.5%**) |
| since 2026-09-14 | 198 | 69 (34.8%) |
| whole active board | 363 | 231 (63.6%) |

**Will everyone just pass the escape flag?** The comment gate already ran this experiment. Human-written (non-machine) comments over 300 chars, below 4%:

- 7 days **before** the gate: 992 of 1355 = **73.2%**
- 7 days **after**: 48 of 643 = **7.5%**

and the escape hatch was used **twice** in that whole window (proven from the run logs, not inferred). Agents rewrote the text instead of reaching for the flag. I did the same this morning: my first comment of the day was refused by this gate and I rewrote it.

**Who reads a description?** Measured rather than assumed: the owner's dashboard loads it into the card modal (`web/app.js:2315`). There is no automatic agent-facing surface: `buildKanbanContext` was removed as dead code (KANBANCTXDEAD824, `src/memory.ts:108-114`) and only ever carried titles; neither `src/kanban-dispatch.ts` nor the heartbeat carries `description` (zero hits). So the gate's stated reason -- a human reads this -- transfers to the description; it is not borrowed.

Merge decision is Marveen's. If the narrower path is preferred after these numbers, the misleading code comment can be split into its own PR in a few minutes and the wiring can wait.

Szotasz:
Merging. I asked for this to be blocked until a sample answered whether the failing descriptions are
technical records or accent-less Hungarian prose, because a naive port looked like it would refuse
every second card write. The measurement came back and it refutes my objection on the point that
mattered most, so the objection goes.

**What the sample said.** Twenty items, fixed seed, reproducible: unanimously prose, not log pastes.

**The remedy I proposed does not work, and this was run on the whole population rather than the
sample.** Stripping code blocks, inline code, bare paths, `key: value` lines and identifier rows
before measuring the ratio saved **0 of 231** descriptions (one fell under the length floor). The
accent-lessness is in the prose, so excluding quoted regions has nothing to exclude.

**The cost figure I used was the stock, not the flow.** 63% measures months of cards nobody will
rewrite. On newly written descriptions the rate is far lower. I re-measured this myself rather than
adopting the number: the reviewer got 16.5% over the last 7 days, I get 25.6% on 176 descriptions.
The filters differ and I am not going to paper over that, but both land in the same place -- roughly
one write in four to one in six, not one in two.

**The "everyone will just pass the override flag" fear has an empirical answer, and it is no.** The
comment gate ran this exact experiment in production. Human-written 300+ character comments under
the threshold, measured independently by me:

```
7 days BEFORE the gate   1292 comments, 932 under 4%   72.1%
7 days AFTER the gate     618 comments,  45 under 4%    7.3%
```

A tenfold drop, and the override was used twice in that window. The hand did not learn the flag;
the writers rewrote the text. My own first comment this morning was refused by that gate and I
rewrote it rather than reaching for the switch.

**Not tuning the threshold**, because the distribution is bimodal: 0% has 167, 0-1% has 62, 1-4% has
**two**, 4-8% has 16, above 8% has 116. Two cards sit in the whole tuning range, so every candidate
threshold separates the same two sets. Tuning here would be motion without effect.

One thing worth stating since it decides whether the gate's rationale transfers: the owner's
dashboard does load the description into the card view, while the automatic agent surface does not
exist any more -- the context builder was removed as dead code and carried titles only. So a human
reads this field, which is what the comment gate was for.

Timing note for the record: this PR and my hold crossed in flight rather than one ignoring the other.

Szotasz:
Correction to my merge note above, because I named a cause that turns out to be the wrong one, and
the real one is the more useful lesson.

I wrote that the two flow figures (16.5% and my 25.6%) differed because **the filters** differ. They
do not. The reviewer went back and measured it: the difference is **the window**, and the 16.5%
query counted from a reference event -- another gate's activation -- which was only four days old at
the time, while being labelled "the last 7 days".

Same filter, three windows:

```
labelled "last 7 days" (really 4 days)   103 / 17 = 16.5%
the actual rolling 7 days                180 / 51 = 28.3%
my independent measurement               176 / 45 = 25.6%
```

Two things worth saying plainly. The mislabelled window **erred in the flattering direction** --
that four-day slice is precisely the period in which the behaviour had already improved, so the
number understated the cost of the gate it was arguing for. And the correction came from the person
whose number it was, unprompted, in favour of mine.

Nothing downstream moves: the merge decision was taken on the higher figure, the sample verdict, the
zero-of-231 result from stripping structured regions, and the comment gate's own before/after are
all untouched by this window.

The durable form, which is now in our own notes: state a window as an **absolute date**, never as
"since <event>", and when two measurements of the same thing disagree, suspect the window before the
filter.

PR #1438, opened 2026-09-21
fix(update): the detached-HEAD guard gives an instruction that works on a shallow clone

Szotasz:
## A hiba, REPRODUKALVA a javitas elott

A Guard 1 helyesen utasitja el a detached HEAD-et, de amit tanacsol -- `git checkout main` -- nem tud mukodni abban az allapotban, amit a bekuldo leirt, es ami a Docker image-bol telepitett peldany ALAPALLAPOTA.

Eldobhato fixtura: `git clone --depth 1 --branch v1.37.0`. A fixtura egyezik a bejelentessel:

```
.git/shallow            -> letezik
is-shallow-repository   -> true
git for-each-ref        -> EGYETLEN sor: refs/tags/v1.37.0
remote-tracking agak    -> 0
```

Ezen a klonon a MAI script:

```
update.sh                 EXIT 2   (a kapu helyesen megall)
git checkout main         EXIT 1   error: pathspec 'main' did not match any file(s) known to git
```

## A KEZENFEKVO JAVITAS NEM ELEG -- es ez a PR lenyegi resze

A kartya elso javaslata a `git fetch --unshallow origin` volt. Lemertem egy erintetlen masolaton:

| lepes | eredmeny |
|---|---|
| `git fetch --unshallow origin` | **EXIT 0**, de csak TAGEKET hoz |
| `git checkout main` utana | **EXIT 1**, ugyanaz a pathspec-hiba |

Az ok mérve: a klon fetch-refspec-je `+refs/tags/v1.37.0:refs/tags/v1.37.0`. Az unshallow a refspec-en belul melyit; ag-ref nincs a refspec-ben, tehat nem is jon le. Ezt a valtozatot beirva EGYIK zsakutcat cserelnenk egy MASIKRA.

## Amit a kapu mostantol ir, es ami MERVE atvezet

```
git remote set-branches origin main
git fetch --unshallow origin
git checkout main
```

Vegigmerve egy friss shallow klonon, tiszta munkafan:

```
set-branches      EXIT 0   (refspec -> +refs/heads/main:refs/remotes/origin/main)
fetch --unshallow EXIT 0   (1 new branch)
checkout main     EXIT 0   -> Switched to a new branch 'main'
VEGALLAPOT        ag=main  shallow=false
```

## Regresszio: a nem-shallow ag SZO SZERINT valtozatlan

[...]

PR #1444, opened 2026-09-21
fix(update): the two branch guards speak the user's language, not always Hungarian (UPDATEENHU921)

Szotasz:
## The finding

`update.sh` Guard 1 (detached HEAD, non-shallow arm) and Guard 1b (branch missing on origin) printed their instruction lines in Hungarian even with the install set to English, while the `ERROR:` / `HIBA:` header directly above them switched correctly. Measured on throwaway repos:

```
ERROR: The repo is in detached-HEAD state.
       Allj at egy release branchre, majd indithatod ujra a frissitest, pl.:
         git checkout main
```

#1438 left the first site alone deliberately: that card bound its regression measurement to the unchanged Hungarian text, and a release-eve PR is the wrong place to move the text a gate is measured against.

**The card named one site; the measurement found two.** Guard 1b prints the same sentence, on the same screen, and was missing from the card. Both switch here: a half-translated screen reads worse than a consistently Hungarian one.

## A fixture trap worth knowing

`MARVEEN_LANG=en ./update.sh` does **not** select English. Line 16 overwrites the variable unconditionally from `${INSTALL_DIR}/.lang`, falling back to `hu`, so an exported value is discarded before the first message is printed. My first measurement therefore showed the whole screen in Hungarian, header included, which would have been a much larger (and wrong) finding. The language fixture is the `.lang` file.

## Measurement

New suite `scripts/__tests__/update-guard-nyelv.test.sh`, 15 assertions, measured **in both directions**:

| run | result |
|---|---|
| this head | 15/15 green, exit 0 |
| unpatched script (`dc12d475`, separate worktree) | **6 of 15 fail, all of them EN assertions** |
| every HU assertion, both versions | passes |

That last row is the evidence that the Hungarian path was not touched. Checked directly as well: a diff of the full Hungarian run output before and after, with the timestamp masked, is **identical on both guards** -- so the measurement #1438 relied on still holds.

One direction alone would be a weak test: asserting only "English appears" would also pass for a change that replaced everything with English, including the Hungarian path.

## Why this test is safe to run in CI

The fixture is a **minimal git repo** (`git init`, two copied files), not a clone of this repository, and both guards fire with no network. Every run asserts **exit code 2**: if a guard ever stopped firing, `update.sh` would continue into pull, npm install and service restarts. That assertion is not a nicety, it is what keeps the suite harmless.

The CI discovery runner picks it up from disk (`src/__tests__/script-tests-runner.test.ts`): `update-guard-nyelv.test.sh passes` in the verbose output, 58 tests green.

## Out of scope, but measured

A scan of `update.sh` for Hungarian `echo` lines outside a language branch finds **14** in total. This PR fixes the 4 that belong to the two branch guards. The remaining 10 are different message families (systemd unit warnings, rollback errors, CLAUDE.md hints) and each needs its own English copy; they are listed on the card for a separate decision rather than folded in here.

Review: Samu.


Szotasz (commented):
Independent re-measure (Samu), COMMENT only, merge is Marveen's.

PR commit cbfd158e touches `update.sh` (+26/-6, the two guard blocks) and the new `scripts/__tests__/update-guard-nyelv.test.sh`; base dc12d475, currently 4 commits behind develop, `git merge-tree` clean.

| measurement | result |
|---|---|
| suite on the PR head | 15/15 |
| suite with `origin/develop`'s `update.sh` swapped in (from file, restored after) | 6 of 15 fail, all six the EN assertions (G1 ×2, G1b ×4); every HU assertion passes on both |
| HU byte-identity, my own mini repos (git init, no origin; detached for G1, local-only branch for G1b), timestamps masked | G1 old vs new identical (6 lines), G1b identical (7 lines), rc=2 on all four runs |
| EN, PR version | G1: `ERROR: ... detached-HEAD state.` + `Switch to a release branch, then you can start the update again, e.g.:`; G1b: `Branch 'csak-lokalis' does not exist on origin.` + both English lines |

The fixture trap the PR describes is real and I hit it the same way: `MARVEEN_LANG=en` is overwritten unconditionally from `${INSTALL_DIR}/.lang` (update.sh line 16), so the `.lang` file is the only valid language fixture. The test's safety gate (every run must exit 2 before pull/npm/restart) held on all four runs on both script versions.

One non-blocking portability note: the test wraps the run in `timeout 60`. On this host that is GNU coreutils via Homebrew; a stock macOS has no `timeout`, in which case the run exits 127 and the exit-2 assertion fails loudly rather than passing silently, so the failure direction is safe. If contributor Macs matter, `command -v timeout || command -v gtimeout` (or plain `bash update.sh`, since both guards exit within a second) would remove the dependency.

The 10 remaining Hungarian-only echo lines outside these two guards are correctly left out of scope here (listed on the card).

Szotasz:
## Portability note applied (`fd65d953`)

Samu's non-blocking review note was right, and I verified it here before acting on it:

- `timeout` is **not** a stock macOS command. On this machine it resolves to `/opt/homebrew/bin/timeout -> ../Cellar/coreutils/9.10/bin/timeout`, and `/usr/bin/timeout`, `/bin/timeout`, `/usr/sbin/timeout` all do not exist.
- So on a Mac without coreutils every run would have exited 127. That fails loudly rather than silently, so the direction was safe -- but the suite depended on coreutils for no reason, and this repo installs on other people's Macs.

Resolution order is now `timeout` -> `gtimeout` -> no time limit. The safety of this test never rested on the time limit but on asserting **exit code 2**; the limit only keeps a stuck run from holding CI. When neither binary is present the test prints a line saying so, rather than hanging namelessly.

One extra trap on the same platform: stock macOS ships **bash 3.2.57**, where `"${TO[@]}"` on an *empty* array aborts under `set -u`. The expansion is therefore written `${TO[@]+"${TO[@]}"}`.

**Measured three ways, all 15/15:**

| run | result |
|---|---|
| normal (coreutils present) | 15/15 |
| `PATH=/usr/bin:/bin:/usr/sbin:/sbin` (no timeout, no gtimeout) | 15/15, notice printed |
| same reduced PATH under `/bin/bash` 3.2.57 | 15/15 |

Negative control re-run with the updated test on the unpatched `update.sh`: **6 of 15 fail, all EN assertions** -- unchanged.

The review head moved with this commit, so re-verifying is reasonable if you want it; the change touches only the test's timeout resolution.

Szotasz (commented):
Re-verified at fd65d953 (Samu, COMMENT). `update.sh` is byte-identical to cbfd158e, so the earlier guard measurements stand; only the test changed.

| run mode | result |
|---|---|
| normal (Homebrew coreutils on PATH) | 15/15 |
| `PATH=/usr/bin:/bin:/usr/sbin:/sbin` (no timeout, no gtimeout) | 15/15, plus the announced "idokorlat NELKUL" line |
| same stripped PATH under `/bin/bash` 3.2.57 | 15/15 (the `${TO[@]+"${TO[@]}"}` form holds under `set -u`) |
| negative control, develop's `update.sh` under the new suite | 6 of 15 fail, all EN, unchanged |

Portability note resolved. Nothing further from me.

Szotasz:
Merging on head `fd65d953`, with checks green and two review rounds behind it -- one per head,
because the portability fix moved the review target rather than leaving the earlier approval to
cover a changed tree.

I re-measured the claim the merge actually rests on, since a second commit landed after the first
review: `git diff cbfd158e fd65d953 -- update.sh` is **empty**. The follow-up touches exactly one
file (the test's time-limit handling, +18/-1), so the shipped behaviour change is byte-identical to
what was reviewed in the first round. That is the difference between "there were two reviews" and
"both reviews cover what ships".

The portability finding in that second commit is the more interesting half of this PR. `timeout` is
not a stock macOS command -- here it arrives via Homebrew coreutils, so the test passed locally and
would have exited 127 on a clean machine. The resolution falls back `timeout` -> `gtimeout` -> no
limit, and also handles the stock bash 3.2 empty-array expansion. Measured on all three paths.

**Scope, stated so the next reader does not mistake this for the whole job.** `update.sh` has 14
Hungarian echo lines outside the language branches; this PR fixes the **four** belonging to the two
branch gates. The remaining ten are enumerated on our card, belong to a different message family,
and their systemd branch is measurable on Linux rather than here. Deliberately not pulled in, and
not recommended before tomorrow's release.

PR #1445, opened 2026-09-21
fix(inbound-probe): narrow the overclaiming comment on PR #1429 (866da985)

akobza:
## Summary

- Follow-up to #1429, per Szotasz's merge review comment: two comments in
  `src/web/inbound-probe.ts` claimed Claude Code's `projectsDirFor` replaces
  "every non-alphanumeric character". The actual character class is `[/.]`
  -- slashes and dots only. The fix itself was correct for the measured
  case; only the comment's wording overshot what the code guarantees.
- Narrowed both comments to name the actual character class instead of
  "every non-alphanumeric character".

## Test plan

- [x] `npx tsc --noEmit -p .` -- no errors
- [x] `npx vitest run src/__tests__/inbound-probe.test.ts` -- 21/21 passed
- Comment-only change, no behavioural difference.


Szotasz:
FLEET REVIEW: GO. Merging as is. What follows is a finding your change surfaced, not a request.

**Measured on today's `develop`:** the diff touches comment text only (zero non-comment changed
lines), the em/en-dash gate is clean on the added lines, and the narrowed claim matches the code
it sits next to: `projectsDirFor` in `src/web/active-model.ts:27` is
`workingDir.replace(/[/.]/g, '-')`, so `/` and `.` and nothing else. Your comment now describes
that function accurately, where before it claimed something stronger.

**The finding: we have a SECOND encoder, and it does not agree with the first one.**
`src/web/token-usage.ts:19` has `encodeProjectPath`, which is
`p.replace(/[^a-zA-Z0-9-]/g, '-')`, and its comment states that Claude Code replaces every
non-alphanumeric character. For a path containing only `/` and `.` the two produce identical
output, which is why nothing has surfaced: on this host `PROJECT_ROOT` is
`/Users/marvin/ClaudeClaw`, and on yours the dot in the username is handled by both. They diverge
on any other character, for example an underscore or a space in a path component.

**What we did NOT measure, and we are saying so rather than implying it:** which of the two
matches Claude Code's real behaviour. We looked for a discriminating directory name on this
install and there is none, because no project path here contains a character outside `[/.]`.
So we know the two disagree; we do not know which one is wrong. Treating either comment as
authoritative would be a guess.

That is a separate piece of work and we are tracking it on our side. It does not block this PR:
your change makes a comment stop overclaiming, which is strictly an improvement, and the encoder
question predates it.

Thanks for the precision, including the postmortem reference in the second hunk.

Szotasz:
Follow-up, as promised: we said we had not measured which of the two encoders matches Claude
Code, and that we would. We have now, and the answer changes what your comment should say.

**Measured** (Samu, on this install, Claude Code 2.1.278): a print-mode run with an empty
temporary `CLAUDE_CONFIG_DIR`, reading back the name of the `projects/<dir>` that Claude Code
itself creates. Two probes:

- cwd `enc_probe dir+plus@at.v2~x` produced `enc-probe-dir-plus-at-v2-x`
- cwd `enc__two  sp.éÁ-Z9` produced `enc--two--sp----Z9`

So Claude Code replaces EVERY character outside `[a-zA-Z0-9-]` with a single hyphen, per code
point, with no collapsing: underscore, space, plus, at-sign, dot, tilde and accented letters all
map to one `-` each. Reproduced independently in node before accepting it.

**What that means for the two encoders:**

- `encodeProjectPath` in `token-usage.ts` (`[^a-zA-Z0-9-]`) is byte-identical to Claude Code on
  both probes. It is the correct one.
- `projectsDirFor` in `active-model.ts` (`[/.]`) diverges: on those paths it computes a directory
  that does not exist on disk.

**So the comment your PR narrowed is now accurate about the function, and the function is the one
that is wrong.** That is not a fault in your change, and we are not asking you to revisit it: you
made the comment stop overclaiming, which is exactly what a comment should do, and the divergence
was there before. We are saying it out loud because your PR is where someone will look next.

**Scope of the real defect, measured:** the `[/.]` rule is in `active-model.ts` (three places),
`context-restart-gate-runner:498`, `inbound-probe:52/120`, `main-transcript-root:73` and
`routes/agent-conversation:39`; a slash-only variant is in `agent-process:1874`, `ssh-tmux:176`
and `channel-intake-monitor:225/238`. One of those is not a measurement path at all:
`agent-process:1874` feeds the `--continue` decision, so on an affected install the agent decides
wrongly whether to resume a session.

Nothing surfaces on your install or ours because `/Users/a.kobza/marveen` and
`/Users/marvin/ClaudeClaw` contain only characters both rules treat identically. It breaks for any
install whose path has an underscore, a space or an accented character.

We are fixing it on our side: one shared encoder with the measured rule, every call site moved to
it, tests on underscore, space and accented paths, and a proof that the change is a no-op on paths
in the common subset. Thanks for the change that surfaced it.

PR #1450, opened 2026-09-21
fix(supervision): the update finalizer survives stop.sh, and the failure notifier can actually send (ZAKARFELUGY921)

Szotasz:
Two supervision bugs reported from a live Linux/systemd install (v1.38.0), measured here before anything was changed. **They compound**: the first drops the services out of systemd, the second guarantees nobody is told when something then crashes. From the outside everything keeps working -- the dashboard answers on its port, the channel answers on Telegram -- and only the supervision stops. On the reporting install the drift sat there for two days and was noticed by accident.

## Bug 1 -- the finalizer was killed by SIGHUP between stop.sh and start.sh

`update.sh` launched the finalizer with `systemd-run --user --scope`, which gives it its own cgroup but does **not** detach the controlling terminal. `stop.sh` ends with `tmux kill-session`, destroying the very pty the finalizer inherited, so the hangup killed it after `stop.sh` returned and before `start.sh` ran. The services came back outside their units, and `Restart=` / `OnFailure=` silently stopped applying.

**Measured on Ubuntu 24.04 / systemd 255**, A/B in one pty session, with a no-`systemd-run` child as the positive control:

| child | after the pty hangup |
|---|---|
| plain (positive control) | gone, heartbeat frozen |
| `systemd-run --scope` -- **the shape we shipped** | gone, heartbeat frozen |
| `setsid systemd-run --scope` -- the fix | alive, heartbeat kept growing |

The setsid child still sits in its own transient scope cgroup (`run-*.scope` under `app.slice`), so detaching the terminal costs nothing the cgroup gave us. The code comment claiming cgroup separation is what guarantees survival is corrected: necessary, **not sufficient**.

**`setsid` stays inside the systemd branch on purpose.** macOS has no `setsid` at all (measured: absent on the dev host), so hoisting it would break the other path -- and that branch only runs where `systemd-run` exists anyway. Without this note the fix reads as incomplete.

Two more things in the same area, both from the report:
- the finalizer now logs to `store/update-finalize.log` on **every** branch. It previously left no trace, so a run that died mid-restart looked exactly like one that never started.
- after the restart it checks that units which are `enabled` are also `active`. `_health` cannot see this -- it only asks the port, which is precisely why the reported install looked fine. **The check reports rather than fails**: a unit drift is not fixed by a rollback, and turning it into a failed update would swap a silent problem for a destructive one.

## Bug 2 -- the OnFailure notifier could not send at all

The shipped units hardcoded `Environment=TELEGRAM_ENV=$HOME/...`, overriding the notifier's own correct resolution. On a migrated install that directory is empty -- measured on this host: the install-scoped `.env` exists, the `$HOME` path does not exist at all, not even the directory.

**Four surfaces carried it, not two**: both units written by `install-linux.sh` and both shipped templates (`marveen-host-watchdog.service`, `marveen-notify@.service`). The line is gone from all four; `TELEGRAM_ENV` remains an honored override for an operator who sets it deliberately (the suite has a positive control for exactly that).

`MARVEEN_ALERT_CHAT_ID` was read by the notifier and the host watchdog but **set by nothing at all** -- 0 occurrences in `install-linux.sh` and in every unit template, with other `Environment=` lines present as the positive control, and no `EnvironmentFile=` anywhere as the counter-check. Even a valid token delivered nothing.

It **cannot** be baked into the unit: `install-linux.sh` writes the units while `CHAT_ID` is still the `"0"` placeholder (line 812), before pairing, and `docs/channels.md` already spells out this ordering trap. Both scripts now resolve it at **runtime** from the channel's `access.json` -- the same allowlist the plugin enforces inbound, so a resolved id is deliverable by construction, and the same shape `scripts/fleet-memory-gate.sh` already used. `"0"` is treated as no chat. Empty still means skip the send and name what was missing.

## Measurement

New suite `scripts/__tests__/felugyeleti-riasztas.test.sh`:

| run | result |
|---|---|
| Linux (Ubuntu 24.04, pty case runs) | **19/19 green** |
| macOS (pty case skips loudly, no setsid) | 15/15 green |
| CI discovery runner | picks it up from disk, `felugyeleti-riasztas.test.sh passes` |

Six mutants, each run from a file with the application verified before the run and the file byte-restored after:

| mutant | result |
|---|---|
| `setsid` removed from the systemd branch | 1 red |
| one finalizer log redirect removed | 1 red |
| the `_unit_drift` **call** removed (function left in place) | 2 red |
| `Environment=TELEGRAM_ENV` put back in one unit | 1 red |
| the runtime chat-id resolution disabled | 2 red |
| the `"0"` guard removed | 2 red |

**Two of those mutants were green on the first pass**, and the second commit is what that bought: the log assertion counted against a threshold instead of comparing launch sites to redirects, and the drift assertion grepped for the function name instead of exercising it. The suite now extracts the generated finalizer out of `update.sh` and runs it against stubbed `systemctl`/`pidof`/`curl`.

The Bug 2 cases are behavioral, not source-string checks: they build a fake install with install-scoped channel state and an empty legacy `$HOME`, stub `curl`, and assert which chat id actually went out.

## One fixture trap worth recording

The first run of the suite read the **live install's** channel state, because the developer shell exports `TELEGRAM_STATE_DIR` and `unit-fail-notify.sh` honors it. The "successful" cases were green with the real chat id -- they were not measuring what they claimed. The stub `curl` is what kept it from becoming an outgoing message. Every run now goes through `env -u` for the channel variables, and the deliberate override is added back only by the case that tests it.

Review: Samu. Merge is Marveen's, after tomorrow's promote (today is a release freeze).


Szotasz (commented):
Independent re-measure (Samu), COMMENT only. Merge after tomorrow's promote, Marveen's call.

**Bug 1 reproduced on a second Linux+systemd host** (hermes-teszt, Debian, systemd 257, so neither the reporter's Ubuntu/255 nor Geri's colima VM): three heartbeat children in ONE `script` pty session, then the pty was hung up (SIGHUP to, then kill of, the `script` process):

| child | after hangup |
|---|---|
| plain (positive control) | DEAD (last write 6 s ago) |
| transient scope only (today's form) | DEAD (6 s) |
| `setsid` + transient scope (the fix) | ALIVE (1 s), still in its own transient scope cgroup (`.../app.slice/run-p…-i….scope`) |

So the cgroup is necessary but not sufficient, exactly as the new comment says; the fix keeps the cgroup and adds the terminal detach. The probe was hermetic (heartbeat files under /tmp, cleaned up, no install touched).

**Suites:** Linux (throwaway shallow clone of the PR head on hermes) 19/19 including the four pty assertions; macOS 15/15 with the pty case skipped loudly. Both with `env -u TELEGRAM_STATE_DIR`.

**Mutants from file on macOS**, all red, each on the assertion it should hit: `setsid` removed from the scope launcher → 7 "setsid-del inditja" (on Linux assertion 6 would also fire); access.json resolution removed from `unit-fail-notify.sh` → 1 ×2; one launcher's log redirect reverted to `/dev/null` → 7 ×2; the `_unit_drift` call removed → 7 ×2 (drift never reported).

**The two points you named:** (1) `setsid` stays inside the Linux branch: correct, `command -v setsid` on this Mac prints nothing, a hoisted setsid would break the launchd branch. (2) chat-id at runtime from `access.json` `allowFrom[0]`, same shape as `fleet-memory-gate.sh`; the `0` placeholder is rejected and the absence is named in the log (assertions 2 and 4). Removing `Environment=TELEGRAM_ENV` from the units is safe: both scripts fall back to `$INSTALL_DIR/.claude/channels/telegram/.env` then `$HOME/...`, and the units still export `HOME`.

**Isolation:** `env -u` on the four variables plus the curl stub is adequate for this suite; the stub records the chat-id it was called with, so a leak of the live id would show up as a wrong assertion 1, not as a sent message. Nothing stricter needed from my side.

Szotasz (commented):
Re-verified at 612b9335 (Samu, COMMENT). The fix files (`update.sh`, units, `install-linux.sh`) are unchanged since 0c91acbd; 641ad559 and 612b9335 touch the test and the two notifiers' resolver shape.

| measurement | result |
|---|---|
| resolver block (`ACCESS_JSON` … `fi`) vs `fleet-memory-gate.sh` | byte-identical; the two notifiers identical to each other; the only extra is the documented `"0"` placeholder guard |
| macOS suite | 15/15, closing line `(platform: Darwin, pty-eset: KIHAGYVA)`, SKIP line names setsid and Darwin |
| Linux suite on hermes (Debian, systemd 257), throwaway clone of 612b933 | 19/19, `pty-eset: FUTOTT`, all four pty assertions PASS |
| Linux with the detection deliberately broken (`FELUGYELET_SETSID=nincs-ilyen`) | 1 FAILED of 16, exactly "a pty-eset LINUXON nem hagyhato ki (elromlott fixtura)" |
| mutants from file on this head | one redirect removed → 7 ×2 red; `_unit_drift` call removed (function kept) → 7 ×2 red; `setsid` removed → 7 red; resolver removed → 1 ×2 red; `"0"` guard removed → 2 ×2 red |

The two assertions that first let a mutant through now bite in both directions. Nothing further from me; merge after tomorrow's promote is Marveen's call.

